PR #15, opened 2017-05-25
New unsigned datatypes, Removed  std::to_string

Ninja-007:
Removed std::to_string usage
Added unsigned char
Added signed char
Added unsigned short int
Added unsigned int
Removed const return types to avoid complier warnings

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -214,6 +214,20 @@ namespace sqlite_orm {
             return res;
         }
     };
+
+	//Note unsigned char used for storing integer values, not char values.
+    template<>
+    struct type_printer<unsigned char> : public integer_printer {};
+
+    template<>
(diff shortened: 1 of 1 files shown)


fnc12:
why did you removed `const` from `const bool operator==(const iterator_t &other) const`? It is recommended to add const to make statements like this `(it1 == it2) = it3;` unavailable

Ninja-007:
I was getting complier warning. 

sqlite_orm.h:2651: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
                 const bool operator==(const iterator_t &other) const {
                                                               
gcc (GCC) 6.2.1 20160916 (Red Hat 6.2.1-3)

fnc12:
Ok. Thanks you very much. I accept your pull request and remove `cout`s at all in the next commit.

Ninja-007:
Thanks. :)

PR #17, opened 2017-05-25
Complier warnings fixed

Ninja-007:
Changed auto to size_t in for loops which involve std::collection to remove complier warnings of comparision between signed and unsigned.


fnc12:
Looks like you copy and pasted my latest commit code instead of merging it. Let me fix warnings by myself - other improvements already exist

PR #28, opened 2017-06-08
refacored sync_schema

Ninja-007:
refacored sync_schema


fnc12:
Thanks for standard lib style notation - all public functions must have it cause if once ISO C++ wants to add an sqlite lib it got to be `sqlite_orm` ;)

PR #31, opened 2017-06-08
Updated result values of sync_schema

Ninja-007:
Updated result values of sync_schema

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -2253,22 +2253,14 @@ namespace sqlite_orm {
         sqlite3 *db = nullptr;
     };
     
-    enum class sync_schema_result {
-        dropped_and_created,
-        created,
-        columns_changed,    //  data isn't altered
-        synced, //  everything is ok
-    };
     
-    inline std::ostream& operator<<(std::ostream &os, sync_schema_result value) {
-        switch(value){
-            case sync_schema_result::columns_changed: return os << "columns_changed";
-            case sync_schema_result::created: return os << "created";
-            case sync_schema_result::dropped_and_created: return os << "dropped_and_created";
-            case sync_schema_result::synced: return os << "synced";
-        }
-    }
-    
+enum class sync_schema_result {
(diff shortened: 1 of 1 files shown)


fnc12:
why do you change enum names? There is no need

Ninja-007:
I made them more descriptive.

Ninja-007:
Well, it is clear for you. But for users of library who are not deep in the library code. It is hard understand short enum values. It will be great if you keep it.

fnc12:
1) `tbl` is uncomfortable short version for eyes.
2) `new_columns_added` is not correct cause it can be returned in case if columns deleted 
3) `already_in_sync` and `synced` has same meaning but `already_in_sync` has more chars - not good
4) dropped_and_recreated is ok

Ninja-007:
Thanks
1. tbl can be renamed to table.
2. I believe in case of deletion of excess columns. It will go to                 if(!gottaCreateTable){  //  if all storage columns are equal to actual db columns but there are excess columns at the db..
                    if(dbTableInfo.size() > 0){
                        //extra table columns than storage columns
                        _**if(!preserve){
                            gottaCreateTable = true;
                        }else{                           
                            backup_table(db);
                        }**_
                    }
                }
                if(gottaCreateTable){
                    this->drop_table(this->table.name, db);
                    this->create_table(db, this->table.name);
                    res = decltype(res)::dropped_and_recreated;
                }
3. After calling sync_schema, storage and table has to be "synced". "already_in_sync" indicates that nothing is done in the function because nothing is needed.
4. Thanks.

fnc12:
1. ok
2. you're right but only if `preserve` is `false`. Otherwise `backup_table` fires and `res` is assigned with `columns_changed` (check latest commit - I've fixed it)
3. Just a naming problem. For me it is like iPhone syncing with iTunes - it syncs every time but if it was synced before that it is synced now. Thanks why enum values is called `synced`. But ok - we can change it to `already_in_sync` (don't forget to change `operator<<` with it)

Ninja-007:
Thanks,
I removed << operator

PR #32, opened 2017-06-08
Updated sync_schema return values

Ninja-007:
Updated sync_schema return values

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -2855,8 +2855,6 @@ namespace sqlite_orm {
                         this->create_table(db, this->table.name);
                         res = decltype(res)::dropped_and_recreated;                            
                     }
-                    else assert(0);
-
                 }
             }
             auto r = Super::sync_schema(db, preserve);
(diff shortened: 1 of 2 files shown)


fnc12:
`sync_schema_result` and `sync_schema_status` have the same meaning. Let's remove `sync_schema_status` and replace it with `sync_schema_result`

Ninja-007:
ok. Can you please make syny_schema_status publically available?

fnc12:
it's already public isn't it?

Ninja-007:
Yes, it is.

fnc12:
Why are you asking me to make it public if it is already public?

Ninja-007:
The separation between impl and public Storage class does confuse me sometimes. Sorry.

Ninja-007:
There is a difference between sync_schema_result and sync_schema_status. sync_schema_status tells that what should be done and sync_schema_result what is done.

fnc12:
@Ninja-007 yes but enum values maps to each other easily. It's better to write to docs that `schema_status` returns the same that `sync_schema` returns for concrete table but affects nothing

Ninja-007:
Ok, I leave it upto you.

PR #34, opened 2017-06-13
Included all table names in '' to main consistency

Ninja-007:
Some SQL queries where missing '' tables names. These create problem when table name is like expression(e.g. 1)


Ninja-007:
If you merge this pull request, you can add some test cases for table name like a number. BTW. SQLite support expression as tablename as long as it is ''.

PR #36, opened 2017-06-14
added get_table_list

Ninja-007:


Ninja-007:
Do you know what tests failed? I commited one method. But it is not used in tests, so there should not be runtime error because of it.

fnc12:
@Ninja-007 it isn't your fault - something fails in `insert.cpp` after I've added `insert_range`

fnc12:
sqlite version on my local machine (3.16.0) is higher that one in travis (3.7.9). We need to solve this

Ninja-007:
hmmm

fnc12:
@Ninja-007 travis installs sqlite 3.7.9 by default in ubuntu. Do you know how to specify 3.16.0 or 3.9 at least?

Ninja-007:
There are multiple ways. 
1. Using custom lib instead of preinstalled lib: If you have installation access to travis, you can install the required package or compile from source from and install. Then instead of using default sqlite lib, use the installed lib. Tip: You can use -Lpath_to_lib for static linking. For dynamic linking, you need to modify LD_LIBRARY_PATH also.
2. Adding sqlite locally in your project instead of default travis. I prefer this method. It reduces headache of compilation and install. You can easily upgrade

fnc12:
@Ninja-007 ok. I think a better way is to use precompiled binaries. Thanks for static linking tip. I was wondering how to perform it.
Why do you need `get_table_list`?

Ninja-007:
Good that my comment you found useful. 
I need to run some checks on each table of after start of application. At that moment, I do not know number of tables and names of tables. 
These tables are independent of each other but have same schema. They are mapped to same C++ class. Tables names are needed to create storage object.

fnc12:
Why do you have tables mapped to the same c++ class? Anyway for this case you must have several storages

Ninja-007:
Because they have same schema. I create storage locally when needed.

fnc12:
@Ninja-007 I don't understand your goals at all. Could you please show me more code please - I need to understand developers' goals before I commit the code. Thanks

Ninja-007:
It is simple. There are two application. One application creates databases and another uses the databases. The application developed by different teams. Errors or wrong data in the databases will cause crash in the user application.

/*!
\class Car
\brief Each Car will have separate table in database.
*/
class Car
{
    uint16_t mId; // unique id of car. Name of table is based on table. 
    uint8_t mColor; //primary key in table
    std::string mInfo;
}

auto DB = make_storage(DBPath);

std::vector<std::string> tableNames;
DB.get_list_tables(tableNames);
for(auto& table:tableNames)
{
    auto Storage =  make_storage(DBPath,
                        make_table(table,
                                   make_column("Color", &Car::mColor, primary_key()),
                                   make_column("Info", &Car::mInfo)
                                   )
                        );
    
    //conduct some checks on Storage e.g. Presence of some colors, total number of colors etc.
}

fnc12:
@Ninja-007 now I see. What you need is a dynamic storage. Dynamic means 'schema isn't known of not fully known at compile time'. What `sqlite_orm` offers is a static storage. It means 'whole schema is known at compile time'. That's it.

fnc12:
BTW why don't you store all cars in a single storage/table? Just add one more field to it like `model_id` or `app_id`

fnc12:
One type means one table. This is how relation databases work. You now you have incorrect database architecture.

Ninja-007:
Because of design decisions. For example, there is some readymade car data in csv format available which can be easily mapped with separate tables as compared to single table.

fnc12:
CSV can be imported to sqlite with `sqlite_orm` very simple without changing db architecture. I argue cause your way is not how many people use relational databases so `get_table_list` isn't required at all.

Ninja-007:
It is ok, if you do not commit it. There are some constraints due to which I cannot change DB design. get_table_list is general function. Developers can use it for other purposes as well.

fnc12:
But I can add `get_table_names` with `std::vector<std::string>` return type (non reference from argument list) with functionality you've written here. But your architecture is weird

Ninja-007:
That will work also, Thanks. Theory and practice is not always same. It is hard to bring all people on same page. But work must be finished.

fnc12:
Ok. Please change function name to `table_names` (not `table_list` cause function returns strings not table objects) and without `get_` notation. `get` and `set` notation makes code longer. Also you may omit `if(db)` check - `db` isn't null there. And please add return type instead of first argument. And I'll accept PR. Thanks

PR #38, opened 2017-06-16
added sqlite3_db_release_memory

Ninja-007:
added sqlite3_db_release_memory for memory constrainted applications

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -5152,11 +5152,11 @@ namespace sqlite_orm {
         }
         
         /**
-        * \fn release_memory
+        * \fn db_release_memory
         * \brief Releases freeable memory of database. It is function can/should be called periodically by application,
         * if application has less memory usage constraint.
         */
-        int release_memory() {
+        int db_release_memory() {
             std::shared_ptr<database_connection> connection;
             sqlite3 *db;
             if(!this->currentTransaction){
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -5151,10 +5151,12 @@ namespace sqlite_orm {
             }
         }
         
+#indef SQLITE_VERSION_NUMBER > 3007010
         /**
         * \fn db_release_memory
         * \brief Releases freeable memory of database. It is function can/should be called periodically by application,
         * if application has less memory usage constraint.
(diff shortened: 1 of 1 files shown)


Ninja-007:
It is available here. int sqlite3_db_release_memory(sqlite3*);
Is this problem because of old sqlite3 version?

fnc12:
travis fails. Please fix it. And it is better to call this function `db_release_memory` cause there is another function `release_memory` ([info](https://sqlite.org/c3ref/db_release_memory.html)) and there can be misunderstanding

fnc12:
Doesn't matter the version - if function exists in `libsqlite3` then we can add interface for in in the lib. If user wants to know what sqlite version he/she uses he/she can call `libversion()` member function in storage.

fnc12:
two options to fix travis:
1) add compile time check for sqlite version (or just macros to enable/disable the newest functions)
2) increase sqlite version in travis but it doesn't help users with older sqlite version.
What you think?

Ninja-007:
include/sqlite_orm/sqlite_orm.h:5168:20: error: use of undeclared identifier

      'sqlite3_db_release_memory'; did you mean 'sqlite3_release_memory'?

            return sqlite3_db_release_memory(db);

Ninja-007:
First approach has disadvantage in terms of maintainence. After few releases from SQLITE, the code may end up in many ifdefs. People get code of this library. They can comment out or delete the code.

Second approach is better, it keeps up with latest sqlite version. 

First approach is OK for time being. But it will be problematic in long run.

fnc12:
I think we need to use [this](https://sqlite.org/c3ref/c_source_id.html) macro. This is how other libraries work that depend on new c++ standard features

Ninja-007:
Yes, it is possible. What is travis SQLITE_VERSION_NUMBER? 
BTW: Did you mention anywhere in document about minimum sqlite version of this lib? If no, you should.

fnc12:
`SQLITE_VERSION_NUMBER` is not about travis. Functions in the lib must be inclosed in `#ifdef SQLITE_VERSION_NUMBER < ?`

Ninja-007:
The sqlite3_db_release_memory() added in  2012-01-16 (3.7.10). https://sqlite.org/changes.html

Ninja-007:
BTW, travis sqlite is from 2012, very old!!

fnc12:
yes, it is very bad. I'm gonna fix it - download binary for linux in tar, untar it and link with app binary

PR #39, opened 2017-06-16
added pp verion check

Ninja-007:


fnc12:
`>=` not `>`

fnc12:
let's wait for travis and of everything is ok I'll merge it and uncomment `char()` function and add version check for it too

fnc12:
please run tests before commiting (tests.cpp). You can also have different targets in your IDE for every example (like I do in Xcode). BTW that's your IDE?

Ninja-007:
QtCreater.

PR #65, opened 2017-11-22
like_t condition to binary_condition. Now it compatible with OR operator

danilova:
Example of usage:
```c++
storage.get_all<A> (
    left_join<B>(on( is_equal(&B::col_2, &A::col_1))),
    where ( 
        like (c( &A::col_2), "search_strirng")
        or
        like (c( &B::col_3), "search_strirng")
    )
)
```

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -1016,9 +1016,11 @@ namespace sqlite_orm {
         };
         
         template<class A, class T>
-        struct like_t {
+        struct like_t: public binary_condition<A, T> {
             A a;
             T t;
+
+            using binary_condition<A, T>::binary_condition;
             
             operator std::string() const {
                 return "LIKE";
@@ -1460,6 +1462,13 @@ namespace sqlite_orm {
     conditions::like_t<A, T> like(A a, T t) {
         return {a, t};
     }
+
+    template<class T, class R>
+    conditions::like_t<T, R> like(internal::expression_t<T> expr, R r) {
+        return {expr.t, r};
+    }
+
+
     
     namespace core_functions {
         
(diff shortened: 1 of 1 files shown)


fnc12:
let me check this code out. BTW `like (c( &A::col_2), "search_strirng")` could be replaced with `like (c&A::col_2, "search_strirng")`

fnc12:
@danilova `c(...)` function is required only for operator overloads using. E.g. `where(c(&User::id) == 5)`

fnc12:
You removed 
```c++
template<class A, class T>		
std::string process_where(conditions::like_t<A, T> &l);
```
that's not correct cause `like` is an operator first of all. Of course there is a function `like` that acts just like operator but I got a better idea of performing what you want. I'll create a class called `binary_operator` that also will be applied to operator overloading just like `binary_condition` does. `binary_operator`'s purpose is to handle binary operators like `+`, `-` etc.

fnc12:
sorry but I got to refuse your pull request cause it has some logical errors. Your idea is cool and I've added or/and support for `like` in the latest commit. But `like` isn't binary condition. It is operator firstly (and is function afterwards). I'm gonna add arithmetic operators support and they also will require or/and operators support. So there will be `condition_t` analog struct called `operator_t` I guess (or I'll just use `condition_t` for all cases). I mean if you need to add or/and operator support to some class you have to derive it from `condition_t`:
```c++
template<
        class L,
        class R,
        typename = typename std::enable_if<std::is_base_of<conditions::condition_t, L>::value && std::is_base_of<conditions::condition_t, R>::value>::type
    >
    conditions::and_condition_t<L, R> operator &&(const L &l, const R &r) {
        return {l, r};
    }
    
    template<
        class L,
        class R,
        typename = typename std::enable_if<std::is_base_of<conditions::condition_t, L>::value && std::is_base_of<conditions::condition_t, R>::value>::type
    >
    conditions::or_condition_t<L, R> operator ||(const L &l, const R &r) {
        return {l, r};
    }
```

Your previous PR was merged. Thanks for contributing. Please feel free to open issues here or contact me with email.

PR #80, opened 2018-01-22
CMake for examples and tests.

DerShokus:


DerShokus:
`left_and_inner_join` fails. It wasn't called in the .travis.yml. Should it be excluded from the script?

fnc12:
@DerShokus yes it should

DerShokus:
@fnc12 fixed

fnc12:
Thanks. CMake is what I've been missing

fnc12:
@DerShokus can you please fix `#include <codecvt>` missing error form here https://travis-ci.org/fnc12/sqlite_orm ? Looks like we need to change compiler or std lib from gcc to clang in CI. Thanks

PR #104, opened 2018-02-24
Updated test and example include paths to fix #103

### LINKED ISSUE #103: Tests and examples bad include path
I am building from the root directory with:
```
cmake .
make
```
I'm getting the error
```
fatal error: 'sqlite_orm.h' file not found
#include "sqlite_orm.h"
```

I believe this is because in the root `CMakeLists.txt`, line 7, the include directory is set to `${CMAKE_CURRENT_SOURCE_DIR}/include`, so examples and tests should be trying to include `sqlite_orm/sqlite_orm.h`.  When I change the inclusions to this, it works.

I'm happy to make the necessary change and make a pull request, but I'm wondering, is it better to change the inclusion path in the example and test files, or is it better to change the path in the root `CMakeLists.txt` file?

jeresch:
Pull request to go with issue #103.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,53 +15,53 @@ script:
   - cp -r sqlite-amalgamation-3190300/* sqlite_amalgamation
   - rm sqlite-amalgamation-3190300.zip
   - clang -c sqlite-amalgamation-3190300/sqlite3.c -o sqlite.static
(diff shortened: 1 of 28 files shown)


fnc12:
everything is nice but please change `#include "sqlite_orm/sqlite_orm.h"` to `#include <sqlite_orm/sqlite_orm.h>` - sqlite_orm is a library

jeresch:
Hopefully this is better!

PR #135, opened 2018-04-05
add_library(... INTERFACE), std::iterator_traits fixes

gardell:
Hello! Thanks for a great library! I had some minor issues with it I'd like to discuss, some of these changes might be more relevant than others, but here it goes!

## add_library(... INTERFACE)
By declaring an `INTERFACE` library (header-only),
the `include` directory can be specified once with the library
instead of with each user of it (`examples`, `tests` etc).

## std::iterator_traits fixes
I got some compiler errors in `iteration.cpp` using `g++7`,
since the `storage_t::view_t::iterator_t` seem to lack the standard
iterator `typedef`s like `value_type` etc.
http://en.cppreference.com/w/cpp/iterator/iterator_traits
specifies the fields needed. As can be read there:
```
If Iterator does not have all five member types difference_type, value_type, pointer, reference, and iterator_category, then this template has no members by any of those names (std::iterator_traits is SFINAE-friendly)
```
which means that all 5 need to be declared.
I am tempted to use `std::iterator` (see http://en.cppreference.com/w/cpp/iterator/iterator)
which would save a few lines, but as it is deprecated with C++17 I decided not to.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,14 @@
 cmake_minimum_required (VERSION 3.2)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 project(sqlite_orm)
 
 option(SqliteOrm_BuildTests "Build sqlite_orm unit tests" ON)
 
 set(SqliteOrm_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/include")
+add_library(sqlite_orm INTERFACE)
+target_include_directories(sqlite_orm INTERFACE include)
 
 # Tests
 include(CTest)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -4,11 +4,5 @@ file(GLOB files "*.cpp")
 foreach(file ${files})
     get_filename_component(file_basename ${file} NAME_WE)
     add_executable(${file_basename} ${file})
-    set_target_properties(${file_basename} PROPERTIES
-            CXX_STANDARD 14
-            CXX_STANDARD_REQUIRED ON
-    )
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++ -lc++abi")
-    target_include_directories(${file_basename} PRIVATE ${SqliteOrm_INCLUDE})
-    target_link_libraries(${file_basename} PRIVATE dl sqlite3 pthread)
+    target_link_libraries(${file_basename} PRIVATE sqlite_orm sqlite3)
 endforeach()
--- a/tests/third_party/sqlite/CMakeLists.txt
+++ b/tests/third_party/sqlite/CMakeLists.txt
@@ -16,3 +16,4 @@ add_custom_command(
     ${CMAKE_COMMAND} -E tar xfz ${SQLITE3_ARCH_NAME})
 
 add_library(sqlite3 sqlite-amalgamation-3210000/sqlite3.c)
+target_link_libraries(sqlite3 dl pthread)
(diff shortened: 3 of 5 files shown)


CMakeLists.txt
```diff
@@ -1,10 +1,14 @@
 cmake_minimum_required (VERSION 3.2)
+set(CMAKE_CXX_STANDARD 14)
```
gardell: I put this once in root instead of under `examples` and `tests`, since it applies not only to them, but to the actual library as well

CMakeLists.txt
```diff
@@ -1,10 +1,14 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 project(sqlite_orm)
 
 option(SqliteOrm_BuildTests "Build sqlite_orm unit tests" ON)
 
 set(SqliteOrm_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/include")
+add_library(sqlite_orm INTERFACE)
+target_include_directories(sqlite_orm INTERFACE include)
```
gardell:
```
set(SqliteOrm_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/include")
```
is never used anymore, since `examples` and `tests` now just depend on `sqlite_orm`, which in turn declares its `include`-directory. However, I decided to keep this line since users of your library might depend on it. Not sure if you'd want to deprecate it though?

examples/CMakeLists.txt
```diff
@@ -4,11 +4,5 @@ file(GLOB files "*.cpp")
 foreach(file ${files})
     get_filename_component(file_basename ${file} NAME_WE)
     add_executable(${file_basename} ${file})
-    set_target_properties(${file_basename} PROPERTIES
-            CXX_STANDARD 14
-            CXX_STANDARD_REQUIRED ON
-    )
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++ -lc++abi")
```
gardell: The `-stdlib=libc++` gave me compiler errors since it was not understood by my compiler, not sure what it is trying to accomplish?

tests/third_party/sqlite/CMakeLists.txt
```diff
@@ -16,3 +16,4 @@ add_custom_command(
     ${CMAKE_COMMAND} -E tar xfz ${SQLITE3_ARCH_NAME})
 
 add_library(sqlite3 sqlite-amalgamation-3210000/sqlite3.c)
+target_link_libraries(sqlite3 dl pthread)
```
gardell: Since it's actually `sqlite3` that depend on `dl` and `pthread`. So these dependencies have been removed from `example` and `tests` above.

examples/CMakeLists.txt
fnc12: I think you better talk to previous cmake commiter cause I know nothing about it - I don't use cmake and merged his PR just because some people asked about it

fnc12:
everything is ok, I'm ready to merge but please create a PR into `dev` branch, not `master`

examples/CMakeLists.txt
gardell: Good idea, I was looking at the commit history for that file and found you with https://github.com/fnc12/sqlite_orm/commit/b0dca907355013e5d9580fba19306da4dfba036b but the rest of the file is written by @DerShokus . I assume it's them I should ask. So, @DerShokus what is your opinion on the matter? Thanks

fnc12: yes, @DerShokus created cmake scripts. Also please change the target branch to `dev` and I'll merge the PR. Thanks. Your problem with iterators appeared because my CI service runs only on mac. I set it up like this cause ubuntu is too old for C++14 there. It is better to have CI for win, mac, ios, android, linux, orange pi. I'm thinking about this

gardell:
Thanks for merging this PR! Sorry I forgot to change the branch to dev!

PR #247, opened 2019-03-13
Prefer unique_ptr over shared_ptr

mrkline:
Most scenarios don't need shared ownership of data, so we should prefer unique_ptr when possible. It's more efficient, doesn't require atomic ops for a reference count (which can cause major slowdowns on weakly-ordered platforms like ARM), and can be promoted by the user to a
shared_ptr as-needed. (Conversely, you _can't_ go from shared back to unique.)

This replaces all instances of shared_ptr with unique_ptr except for two
cases:

1. iterator_t's current item is shared among copies,
   as move-only iterators aren't currently allowed.
   (http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0902r0.html)

2. The internal::connection returned by get_or_create_connection() is
   shared among those who call that function.

Another interesting step would be to use optional<T>, either from Boost
or from C++17, to avoid heap allocations where they aren't necessary.

See also:
https://herbsutter.com/2013/05/29/gotw-89-solution-smart-pointers/

I understand that this would be a fairly significant change to the public API, so let me know how you'd like to handle that if you like the changes.

### DIFF
--- a/examples/update.cpp
+++ b/examples/update.cpp
@@ -30,10 +30,10 @@ inline auto initStorage(const std::string &path) {
 
 using Storage = decltype(initStorage(""));
 
-static std::shared_ptr<Storage> stor;
+static std::unique_ptr<Storage> stor;
 
 int main(int argc, char **argv) {
-    stor = std::make_shared<Storage>(initStorage("update.sqlite"));
+    stor = std::make_unique<Storage>(initStorage("update.sqlite"));
     stor->sync_schema();
     stor->remove_all<Employee>();
 
--- a/README.md
+++ b/README.md
@@ -121,10 +121,10 @@ try{
 }
 ```
 
-Probably you may not like throwing exceptions. Me too. Exception `std::system_error` is thrown because return type in `get` function is not nullable. You can use alternative version `get_no_throw` which returns `std::unique_ptr` and doesn't throw `not_found_exception` if nothing found - just returns `nullptr`.
+Probably you may not like throwing exceptions. Me too. Exception `std::system_error` is thrown because return type in `get` function is not nullable. You can use alternative version `get_pointer` which returns `std::unique_ptr` and doesn't throw `not_found_exception` if nothing found - just returns `nullptr`.
 
 ```c++
-if(auto user = storage.get_no_throw<User>(insertedId)){
+if(auto user = storage.get_pointer<User>(insertedId)){
     cout << "user = " << user->firstName << " " << user->lastName << endl;
 }else{
     cout << "no user with id " << insertedId << endl;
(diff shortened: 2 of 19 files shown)


fnc12:
Hello. Thanks for such a brilliant PR. I thought many times about changing `std::shared_ptr` to something that doesn't use shared memory logic cause it is not necessary in most cases. The reason why `std::shared_ptr` is almost everywhere - I did not know exactly where I may need shared memory in future when I was just creating the lib. I did not touch it cause I'm waiting for optional from C++17 - it doesn't require heap allocation as you mentioned. 
I'd like to merge your PR but it has a little problem - it breaks backward compatibility cause it changes public API in `get_no_throw` function and some aggregation functions. If we do so we need to increment major version number of the lib I think it would be better to leave pubic API the same but commit all other internal changes cause they optimize usage of the lib.
I think it would be better to update public API with one mote template argument which can help to specify return type and leave `std::shared_ptr` default. I think I can do it some day or you if you want.
Thanks again

mrkline:
Thanks for the warm reception! And sorry for the delay in my response - I've been busy the last few weeks.

To preserve the existing API, I'm pushing a commit that makes `get_no_throw()` return a `shared_ptr` again. This patch also adds a new function, `get_pointer()`, to return the `unique_ptr`. If you'd rather name it something else, let me know, but I think most people would prefer a new function with a its own name instead of adding another template argument. `get_no_throw<User, unique_ptr>` (or something similar) seems much more cumbersome than `get_pointer<User>`.

This leaves a few user-facing functions such as `min()` and `max()` returning `unique_ptr` instead of `shared_ptr`, but because of how interoperable the two pointer types are, I don't think this is a large problem. Note that the relevant tests (like `testDifferentGettersAndSetters()` haven't had to change at all! Use seems to fall into one of two camps:

1. The result is used immediately, e.g., `if (*storage0.sum(&User::id) > 42) ...` or stored in an `auto` variable, e.g., `auto sum = storage0.sum(&User::id);`. In both of these cases, type deduction will take care of everything.

2. The result is stored in a `shared_ptr`, e.g., `shared_ptr<int> sum = storage0.sum(&User::id)`, but since you can construct a `shared_ptr` from a a `unique_ptr&&`, the code continues to compile.

Users' code would only stop working if they stored the result in an `auto` variable, then tried to assign it to a `shared_ptr`:
```
auto sum  = storage0.sum(&User::id);
shared_ptr<int> sum2 = sum; // compile-error
```
I can't imagine many people mixing paradigms like that.

Let me know what you think, and thanks again.

examples/update.cpp
```diff
@@ -47,7 +47,7 @@ int main(int argc, char **argv) {
     stor->replace(Employee{ 7, "James", 24, "Houston", 10000.0 });
 
     //  show 'COMPANY' table contents
-    for(auto &employee : stor->iterate<Employee>()) {
+    for(auto&& employee : stor->iterate<Employee>()) {
```
fnc12: why do you use rvalue ref here?

mrkline: Ah - that's a vestige from when I was trying to get the `shared_ptr` out of `iterator_t`, before I realized that C++ doesn't currently alow move-only iterators. I'll back it out.

fnc12: Thanks. Once it is done I'll be ready to merge the PR

mrkline: Done! See 7304178

PR #248, opened 2019-03-20
Use getter to access db from storage

oost:
Not sure how to build dev into single header but fixed access in both places


fnc12:
Hi. It is a bug I want to fix a lot of time but cannot find time for it. Thanks for fixing it.
BTW to generate `sqlite_orm.h` go to root dir of the repo and run `python third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .`

PR #254, opened 2019-03-30
Add cmake config module. (fixes #249)

### LINKED ISSUE #249: Add CMake config module
ITNOA

Hi,

Thanks for great library, I suggest to add [CMake config module](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html) to found this library easily in projects that uses CMake build system with [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html#id3) command.

if you agree with me, I can provide PR for this.

soroshsabz:
ITNOA

This PR resolved #249 and added cmake config module for work properly with another cmake projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,13 +2,33 @@ cmake_minimum_required (VERSION 3.2)
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-project(sqlite_orm)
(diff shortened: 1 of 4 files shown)


soroshsabz:
@fnc12 appveyor does not work, but I think it is not because my patch, Are sure appveyor work before my PR?

thanks

fnc12:
@soroshsabz Thanks for the PR. Appveyor did not work at all cause it is just connected not configured.

PR #259, opened 2019-04-01
Add header file to install path. (see #249)

soroshsabz:
ITNOA

This PR add sqlite_orm.h to install path, so when run `make install` command install header file in system.


fnc12:
thanks

PR #280, opened 2019-04-12
Update README.md

soroshsabz:
Add cmake instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -732,6 +732,26 @@ For more details please check the project [wiki](https://github.com/fnc12/sqlite
 
 Use popular package manager like [vcpkg](https://github.com/Microsoft/vcpkg) and just install it with `vcpkg install sqlite_orm` command.
 
+Or you can use below instructions
+
+```bash
+git clone https://github.com/fnc12/sqlite_orm.git sqlite_orm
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -732,6 +732,26 @@ For more details please check the project [wiki](https://github.com/fnc12/sqlite
...
+git clone https://github.com/fnc12/sqlite_orm.git sqlite_orm
+cd sqlite_orm
+mkdir compile
+cd compile
+cmake ..
+cmake --build .
+sudo make install
+```
+
+then you can just include `sqlite_orm.h` that install in system-wide header files location or if you use cmake build system you can just add below commands in CMakeLists.txt
```
fnc12: 1) `that is installed` instead of `that install`
2) `in case you use cmake build` instead of `if you use cmake build`

fnc12:
this is a great PR. Thanks for contributing. Please apply fixes I wrote about in comments and I'm ready to merge it

README.md
soroshsabz: @fnc12 I changed it

soroshsabz:
@fnc12 Thanks for reviewing, I resolve your review :)

PR #284, opened 2019-04-24
Various fixes for GCC 8.3 and Clang 8 warnings

mrkline:
The project I'm using sqlite_orm for is built with fairly stringent
warnings (-Wall -Wextra -Wpointer-arith -Wconversion -Wsign-conversion
-Wno-maybe-uninitialized -Werror). This resolves warnings that came out
of that set, mostly related to:

- Sign conversion errors, especially regarding size_t <-> int.
  To preserve the existing API, all count() functions continue to return
  int, but this requires casts whenever you compare those to STL .size()
  calls.

- Unused args

- Catching exceptions by value

- Unreachable locations - I'm currently calling assert(false) to
  convince the compiler that we don't reach the end of some functions
  (and exit without returning a value).
  Let me know if something else is preferred - e.g. an UNREACHABLE macro
  that resolves to __builtin_unreachable() in GCC and Clang.

### DIFF
--- a/dev/constraints.h
+++ b/dev/constraints.h
@@ -1,6 +1,5 @@
 #pragma once
 
-#include <cassert>
 #include <string>   //  std::string
 #include <tuple>    //  std::tuple, std::make_tuple
 #include <sstream>  //  std::stringstream
--- a/dev/constraints.h
+++ b/dev/constraints.h
@@ -315,7 +315,7 @@ namespace sqlite_orm {
                     case decltype(argument)::nocase: return "NOCASE";
                     case decltype(argument)::rtrim: return "RTRIM";
                 }
-                assert(false);
+                throw std::domain_error("Invalid collate_argument enum");
             }
         };
         
--- a/dev/constraints.h
+++ b/dev/constraints.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cassert>
 #include <string>   //  std::string
 #include <tuple>    //  std::tuple, std::make_tuple
 #include <sstream>  //  std::stringstream
@@ -314,6 +315,7 @@ namespace sqlite_orm {
                     case decltype(argument)::nocase: return "NOCASE";
                     case decltype(argument)::rtrim: return "RTRIM";
                 }
+                assert(false);
             }
         };
         
--- a/dev/row_extractor.h
+++ b/dev/row_extractor.h
@@ -147,7 +147,7 @@ namespace sqlite_orm {
         std::vector<char> extract(const char *row_value) {
             if(row_value){
                 auto len = ::strlen(row_value);
-                return this->go(row_value, static_cast<int>(len));
+                return this->go(row_value, len);
             }else{
                 return {};
             }
@@ -161,7 +161,7 @@ namespace sqlite_orm {
         
     protected:
         
-        std::vector<char> go(const char *bytes, int len) {
+        std::vector<char> go(const char *bytes, size_t len) {
             if(len){
                 std::vector<char> res;
                 res.reserve(len);
@@ -209,7 +209,7 @@ namespace sqlite_orm {
         std::vector<char> extract(const char *row_value) {
             if(row_value){
                 auto len = ::strlen(row_value);
-                return this->go(row_value, static_cast<int>(len));
+                return this->go(row_value, len);
             }else{
                 return {};
             }
@@ -217,13 +217,13 @@ namespace sqlite_orm {
         
         std::vector<char> extract(sqlite3_stmt *stmt, int columnIndex) {
             auto bytes = static_cast<const char *>(sqlite3_column_blob(stmt, columnIndex));
-            auto len = sqlite3_column_bytes(stmt, columnIndex);
+            auto len = static_cast<size_t>(sqlite3_column_bytes(stmt, columnIndex));
             return this->go(bytes, len);
         }
         
     protected:
         
-        std::vector<char> go(const char *bytes, int len) {
+        std::vector<char> go(const char *bytes, size_t len) {
             if(len){
                 std::vector<char> res;
                 res.reserve(len);
--- a/examples/blob.cpp
+++ b/examples/blob.cpp
@@ -14,7 +14,7 @@ struct User {
     std::vector<char> hash; //  binary format
 };
 
-int main(int argc, char **argv) {
+int main(int, char **) {
     using namespace sqlite_orm;
     auto storage = make_storage("blob.sqlite",
                                 make_table("users",
--- a/examples/enum_binding.cpp
+++ b/examples/enum_binding.cpp
@@ -1,5 +1,7 @@
 
 #include <sqlite_orm/sqlite_orm.h>
+
+#include <cassert>
 #include <iostream>
 #include <memory>
 
@@ -30,6 +32,7 @@ std::string GenderToString(Gender gender) {
         case Gender::Female:return "female";
         case Gender::Male:return "male";
     }
+    assert(false); // unreachable
 }
 
 /**
--- a/examples/enum_binding.cpp
+++ b/examples/enum_binding.cpp
@@ -1,7 +1,6 @@
 
 #include <sqlite_orm/sqlite_orm.h>
 
-#include <cassert>
 #include <iostream>
 #include <memory>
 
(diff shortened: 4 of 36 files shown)


dev/row_extractor.h
```diff
@@ -209,21 +209,21 @@ namespace sqlite_orm {
...
+                return this->go(row_value, len);
             }else{
                 return {};
             }
         }
         
         std::vector<char> extract(sqlite3_stmt *stmt, int columnIndex) {
             auto bytes = static_cast<const char *>(sqlite3_column_blob(stmt, columnIndex));
-            auto len = sqlite3_column_bytes(stmt, columnIndex);
+            auto len = static_cast<size_t>(sqlite3_column_bytes(stmt, columnIndex));
```
fnc12: are you sure that casting from `int` to `size_t` produces a warning?

examples/enum_binding.cpp
```diff
@@ -30,6 +32,7 @@ std::string GenderToString(Gender gender) {
         case Gender::Female:return "female";
         case Gender::Male:return "male";
     }
+    assert(false); // unreachable
```
fnc12: Better use `throw`

examples/exists.cpp
```diff
@@ -77,18 +77,18 @@ int main(int argc, char **argv) {
...
-    storage.replace(Agent{"A011", "Ravi Kumar", "Bangalore", 0.15, "077-45625874"});
-    storage.replace(Agent{"A010", "Santakumar", "Chennai", 0.14, "007-22388644"});
-    storage.replace(Agent{"A012", "Lucida", "San Jose", 0.12, "044-52981425"});
-    storage.replace(Agent{"A005", "Anderson", "Brisban", 0.13, "045-21447739"});
-    storage.replace(Agent{"A001", "Subbarao", "Bangalore", 0.14, "077-12346674"});
-    storage.replace(Agent{"A002", "Mukesh", "Mumbai", 0.11, "029-12358964"});
-    storage.replace(Agent{"A006", "McDen", "London", 0.15, "078-22255588"});
-    storage.replace(Agent{"A004", "Ivan", "Torento", 0.15, "008-22544166"});
-    storage.replace(Agent{"A009", "Benjamin", "Hampshair", 0.11, "008-22536178"});
+    storage.replace(Agent{"A007", "Ramasundar", "Bangalore", 0.15, "077-25814763", ""});
```
fnc12: what warning does this fix removes?

include/sqlite_orm/sqlite_orm.h
```diff
@@ -5340,28 +5343,26 @@ namespace sqlite_orm {
                     using member_pointer_t = typename column_type::member_pointer_t;
                     using getter_type = typename column_type::getter_type;
                     using setter_type = typename column_type::setter_type;
-
-					// GCC has bug that reported in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64095 so static_if must have at lease one input
                     if(!res){
-                        static_if<std::is_same<C, member_pointer_t>{}>([&res, &obj, &col](const C& c){
```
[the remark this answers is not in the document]
mrkline: I didn't make any manual changes to the single-header `sqlite_orm.h` - I just ran the amalgamation script to update it from the changes in `dev/`. It seems like e422eb2976f101419299ae6146473e15bacf135c changed it here instead of in `dev/table.h` there. I'll update `table.h` to prevent this from happening again.

dev/row_extractor.h
mrkline: With `-Wsign-conversion`, yes: https://godbolt.org/z/nYqz5V

C++ implicitly promotes signed types to unsigned, but this can cause some unexpected behavior (https://www.youtube.com/watch?v=wvtFGa6XJDU), which is why some projects build with that warning.

examples/exists.cpp
mrkline: Recent compilers complain if field initializers are used for some fields but not all:
```
sqlite_orm/examples/exists.cpp:80:82: error: missing initializer for member ‘Agent::country’ [-Werror=missing-field-initializers]
     storage.replace(Agent{"A007", "Ramasundar", "Bangalore", 0.15, "077-25814763"});
                                                                                  ^
```

examples/enum_binding.cpp
mrkline: Done.

dev/constraints.h
```diff
@@ -314,6 +315,7 @@ namespace sqlite_orm {
                     case decltype(argument)::nocase: return "NOCASE";
                     case decltype(argument)::rtrim: return "RTRIM";
                 }
+                assert(false);
```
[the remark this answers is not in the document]
mrkline: Done.

include/sqlite_orm/sqlite_orm.h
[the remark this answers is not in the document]
fnc12: Yes, it is exactly what I mean. Thank you

examples/enum_binding.cpp
```diff
@@ -1,5 +1,7 @@
 
 #include <sqlite_orm/sqlite_orm.h>
+
+#include <cassert>
```
fnc12: This include is not required now

examples/nullable_enum_binding.cpp
```diff
@@ -1,4 +1,5 @@
 
+#include <cassert>
```
fnc12: Not required

include/sqlite_orm/sqlite_orm.h
```diff
@@ -400,6 +400,7 @@ namespace sqlite_orm {
 }
 #pragma once
 
+#include <cassert>
```
fnc12: Please remove

include/sqlite_orm/sqlite_orm.h
```diff
@@ -5340,8 +5343,8 @@ namespace sqlite_orm {
                     using member_pointer_t = typename column_type::member_pointer_t;
                     using getter_type = typename column_type::getter_type;
                     using setter_type = typename column_type::setter_type;
-
-					// GCC has bug that reported in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64095 so static_if must have at lease one input
+                    // Make static_if have at least one input as a workaround for GCC bug:
```
fnc12: A very important comment. Thank you

examples/blob.cpp
```diff
@@ -3,7 +3,6 @@
 #include <string>
 #include <vector>
 #include <iostream>
-#include <cassert>
```
fnc12: If you did not add `#include <cassert>` into this file then you don't need to delete it. It is required cause `assert` is used in this file. I asked to only remove includes which were added by you and became unused

mrkline: Sure, but is there any reason to keep these includes around if they're not used by the examples?

mrkline: If you would prefer a separate PR to remove unneeded `includes`, I've updated this one so that the ones in `examples/` are left unchanged.

fnc12: This include is needed: `assert` macro is used 5 times in this source file

dev/constraints.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cassert>
```
fnc12: Please remove

examples/blob.cpp
mrkline: Hm, something else must have been pulling it in, as the example continued to build without it (both on my machine and in CI).

dev/constraints.h
mrkline: Done, sorry for missing this one - I was just looking in `examples/` before.
As best I can tell, this PR no longer contains any new `#include <cassert>`:
```
$ git df origin/dev HEAD | grep cassert
<no output>
```

examples/blob.cpp
fnc12: Yes, it builds cause `<cassert>` is included somewhere else in another included header implicitly. But we must not rely on it cause it is not guaranteed by standard. You also can use `std::string` in a source which does not have `#include <string>` but has `#include <map>` on OS X clang but this code will not compile on Linux GCC.

fnc12:
Thanks

mrkline:
Thanks for your help and patience!

PR #285, opened 2019-04-26
refactored process conditions

fnc12:


dev/view.h
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include <memory>   //  std::shared_ptr
+#include <string>   //  std::string
+#include <utility>  //  std::forward, std::move
+#include <sqlite3.h>
```
soroshsabz: ITNOA

I think it is good to change order include non standard library from standard library

dev/view.h
```diff
@@ -0,0 +1,61 @@
...
+            view_t(storage_type &stor, decltype(connection) conn, Args&& ...args_):
+            storage(stor),
+            connection(std::move(conn)),
+            args(std::make_tuple(std::forward<Args>(args_)...)){}
+            
+            size_t size() {
+                return this->storage.template count<T>();
+            }
+            
+            bool empty() {
```
soroshsabz: I think empty can be right const function

dev/view.h
```diff
@@ -0,0 +1,61 @@
...
+            storage_type &storage;
+            std::shared_ptr<internal::database_connection> connection;
+            std::tuple<Args...> args;
+            
+            view_t(storage_type &stor, decltype(connection) conn, Args&& ...args_):
+            storage(stor),
+            connection(std::move(conn)),
+            args(std::make_tuple(std::forward<Args>(args_)...)){}
+            
+            size_t size() {
```
soroshsabz: size can be right const function

dev/view.h
```diff
@@ -0,0 +1,61 @@
...
+                return this->storage.template count<T>();
+            }
+            
+            bool empty() {
+                return !this->size();
+            }
+            
+            iterator_t<self> end() {
```
soroshsabz: I think it is good to provide right const `end` and `begin` function and also provide `cend` and `cbegin` function for const access to `view`

tests/tests.cpp
```diff
@@ -12,6 +12,38 @@ using namespace sqlite_orm;
...
+    for (int i = 0; i < 255; i++){
+        key[i] = i;
+    }
+    
+    Test v;
+    v.key = key;
+    
+    db.insert(v);
+    
+    for(auto &obj : db.iterate<Test>()){
```
soroshsabz: Add one test for `auto const& obj`

tests/tests.cpp
```diff
@@ -12,6 +12,38 @@ using namespace sqlite_orm;
...
+        std::vector<char> key;
+    };
+    
+    auto db = make_storage("",
+                           make_table("Test",
+                                      make_column("key", &Test::key),
+                                      make_column("id", &Test::id, autoincrement(), primary_key())));
+    db.sync_schema(true);
+    
+    std::vector<char> key;
```
soroshsabz: use below code instead of this

```cpp
std::vector<int> v(255) ; // vector with 255 ints.
std::iota (std::begin(v), std::end(v), 0); // Fill with 0, 1, ..., 254.
```

https://stackoverflow.com/a/17694752/1539100

dev/view.h
fnc12: `empty` and `size` access storage which alters its state so these functions cannot be right const

PR #298, opened 2019-05-16
Some refactoring in statement. (see #296)

### LINKED ISSUE #296: [Bug] replace does not compile for private class members
ITNOA

If I add some little code in private_class_members.cpp in example like below

```cpp
	Player p = storage.get<Player>(soloId);
	soloId.setName("kadan");
	storage.replace(p)
```

I got compile error like below

```
sqlite_orm.h:9126:37: error: lvalue required as unary ‘&’ operand
                             value = &(((o).*(c.getter))());
```

soroshsabz:
ITNOA

This PR some refactor on storage

### DIFF
--- a/dev/error_code.h
+++ b/dev/error_code.h
@@ -18,6 +18,7 @@ namespace sqlite_orm {
         cannot_start_a_transaction_within_a_transaction,
         no_active_transaction,
         incorrect_journal_mode_string,
+        not_implemented
     };
     
 }
@@ -51,6 +52,8 @@ namespace sqlite_orm {
                     return "Cannot start a transaction within a transaction";
                 case orm_error_code::no_active_transaction:
                     return "No active transaction";
+                case orm_error_code::not_implemented:
+                    return "Not implemented";
                 default:
                     return "unknown error";
             }
--- a/dev/error_code.h
+++ b/dev/error_code.h
@@ -17,8 +17,7 @@ namespace sqlite_orm {
         table_has_no_primary_key_column,
         cannot_start_a_transaction_within_a_transaction,
         no_active_transaction,
-        incorrect_journal_mode_string,
-        not_implemented
+        incorrect_journal_mode_string
     };
     
 }
@@ -52,8 +51,6 @@ namespace sqlite_orm {
                     return "Cannot start a transaction within a transaction";
                 case orm_error_code::no_active_transaction:
                     return "No active transaction";
-                case orm_error_code::not_implemented:
-                    return "Not implemented";
                 default:
                     return "unknown error";
             }
--- a/dev/storage.h
+++ b/dev/storage.h
@@ -2547,8 +2547,8 @@ namespace sqlite_orm {
                     cols.for_each([&o, &index, &stmt, &impl] (auto &m) {
                         using column_type = typename std::decay<decltype(m)>::type;
                         using field_type = typename column_result_t<self, column_type>::type;
-                        const field_type value = impl.table.template get_object_field<field_type>(o, m);
-                        statement_binder<field_type>().bind(stmt, index++, value);
+                        const field_type* value = impl.table.template get_object_field_pointer<field_type>(o, m);
+                        statement_binder<field_type>().bind(stmt, index++, *value);
                     });
                     if (sqlite3_step(stmt) == SQLITE_DONE) {
                         return int(sqlite3_last_insert_rowid(connection->get_db()));
--- a/dev/storage.h
+++ b/dev/storage.h
@@ -2752,23 +2752,6 @@ namespace sqlite_orm {
                 return index;
             }
 
-            template<class O>
-            unsigned int fill_statement_for_each_non_primary_key_column(const O& o, unsigned int index, sqlite3_stmt* const stmt) {
-                auto& impl = get_impl<O>();
-                impl.table.for_each_column([&o, &index, &stmt, &impl] (auto c) {
-                    using field_type = typename decltype(c)::field_type;
-					if (!c.template has<constraints::primary_key_t<>>()) {
-                         if(c.member_pointer) {
-                             statement_binder<field_type>().bind(stmt, index++, (o.*c.member_pointer));
-                         } else {
-                             statement_binder<field_type>().bind(stmt, index++, ((o).*(c.getter))());
-                         }
-                    }
-                });
-
-                return index;
-            }
-
             template<class Func>
             void fill_statement(const std::string& query, Func func) {
                 auto connection = this->get_or_create_connection();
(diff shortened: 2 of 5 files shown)


soroshsabz:
@fnc12 travis error because #299

dev/storage.h
```diff
@@ -2843,7 +2737,87 @@ namespace sqlite_orm {
...
+            }
+
+            template<class O>
+            unsigned int fill_statement_for_each_non_primary_key_column(const O& o, unsigned int index, sqlite3_stmt* const stmt) {
+                auto& impl = get_impl<O>();
+                impl.table.for_each_column([&o, &index, &stmt, &impl] (auto c) {
+                    using field_type = typename decltype(c)::field_type;
+					if (!c.template has<constraints::primary_key_t<>>()) {
+                         if(c.member_pointer) {
+                             statement_binder<field_type>().bind(stmt, index++, (o.*c.member_pointer));
```
fnc12: does it work with composite key?

dev/table.h
```diff
@@ -52,46 +52,60 @@ namespace sqlite_orm {
...
+                return result;
             }
-            
+
+            /**
+             *  Function used to get field value from object by mapped member pointer/setter/getter
+             */
+            template<class F, class C, typename std::enable_if<!std::is_copy_assignable<F>::value, void*>::type = nullptr>
+            F get_object_field(const object_type &obj, C c) {
+                throw std::system_error(std::make_error_code(orm_error_code::not_implemented));
```
fnc12: So you cannot map a field which has not copy ctor? You break back compatibility with this PR cause one could do it before

dev/storage.h
soroshsabz: @fnc12 it is like before, just refactor

dev/storage.h
```diff
@@ -1439,43 +1439,8 @@ namespace sqlite_orm {
...
-                            if(c.member_pointer){
-                                value = &(o.*c.member_pointer);
-                            }else{
-                                value = &((o).*(c.getter))();
-                            }
-                            statement_binder<field_type>().bind(stmt, index++, *value);
-                        }
-                    });
-                    impl.table.for_each_column([&o, stmt, &index] (auto &c) {
-                        if(c.template has<constraints::primary_key_t<>>()) {
```
soroshsabz: @fnc12 How does this work for composite key?

dev/table.h
soroshsabz: @fnc12 Any suggestion for resolve this issue?

dev/storage.h
soroshsabz: @fnc12 Did you want to remove no_primary_key functions?

dev/table.h
soroshsabz: I don't think this code work for not copy assignable before this PR, because for example `*(&(unique_ptr<…>))` is meaningless for storing in DB

soroshsabz: @fnc12 If you want I can remove this function and back them to past form

soroshsabz:
@fnc12 I think it is another issue, so I change my commits and add only refactore, please review it 

thanks

dev/storage.h
fnc12: @soroshsabz Ok, but I do not understand you last comment

soroshsabz:
Which comment?

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dev/table.h
fnc12: Look, we have fixed bug in replace. Now you are talking about another bug which has not issue. If there is not issue and no description with sample reproduction code so the bug doesn't exist. I understand that you know the bug but you are the one. I cannot understand you until bug is reported as it must be reported. You reported a bug with replace - it is nice. You offered a PR. The PR did pass a review and I fixed the bug by myself. So you reported a bug - then the bug is fixed. Please report a bug you are talking right now cause I do not understand what are you talking about

soroshsabz:
@fnc12 I say remove all change that you don't like, so please review it

dev/error_code.h
```diff
@@ -18,6 +18,7 @@ namespace sqlite_orm {
         cannot_start_a_transaction_within_a_transaction,
         no_active_transaction,
         incorrect_journal_mode_string,
+        not_implemented
```
fnc12: where is it used?

dev/storage.h
```diff
@@ -1439,43 +1439,8 @@ namespace sqlite_orm {
...
-                    if (sqlite3_step(stmt) == SQLITE_DONE) {
-                        //  done..
-                    }else{
-                        throw std::system_error(std::error_code(sqlite3_errcode(connection->get_db()), get_sqlite_error_category()));
-                    }
-                }else {
-                    throw std::system_error(std::error_code(sqlite3_errcode(connection->get_db()), get_sqlite_error_category()));
-                }
+
+                fill_statement(o, ss.str());
```
fnc12: Unfortunately I cannot merge refactoring cause I am going to refactor these functions in another way. I have already implemented it once in a different project which uses custom `sqlite_orm`'s fork. I want to separate ORM and SQLite access logic to erase copy-paste totally and move some implementations to `.cpp` file

soroshsabz: I suggest use -inl.h pattern for this and I strongly recommend to separate implementation from interface.

but I suggest if this refactor does not bad for you and some time remain to done your project, you can accept this refactor to decreasing code duplication in project, but if you not , I can close pull request

what am I do?

dev/error_code.h
soroshsabz: @fnc12 Ok you right, I forgotten to remove this section from pas commits, but I think it is good error code and useful for future use, if you agree with me, we can keep this or if you disagree with me, I can remove it, so let me know your decision. :)

dev/storage.h
fnc12: I don't think that -inl.h pattern is good. It can be more comfortable for C++ coders but as I said before it is a very popular practice to have implementation in place in different popular languages. I understand that separated implementation reduces compile time for non template classes and I don't say we need to deny it if we would have it but templates have no advantages except readability for some developers. And as I said I cannot merge this refactoring cause I am planing to refactor these functions in a different way.

dev/error_code.h
fnc12: [KISS](https://ru.wikipedia.org/wiki/KISS_(принцип)) please remove it if you don't need it right now

soroshsabz: Ok I remove it

dev/storage.h
soroshsabz: @fnc12 Ok so close this PR if you have planning soon to refactor

soroshsabz: @fnc12 I close it, but I think it's better to accept it until you refactor whole of `storage` still

PR #300, opened 2019-05-17
fixed replace with private class members and added unit tests

fnc12:


soroshsabz:
@fnc12 you want to decline my PR?

soroshsabz:
@fnc12 It is not different to my PR in replace case (instead of I some refactoring to reduce code duplication)

But I want to resolve another problem in `get_object_field_pointer` in `insert` function that you say I break backward compatibility

fnc12:
I just wanted to fix replace bug as it reported in the issue

fnc12:
...I could not accept your fix cause it breaks `std::unique_ptr` field types

soroshsabz:
> ...I could not accept your fix cause it breaks std::unique_ptr field types

...

I change my PR to only refactor please review it

thanks

PR #325, opened 2019-06-24
TravisCI

Farwaykorse:
Cleanup of the Travis configuration.

Functional changes:
- Install Catch2 - usable from [CMake](github.com/catchorg/Catch2/blob/master/docs/cmake-integration.md).
- Let CMake call Ninja instead of Make (per build job speedup: 30-120 seconds).
- ~~Allow setting a different compiler to build the vcpkg executable.~~
  ~~This makes it possible to test the project with older compilers (and AppleClang).~~
- ~~Cache the vcpkg executable (build job speedup: 2-4 minutes).~~
- ~~Cache gtest (build job speedup: 20-50 seconds).~~

Add more configurations:
- Linux: GCC-9
- OSX: AppleClang-10.0.1.
- ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,7 +103,7 @@ before_script:
 # build examples, and run tests (ie make & make test)
 script:
   - cmake --build . --config Debug
-  - ctest --verbose --output-on-failure -C Debug
+  - ctest --verbose --output-on-failure -C Debug -j $(nproc)
 
 # generate and publish GCov coveralls results
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,13 @@ matrix:
       osx_image: xcode10.2
       language: cpp
       env:
-        - VCPKG_ALLOW_APPLE_CLANG: 1
+        - VCPKG_CXX: g++-6 # Compiler to build vcpkg
+        - SQLITE_ORM_OMITS_CODECVT: ON
+      addons:
+        homebrew:
+          packages:
+            - gcc@6
+          update: true
 
     - name: "OSX LLVM/Clang"
       os: osx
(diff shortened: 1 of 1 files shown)


fnc12:
Thanks for your work. Also I think gtest is not required at all cause right now it is not used and I am going to use `catch2` to make unit tests. So for now you can remove gtest or event replace it with catch2. But it is not a requirement

Farwaykorse:
Catch2 has a few different ways to install it:
- As a single header library to download and include in your project (note, there are some plans to deprecate it).
- The new multi-file version, that can be included via a packet manager and picked up by CMake etc.
  - From vcpkg
  - Directly from the git-repository.
  - ~~From Apt on Ubuntu~~ (stuck at the ancient Catch v1.2.0)
  - From homebrew for OSX


For now I add Catch2 from homebrew for OSX, that will be faster than building from source.
For the linux-builds I'll get it from GitHub and build from source.
On AppVeyor vcpkg comes pre-installed, so that would probably be the best choice there.

If you want to keep using vcpkg on Travis, you can revert the commit that removes it.

fnc12:
Everything is fine. Thank you

Farwaykorse:
**Question:** Shall I remove the coveralls code?
The current implementation is not working.

fnc12:
Yeah. If it is not used right now and doesn't exist in the nearest plans it has to be removed. Thanks

Farwaykorse:
I think this should do it.

Build time has been significantly reduced by the removal of unnecessary scripts and the use of Ninja, which does way better build parallelisation than Make.
Even with two extra jobs (one outside the concurrent set of 5) CPU time is ~33 min. and a commit checking roundtrip of ~15 min. (when there are no other commits).
Up to 4 more jobs could be added while keeping within this 15 min. roundtrip time.

It should now be relatively easy to add more configurations to test with.
I added GCC-9 since it gives more and nicer warnings, and AppleClang because it is the default on OSX.

fnc12:
Thanks again. This PR is a good improvement of `sqlite_orm`

Farwaykorse:
Your welcome. It is nice after all the puzzling I've done with Travis, to be able to help out with a project like yours.

fnc12:
@Farwaykorse can you please help me in added include path on osx - travis cannot compile unit tests https://github.com/fnc12/sqlite_orm/pull/334 
Thanks in advance

PR #335, opened 2019-07-05
Install Catch2 on AppVeyor using vcpkg

Farwaykorse:
With this CMake should be able to find Catch2.

Additionally I replaced the variables with AppVeyors `platform` setting.


Farwaykorse:
I do not believe there is any git submodule used in this project.

PR #339, opened 2019-07-05
Fix error in PR #338

Farwaykorse:
@fnc12 you where quick, I made a typo in the last commit to #338


Farwaykorse:
😛 the risk when editing directly on GitHub

PR #391, opened 2019-10-05
refactor get_pointer object creation

DoozyX:
Refactor get_pointer to create the object as a smart pointer at the beginning instead of normal object so no move constructor is needed when creating the unique pointer later.
When used with Qt framework and QObject base class the copy/move constructors are deleted by default.


fnc12:
@DoozyX Thank you. This is very useful. Also: shame to Qt =)

PR #392, opened 2019-10-06
added get all pointer

DoozyX:
Added get_all function that returns the objects as pointers. 
Needed when can't copy the objects so making them pointers can fix the issue.

### DIFF
--- a/dev/ast_iterator.h
+++ b/dev/ast_iterator.h
@@ -150,7 +150,17 @@ namespace sqlite_orm {
                 iterate_ast(get.conditions, l);
             }
         };
-        
+
+        template<class T, class ...Args>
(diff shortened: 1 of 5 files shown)


fnc12:
Hey. This is a very interesting case. Could you please describe it more detailed when object cannot be copied? Is it Qt or something else?

DoozyX:
Currently, all the features that I tested work fine. Nothing to do Qt directly, but by default on QObjects, the copy/move constructor and = operator are deleted so for me it was easier if I have all the objects stored as a smart pointer so I can use them between classes in a more safe way. This is not required but I thought it was nice to have.

fnc12:
Ok, this is nice. @DoozyX thank you

PR #393, opened 2019-10-10
fixed copy of handle when in-memory

biotty:


fnc12:
Pull requests are accepted only in the `dev` branch. Please change PR's destination and ensure that everything works fine. Thanks

biotty:
Our user code happens to copy your handler, which works and has handler semantics.  However, when it's opened as ":memory:" then a copied (as in passed by value to a function for example) the handler variable no longer works, because the currentTransaction is then assumed to always be kept and not reset after each transaction.  I can see that in the dev branch this issue seems to have gone away as a bonus of re-factoring impacting the in-memory behavior of the handle.

fnc12:
https://github.com/fnc12/sqlite_orm/pull/402 this is what I mean

Issue #396, opened 2019-10-13
Add a possibility to create classes from a schema


linux-fan-dave:
It would be good to generate classes and the required make_storage code from a given sql schema.

fnc12:
ok. But how would you specify "create public members" or "create private members with getters/setter"? And how would you specify type of getters and setters?

linux-fan-dave:
I was thinking about a template engine -> something like [mustache](https://github.com/no1msd/mstch). This way the converter creates a data structure with all the tables and all the properties. The template renders this data structure to c++ code. Some sample templates could be provided to create c++ classes or std data structures which contain the code. The user could extend this easily by changing the template.

linux-fan-dave:
I implemented a [sketch](https://github.com/linux-fan-dave/sqlite_orm/commit/22623c3d3e7887f9b0530bfa864a64fbde58b7f9) of the idea. Is there any chance to have code like this being added at all? Does this make any sense?

linux-fan-dave:
A different approach would be to have database schema exporter tool. It could reside in a different git repo and just provide a template for sqlite_orm. However I would prefer to ship it inside of sqlite_orm. The cmake could provide a function to generate source code via db  inspector and generate source code. This is not possible if this tool resides in a different repo

fnc12:
Code gen is a good idea and you made it clear for me with your PR but has nothing common with this lib except include directives in the output. So it would be better to store it in a different repo.

linux-fan-dave:
This is true. It does not share any code. I will go on and make it a repo myself then.

linux-fan-dave:
If I would add an option which would enable the creation

fnc12:
If we shall merge it than we need to make a lot of work before for:
1) fixing code style
2) adding opportunity for specifying type of getters/setter (there are 4 types of getters and 3 types of setters) and whether user wants to use public members instead of getters/setters
3) adding two more libs into this repo is a bad practice cause libs are being updated regularly and we need to sync them. It creates a package manager problem. I don't want to be a package manager and update this sources. Adding a package management software dependency will add more dependencies =) 

What I was thinking about is a code generator with a web interface where you type SQL, press `generate` and get a C++ code with this lib. E.g.: you type `SELECT *  FROM users WHERE id = 5` and get `auto rows = storage.get_all<User>(where(is_equal(&User::id, 5)));` 
or
`CREATE TABLE users (id integer not null primary key, name text not null)`
[...]

PR #397, opened 2019-10-13
#396 implement a schema first code generator

linux-fan-dave:
the implementation is not yet finished but I think it is obvious now where this is going. Is this an option for sqlite_orm?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_store
 examples/simple_neural_network.cpp
+CMakeLists.txt.user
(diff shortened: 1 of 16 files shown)


fnc12:
Hi. Thanks for your work. I don't see the reason why this code has to be placed in `sqlite_orm` repo. It doesn't use any source from it but adds more third party libraries. So it would be useful to store it in a different repo. Also you generate getters by value only. This is not convenient for everybody cause someone uses getters by const ref.

PR #398, opened 2019-10-14
added clang-format

DoozyX:
Added clang-format that I tried to match to your style but not all styles could be the same. I also applied the style formatting to all files that can be done with the following command: `find -name '*.cpp' -o -name '*.h' | xargs clang-format -i`
All the style options can be found on https://clang.llvm.org/docs/ClangFormatStyleOptions.html
You can replace the style with some already predefined styles as used for LLVM, Google, Chromium, Mozilla, WebKit with `clang-format -style=llvm -dump-config > .clang-format` as described in https://clang.llvm.org/docs/ClangFormat.html and then format the code again for all files.
Usually, most editors offer format on save or similar so everyone can use the same style without the need for manual checking.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,123 @@
+---
+Language:        Cpp
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortLambdasOnASingleLine: Empty
+AllowAllArgumentsOnNextLine: false
+AllowAllConstructorInitializersOnNextLine: false
+AllowShortBlocksOnASingleLine: false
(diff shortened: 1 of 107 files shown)


fnc12:
Thank you. Review will be long ~~and full of terrors~~. Also it would be great to add a configuration on CI for checking whether all files reminded in diff are formatted well. Can clang format perform this check?
P.S. I shall merge it once I finish my review

DoozyX:
Do you want CI to fail if it is not formatted correctly only? I think it's not possible by clang-format itself but there are some cmake or python scripts that can use clang-format to check if the formatting is correct.

fnc12:
I think I know how to check whether code is formatted well in a PR. Make a script with code format and `get status` further check

DoozyX:
I think one option is with CMake, example script is here 
https://github.com/ttroy50/cmake-examples/tree/master/04-static-analysis/clang-format
another option is 
https://github.com/Sarcasm/run-clang-format
If you want I can integrate one of those

fnc12:
@DoozyX  Yeah it would be great if yon integrate this. Thank you!

PR #403, opened 2019-10-19
Add clang-format lint action

DoozyX:
This pull request adds clang-format lint action that checks if the code matches the .clang-format file.
It ignores the sqlite_orm.h header because it is generated from the python script and the formatting is not exactly the same.
I can't see the actions tab on your repo, if it doesn't work here you may need to sign up on the beta here https://github.com/features/actions . The GitHub actions will be generally available on November 13.


fnc12:
Thank you!

PR #405, opened 2019-10-21
clang-format lint action not active

DoozyX:
Tried updating the action, currently, it is not triggering the workflow.

### DIFF
--- a/.github/workflows/clang-format-lint.yaml
+++ b/.github/workflows/clang-format-lint.yaml
@@ -4,7 +4,6 @@ on: [push]
 
 jobs:
   build:
-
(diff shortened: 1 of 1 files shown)


DoozyX:
Can you merge it into dev? my guess is that it doesn't work because we added the workflow before it got enabled on this repo, and by making a change to it it will register it. If it doesn't work I've sent a ticket to GitHub about the issue.

fnc12:
done. Thank you

DoozyX:
Still not working I think. On the actions tab, it says "Only repository collaborators may create workflows." maybe you need to remove the old action and make it again yourself?

PR #415, opened 2019-11-06
Extra info when failing to open a database

air2:
Introduce a generic error message function
Introduce a generic throw function. Both are introduced to remove duplicate code and should be used when throwing exceptions (IMHO)


fnc12:
@air2 please merge `dev` branch and pass CI checks again. We've improved code format check and we need to check its work. Also it looks like that you skipped code formatting

PR #417, opened 2019-11-06
Extra info when failing to open database.

air2:
Introduce a generic error message function
Introduce a generic throw function. Both are introduced to remove duplicate code and should be used when throwing exceptions (IMHO)


fnc12:
@air2 thank you. BTW it is not required to recreate a PR in case you make changed. The existing PR will trigger CI and update diff automatically you push new commits

PR #566, opened 2020-07-26
rm compile folder

air-h-128k-il:
a little modification for Mac OSX

Recent default OSX's shell is zsh


fnc12:
@air-h-128k-il please make a brach from `dev` and PR to `dev`

PR #572, opened 2020-08-01
571/535 Feature/dependency managment

Leon0402:
Hi there, 

as requested the changes discussed in the two issues to make using the library more comfortable.

What I change:
- Added sqlite3 as an interface dependency, so all consuming targets that use sqlite_orm have sqlite3 automatically as a dependency -> no manual linking or any work, but sqlite3 obviously still has to be installed and in the right path
- Added catch as a dependency, instead of using find_package it is using fetchContent and therefore will be pulled in automatically
- Make it more FetchContent compatibility by disabling tests and examples if the project is pulled in another project
- Update some of the install logic, so find_package works
- Move cmake folder out of the build folder (you should really avoid that! The whole point of out of source build is that you can just delete the build folder and start from scratch)
- Provide an example on how to use this lib in your own project with FetchContent (which I personally consider the easiest way)
- Provide an example on how to use this lib in your own project with find_package

I tested both examples on my Linux machine. 

Note: I'm not super good with what feature is added when in cmake, so I basically wrote this with my cmake 3.17 knowledge. It could be that I rely on stuff which is done automatically in higher cmake versions. But as it compiles and the examples work, it seems rather unlikely. I do recommend though to bump the cmake version to something higher (say 3.15 which is the latest stable version I think?)


fnc12:
@Leon0402 please fix cmake error on windows
```
mkdir compile
57cd compile
58cmake %SQLITE_ORM_CXX_STANDARD% -DSqliteOrm_BuildTests=ON .. -G %generator% -DCMAKE_TOOLCHAIN_FILE=C:/Tools/vcpkg/scripts/buildsystems/vcpkg.cmake
59CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
60  CMake 3.17 or higher is required.  You are running version 3.16.2
```

Leon0402:
My bad, I bumped the cmake version as part of a test and forgot it :D

Leon0402:
And before that one here is merged, we probably want to change the url in the fetchContent example to point to this repo. 
It's just a little bit difficult as it should presumbably just pick the code from the master or a tag ... but in order to do so, the code here has to be on master :/
Kinda a chicken egg problem

examples/fetch_content/src/main.cpp
```diff
@@ -0,0 +1,46 @@
+#include <sqlite_orm/sqlite_orm.h>
+
+#include <string>
+#include <iostream>
+#include <cassert>
+
+using namespace sqlite_orm;
```
fnc12, line 7: what is this test created for? What does it mean `fetch_content`? If it is one of the examples why is has its own directory instead of a single file with a custom name just like all other examples?

examples/find_package/src/main.cpp
```diff
@@ -0,0 +1,46 @@
+#include <sqlite_orm/sqlite_orm.h>
+
+#include <string>
+#include <iostream>
+#include <cassert>
```
fnc12, line 5: what does this example shows? How cmake works?

examples/fetch_content/src/main.cpp
fnc12: I found the answer in the first message.

examples/find_package/src/main.cpp
fnc12: I found the answer in the first message.

fnc12:
@Leon0402 thank you

fnc12:
linux run failed on CI https://travis-ci.org/github/fnc12/sqlite_orm/builds/714086861 . Please check it out.

fnc12:
damn. Looks like build failed on your branch and travis status did not show up here. Could you please fix it? Here you can find details https://travis-ci.org/github/fnc12/sqlite_orm/builds/714020440 https://travis-ci.org/github/fnc12/sqlite_orm/jobs/714020441

Leon0402:
Hi @fnc12 checked it out, the problems seems that fetchContent isn't available in 3.2. Sorry, as I said, I don't know exactly when which feature was added. It seems like the windows build was sucessfull because it has a much newer version installed (cmake 3.16 I think).
We need at least cmake 3.11. I would go with cmake 3.13 as this is what debian stable supports https://packages.debian.org/de/buster/cmake (which seems like a good indicator to me). It was released about two years ago, what do you think?

Also, should I open a new Pull Request or how does this work? I would additionally also update the docs.

Btw. the development speed is awesome, you react to issues in no time and make it easy to contribute !

fnc12:
You can either make a new PR or reopen existing. Does not matter. Just fix it please cause right now dev branch has failing build. Thank you

PR #591, opened 2020-09-10
Compile error due to insufficient braces.

undisputed-seraphim:
When compiling in Ubuntu 18 with GCC 7.5 and 9.3 (tested both) and with the C++17 flag on, GCC complains that there are missing braces, and fails compilation.

The issue stems from the `-Wmissing-braces` warning, which gets elevated into an error when compiling under stricter modes.

An investigation leads to [discussions such as this](https://stackoverflow.com/questions/11734861/when-can-outer-braces-be-omitted-in-an-initializer-list).

After adding the braces I find that GCC happily compiles the code.


undisputed-seraphim:
I will put it together with the other PR.

PR #598, opened 2020-09-25
sqlite3_busy_handler

undisputed-seraphim:
I propose a possible implementation of a busy_handler function.

As sqlite3 has a pure C interface, it is not possible to simply pass `std::function` or lambdas into `sqlite3_busy_handler()`.
My solution is to shove a `std::function` and the user pointer into a struct, pass it into the C interface as a void pointer, and unpack it from the inside.
This way, we get to also expose a strongly typed interface to the user, through the template.
It can accept lambda captures too.

Warning: I don't really have a way to unit test this. The tests written are to ensure compile-ability only.

### DIFF
--- a/dev/storage_base.h
+++ b/dev/storage_base.h
@@ -124,6 +124,25 @@ namespace sqlite_orm {
                 return sqlite3_busy_timeout(con.get(), ms);
             }
 
+            /*
+             * Ownership is NOT transferred, i.e. data
+             * must not go out of scope for the lifetime of the database.
+             */
+            template<typename T = void>
+            int busy_handler(std::function<int(T *, int)> f, T &data) {
+                handler = {[f{move(f)}](void *ptr, int times) -> int {
+                               return f(reinterpret_cast<T *>(ptr), times);
+                           },
+                           (void *)&data};
+                return sqlite3_busy_handler(
+                    this->get_connection().get(),
+                    [](void *ptr, int times) -> int {
+                        auto h = *reinterpret_cast<handler_t *>(ptr);
+                        return h.fn(reinterpret_cast<T *>(h.ptr), times);
+                    },
+                    &handler);
+            }
+
             /**
              *  Returns libsqlite3 lib version, not sqlite_orm
              */
@@ -339,6 +358,11 @@ namespace sqlite_orm {
             std::unique_ptr<connection_holder> connection;
             std::map<std::string, collating_function> collatingFunctions;
             const int cachedForeignKeysCount;
+
+            struct handler_t final {
+                std::function<int(void *, int)> fn;
+                void *ptr;
+            } handler;
 
             connection_ref get_connection() {
                 connection_ref res{*this->connection};
--- a/dev/storage_base.h
+++ b/dev/storage_base.h
@@ -125,7 +125,7 @@ namespace sqlite_orm {
             }
 
             /**
-             *  Returns libsqltie3 lib version, not sqlite_orm
+             *  Returns libsqlite3 lib version, not sqlite_orm
              */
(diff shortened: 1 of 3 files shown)


fnc12:
hi. Please make code format. One can use this command `clang-format -i -style=file include/sqlite_orm/*.h tests/*.cpp tests/*/*.cpp dev/*.h tests/*/*.h examples/*.cpp`

undisputed-seraphim:
Woops, done

dev/storage_base.h
```diff
@@ -124,8 +124,27 @@ namespace sqlite_orm {
...
+                    std::function<int(T *, int)> fn;
+                    T *ptr;
+                } handler = {std::move(f), ptr};
+                return sqlite3_busy_handler(
+                    this->get_connection().get(),
+                    [](void *ptr, int times) -> int {
+                        auto h = *reinterpret_cast<handler_t *>(ptr);
+                        return h.fn(h.ptr, times);
+                    },
+                    &handler);
```
fnc12, line 143: `handler` will be destroyed once this function exits. Then its' address will be captured as user data. Then dereferencing it will lead to UB. It is got to be fixed.

fnc12 (changes requested):

dev/storage_base.h
undisputed-seraphim: We can put the struct inside the protected area.

dev/storage_base.h
```diff
@@ -124,8 +124,27 @@ namespace sqlite_orm {
                 return sqlite3_busy_timeout(con.get(), ms);
             }
 
+            /*
+             * Ownership is NOT transferred, i.e. data
+             * must not go out of scope for the lifetime of the database.
+             */
+            template<typename T = void>
+            int busy_handler(std::function<int(T *, int)> f, T &data) {
```
fnc12, line 132: I don't understand why do you need this function to be template. Actually `std::function` allows us binding any data we want so it is enough. Dedicated argument for custom data is not required - it can be stored inside lambdas capture. `std::function` must be a member inside `storage_base`. Next in `on_open_internal` you must call `sqlite3_busy_handler` if this member is not null. Custom data argument must be a pointer to a storage and callback that is must be passed to `sqlite3_busy_handler` function must be a static private/protected function in `storage_base` class. Why this idea is better? Cause it has less templates. Templates must exist only if we cannot avoid them. The second is custom data is stored inside `std::function` and this is ok - it solves the issue from the first comment. I'd like you to refactored your solution as I described here. Of course if you have a different opinion or a better solution please feel free to comment it here or email me (the first option is more preferred cause conversation will be available for everyone). If you want I can implement it on my own today or tomorrow.

fnc12 (changes requested):

fnc12:
@undisputed-seraphim please check out this branch https://github.com/fnc12/sqlite_orm/pull/600. I need feedback from you whether it has functionality you want. Thanks

fnc12:
@undisputed-seraphim are you there?

undisputed-seraphim:
Hi sorry for the silence. Life got busy suddenly.  I'll take a look.

dev/storage_base.h
undisputed-seraphim: I have chosen to write it this way for one simple reason: Because it directly correlates to the function signature documented as sqlite, `int(void*, int)`. 
The user of this interface may check `sqlite3_busy_handler`, see that it asks for a function `int(void*, int)`, and see that this interface does not expose that functionality. It is not obvious that the idea is to capture data in the lambdas.
I have seen your merge request. It is good, but please add documentation about the intention.

PR #600, opened 2020-09-28
added busy handler function

fnc12:


undisputed-seraphim:
I see that it has been merged, and I was not available to take a look. Sorry about that.

I'd like some documentation about the intention so that it is clear how this is meant to be used.

PR #605, opened 2020-10-06
Update documentation on CMake

undisputed-seraphim:
Add an example right in the README.


fnc12:
@undisputed-seraphim thanks

PR #609, opened 2020-10-09
Fix/travis

Leon0402:
Here we go again, this time squash it.


fnc12:
merged with squash. Thank you very much. Your nick will be added in special thanks list in the next release information.

PR #612, opened 2020-10-13
Extract sqlite3_exec

undisputed-seraphim:
Many of the function blocks that do `sqlite3_prepare_v2`, `statement_finalizer`, and `perform_step`/`sqlite3_step` are essentially `sqlite3_exec` calls, especially if no results are returned.

### DIFF
--- a/dev/util.h
+++ b/dev/util.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <sqlite3.h>
+#include <string>  //  std::string
 #include <system_error>  //  std::system_error, std::error_code
 
 namespace sqlite_orm {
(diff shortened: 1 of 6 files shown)


dev/util.h
```diff
@@ -14,5 +15,13 @@ namespace sqlite_orm {
                                         sqlite3_errmsg(db));
             }
         }
+
+        static void perform_void_exec(sqlite3 *db, const std::string &query) {
+            int rc = sqlite3_exec(db, query.c_str(), NULL, NULL, NULL);
```
fnc12, line 20: please use `nullptr` instead of `NULL`. `NULL` is legacy C macro and `nullptr` is strong typed

fnc12 (changes requested):

dev/util.h
undisputed-seraphim: Done.

fnc12 (approved):

fnc12:
@undisputed-seraphim thanks!

PR #617, opened 2020-10-20
cmake: Prettify unit test file list

undisputed-seraphim:
Rebasing against dev whenever there is a new change to the unit tests is a pain

Let's do it this way so it is easier to see what has changed


fnc12:
@undisputed-seraphim thanks

PR #627, opened 2020-11-19
Collate prepare_statement functions into another template function.

undisputed-seraphim:
I propose a change to the code to simplify the `prepare` functions by factoring out a `prepare_impl`, since all of them use the same block of code.

I understand that this change is a bit more radical than my previous pull requests, so maybe you will have objections. But it looks nicer this way :) 

Please review and give feedback.

### DIFF
--- a/dev/storage.h
+++ b/dev/storage.h
@@ -707,6 +707,24 @@ namespace sqlite_orm {
                 return res;
             }
 
+            template<typename S>
+            prepared_statement_t<S> prepare_impl(S statement) {
+                auto con = this->get_connection();
+                sqlite3_stmt *stmt;
+                auto db = con.get();
+                using context_t = serializator_context<impl_type>;
+                context_t context{this->impl};
+                context.skip_table_name = false;
+                context.replace_bindable_with_question = true;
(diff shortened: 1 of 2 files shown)


dev/storage.h
```diff
@@ -774,301 +792,93 @@ namespace sqlite_orm {
...
-                context.skip_table_name = false;
-                context.replace_bindable_with_question = true;
-                auto query = serialize(sel, context);
-                if(sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr) == SQLITE_OK) {
-                    return {std::move(sel), stmt, con};
-                } else {
-                    throw std::system_error(std::error_code(sqlite3_errcode(db), get_sqlite_error_category()),
-                                            sqlite3_errmsg(db));
-                }
+                return prepare_impl<select_t<T, Args...>>(std::move(sel));
```
fnc12, line 795: tip: it is not necessary to specify template argument explicitly in this case. One can write `return prepare_impl(std::move(sel));` and that will work

fnc12:
this is a great PR. Thank you. What do you think about the comment I wrote there?

PR #636, opened 2020-12-19
Maybe it's easier to use if we fetch sqlite3 as sub-module of this repo with cmake ?

dhmemi:
sqlite3 lib is not convenient enough as it can not be build and include with cmake. So we have to build an install it in our develop environment before use it. and 'find_package()' of cmake is not always works well. Maybe it's easier to use If sqlite3 can be imported automatically when this repo is imported.

I made some changes in CMakeLists to achieve this goal, and now we can import this repo in other project without pre-install sqlite3, just need:
```cmake

FetchContent_Declare(
  sqlite_orm
  GIT_REPOSITORY https://github.com/dhmemi/sqlite_orm
  GIT_TAG        master
)

FetchContent_MakeAvailable(sqlite_orm)
```

But If we config this repo independently, some error in packaging like below will be encountered:
```
CMake Error: install(EXPORT "SqliteOrmTargets" ...) includes target "sqlite_orm" which requires target "sqlite3" that is not in any export set.
```
I do not know how to deal with this as I'm not familiar with cmake-install. Do you or anyone have any idea about this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ auto storage = make_storage("db.sqlite",
                             make_table("user_types",
                                        make_column("id", &UserType::id, autoincrement(), primary_key()),
                                        make_column("name", &UserType::name, default_value("name_placeholder"))));
+storage.sync_schema();
 ```
 
 Too easy isn't it? You do not have to specify mapped type explicitly - it is deduced from your member pointers you pass during making a column (for example: `&User::id`). To create a column you have to pass two arguments at least: its name in the table and your mapped class member pointer. You can also add extra arguments to tell your storage about column's constraints like `primary_key`, `autoincrement`, `default_value` or `unique`(order isn't important; `not_null` is deduced from type automatically).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
 include(CTest)
 
+option(SqliteOrm_SysSqlite "Use system version of sqlite library" OFF)
+
 ### Dependencies
 add_subdirectory(dependencies)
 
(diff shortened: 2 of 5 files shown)


fnc12:
@dhmemi hi. Thank you for taking part in development of the lib. Actually I can not answer your question about cmake install cause I also don't know how cmake works. I mark this PR as `help wanted`.

fnc12:
something wrong with appveyor and travis. They are required to validate that unit tests and example work fine in your branch before merging this PR. Let me fix it first.

README.md
```diff
@@ -88,6 +88,7 @@ auto storage = make_storage("db.sqlite",
                             make_table("user_types",
                                        make_column("id", &UserType::id, autoincrement(), primary_key()),
                                        make_column("name", &UserType::name, default_value("name_placeholder"))));
+storage.sync_schema();
```
fnc12, line 91: why? You make the first block more complicated. Info about `sync_schema` is noted later and it is ok

fnc12 (changes requested):

README.md
dhmemi: When I first used this library，I followed the README from top to bottom. before I read the `Migrations functionality`, my code was like:
```cpp
struct User{
    int id;
    std::string firstName;
    std::string lastName;
    int birthDate;
    std::unique_ptr<std::string> imageUrl;
    int typeId;
};

struct UserType {
    int id;
    std::string name;
};

int main(){
  using namespace sqlite_orm;
  auto storage = make_storage("db.sqlite",
                              make_table("users",
                                         make_column("id", &User::id, autoincrement(), primary_key()),
                                         make_column("first_name", &User::firstName),
                                         make_column("last_name", &User::lastName),
                                         make_column("birth_date", &User::birthDate),
                                         make_column("image_url", &User::imageUrl),
                                         make_column("type_id", &User::typeId)),
                              make_table("user_types",
                                         make_column("id", &UserType::id, autoincrement(), primary_key()),
                                         make_column("name", &UserType::name, default_value("name_placeholder"))));
  
  User user{-1, "Jonh", "Doe", 664416000, std::make_unique<std::string>("url_to_heaven"), 3 };
      
  auto insertedId = storage.insert(user);
  cout << "insertedId = " << insertedId << endl;      //  insertedId = 8
  user.id = insertedId;
  
  return 0;
}
```
But it didn't work because of no `sync_schema()` and I get an exception which took me a lot of time to debug. Maybe other developers would make the same mistake if we don't add this line at the first code section.

fnc12: Sounds logic but it is not acceptable to everyone. Some people don't use `sync_schema` but just use already existing database with correct schema

fnc12:
is this PR actual?

fnc12:
@dhmemi hi. You need to fix appveyor and travis builds/configurations before merging this PR

dhmemi:
Close this PR temporarily as I have no energy and idea to deal with these issues.

PR #641, opened 2020-12-29
No drop opt

denzor200:
Added build option SQLITE_ORM_PROHIBIT_DROPPING_AND_RECREATING which disallows dropping tables during database structure update. This can be important for applications that are concerned about the safety of their data.

### DIFF
--- a/dev/error_code.h
+++ b/dev/error_code.h
@@ -23,6 +23,7 @@ namespace sqlite_orm {
         failed_to_init_a_backup,
         unknown_member_value,
         incorrect_order,
+        dropping_and_recreating_was_prohibited
     };
 
 }
@@ -63,6 +64,8 @@ namespace sqlite_orm {
                     return "Unknown member value";
                 case orm_error_code::incorrect_order:
                     return "Incorrect order";
+                case orm_error_code::dropping_and_recreating_was_prohibited:
+                    return "Dropping and re-creating tables in sync_schema method was prohibited";
                 default:
                     return "unknown error";
             }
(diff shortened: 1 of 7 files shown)


fnc12:
why just we can't use `sync_schema_simulate`? There is no need to add compile options if this can be solved with during runtime

denzor200:
> why just we can't use `sync_schema_simulate`? There is no need to add compile options if this can be solved with during runtime

1. Nobody canceled anomalies in the multithreaded environment. If the memory has been corrupted by a third-party thread, the sync_schema_simulate method may lie to us that tables will not be deleted. I did not go into the details of the implementation of this method, perhaps it is not.
2. The sync_schema_simulate method has a rather complex implementation, it is rather difficult to convince someone unfamiliar with C++ (CTO, PM) that the system will not drop tables during the update.

fnc12:
> Nobody canceled anomalies in the multithreaded environment. If the memory has been corrupted by a third-party thread, the sync_schema_simulate method may lie to us that tables will not be deleted. I did not go into the details of the implementation of this method, perhaps it is not.

Any anomalies in the multithreaded environment can be solved on client (your) side. E.g. you can use mutex to cover all mutating calls to the storage. BTW you'd have the same issue without `sqlite_orm` but with raw sqlite3. If memory was corrupted by a third-party code then you have to fix this code usage but don't touch `sqlite_orm`. Your project may have 20+ libs and can shoot memory in any of them but it is not the reason to make 20+ PRs in these libs.

> The sync_schema_simulate method has a rather complex implementation, it is rather difficult to convince someone unfamiliar with C++ (CTO, PM) that the system will not drop tables during the update.

`sync_schema_simulate` was created especially for people who does not trust `sync_schema` call. `sync_schema_simulate` has the same logic and complexity that `sync_schema` except altering your database.

Also you PR make `sync_schema` and `sync_schema_simulate` return different values if `SQLITE_ORM_PROHIBIT_DROPPING_AND_RECREATING` is defined and it is incorrect.

Just use `sync_schema_simulate`. It works well enough.

fnc12:
is it actual?

PR #644, opened 2021-01-16
Non-standart primary key checking

denzor200:
This pull request adds a useful compile-time error. You can reproduce it with a simple example:
```c++

using namespace sqlite_orm;

struct User {
    std::string username;
    std::string password;
    bool isActive;
};

int main()
{
    auto storage = make_storage("storage.sqlite", ///
        make_table("users", ///
            make_column("username", &User::username, primary_key()),
            make_column("password", &User::password),
            make_column("isActive", &User::isActive)));
    storage.sync_schema();

    storage.insert(User { "testName", "testPassword2", false });

    return 0;
}
```

In the current sqlite_orm, this example throws a runtime error "NOT NULL constraint failed", which is somewhat misleading.


denzor200:
not tested under Windows .. Tomorrow I will fix it all

denzor200:
@fnc12, fixed

dev/storage.h
```diff
@@ -1027,11 +1028,19 @@ namespace sqlite_orm {
                 auto compositeKeyColumnNames = tImpl.table.composite_key_columns_names();
                 sqlite3_reset(stmt);
                 tImpl.table.for_each_column([&o, &index, &stmt, &tImpl, &compositeKeyColumnNames, db](auto& c) {
+                    using table_type = typename std::decay<decltype(tImpl.table)>::type;
```
fnc12, line 1031: are you sure this check is required here? The point is this function cannot be called without calling `statement_serializator` before. `statement_serializator` has this check already. Or this code got to be moved somewhere at one place to skip code copypasting

dev/table.h
```diff
@@ -26,28 +26,31 @@ namespace sqlite_orm {
              */
             std::string name;
 
-            bool _without_rowid = false;
+            const bool _without_rowid;
```
fnc12, line 29: are you sure that we need this field? I think that we have already enough data to evaluate whether table has PK if we know storage type. And we don't need more data to store

dev/table.h
```diff
@@ -26,28 +26,31 @@ namespace sqlite_orm {
...
+        struct table_without_rowid_t;
+
         /**
-         *  Table interface class. Implementation is hidden in `table_impl` class.
+         *  Template for table interface class. Implementation is hidden in `table_impl` class.
          */
-        template<class T, class... Cs>
-        struct table_t : table_base {
+        template<class T, bool WR, class... Cs>
+        struct table_template : private table_base {
```
fnc12, line 39: why `private` inheritance? Is there any reason for this?

fnc12 (changes requested):

dev/statement_serializator.h
```diff
@@ -962,6 +962,12 @@ namespace sqlite_orm {
                 auto compositeKeyColumnNames = tImpl.table.composite_key_columns_names();
 
                 tImpl.table.for_each_column([&tImpl, &columnNames, &compositeKeyColumnNames](auto& c) {
+                    static_assert(
+                        (is_table_without_rowid<typename std::decay<decltype(tImpl.table)>::type>::value ||
+                         !c.template has<constraints::primary_key_t<>>() ||
+                         std::is_same<typename std::decay<decltype(c)>::type::field_type, int>::value),
```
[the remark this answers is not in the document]
denzor200: Okay, I'll fix it.

dev/table.h
```diff
@@ -19,35 +19,38 @@ namespace sqlite_orm {
 
     namespace internal {
 
+        template<bool WR>
```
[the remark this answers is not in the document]
denzor200: It was resolved in 255eac2 ('table_base' is simple structure now, not a template)

dev/table.h
denzor200: Ok, i will remove this field

dev/table.h
denzor200: Private inheritance prevents conversions from 'table_t ' to 'table_without_rowid_t' (and vice versa).
Perhaps this is excessive paranoia ..
Maybe we dont need 'table_base' structure? Let's just move the 'name' parameter to 'table_template'?

dev/storage.h
denzor200: Okay, I'll find a suitable place and move this check there, without copypasting

dev/column.h
```diff
@@ -109,6 +109,48 @@ namespace sqlite_orm {
...
+             *  Column with noninsertable primary key traits. Common case.
+             */
+            template<class T, class SFINAE = void>
+            struct is_column_with_noninsertable_primary_key : public std::false_type {};
+
+            /**
+             *  Column with noninsertable primary key traits. Specialized case case.
+             */
+            template<class O, class T, class... Op>
+            struct is_column_with_noninsertable_primary_key<
```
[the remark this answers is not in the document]
fnc12: or you can make only one common specification and use `value = !is_column_with_insertable_primary_key`

dev/storage.h
```diff
@@ -600,6 +619,7 @@ namespace sqlite_orm {
             template<class O>
             int insert(const O& o) {
                 this->assert_mapped_type<O>();
+                this->assert_insertable_type<O>();
```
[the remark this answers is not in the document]
fnc12: it is still actual.

dev/storage.h
```diff
@@ -608,6 +628,7 @@ namespace sqlite_orm {
             void insert_range(It from, It to) {
                 using O = typename std::iterator_traits<It>::value_type;
                 this->assert_mapped_type<O>();
+                this->assert_insertable_type<O>();
```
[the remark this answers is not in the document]
fnc12: still actual

dev/table.h
```diff
@@ -264,6 +255,46 @@ namespace sqlite_orm {
...
+        template<class T, class... Cs>
+        struct table_t : table_template<T, false, Cs...> {
+            using table_template<T, false, Cs...>::table_template;
+        };
+
+        /**
+         *  Table interface class with 'without_rowid' tag. Implementation is hidden in `table_impl` class.
+         */
+        template<class T, class... Cs>
+        struct table_without_rowid_t : table_template<T, true, Cs...> {
```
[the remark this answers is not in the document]
fnc12: still actual

dev/table.h
```diff
@@ -264,6 +255,46 @@ namespace sqlite_orm {
                 return res;
             }
         };
+
+        /**
+         *  Table interface class. Implementation is hidden in `table_impl` class.
```
fnc12, line 260: `table_impl` does not exist

tests/unique_cases/nonstandart_primary_key.cpp
```diff
@@ -0,0 +1,52 @@
...
+}
+
+TEST_CASE("Nonstandart primary key - fail test") {
+    struct User {
+        std::string username;
+        std::string password;
+        bool isActive;
+    };
+
+    auto storage = make_storage({},  ///
```
fnc12, line 20: note: strange comments

dev/column.h
```diff
@@ -109,6 +109,48 @@ namespace sqlite_orm {
             }
         };
 
+        // we are compelled to wrap all sfinae-implemented traits to prevent "error: type/value mismatch at argument 2 in template parameter list"
+        namespace sfinae {
+            /**
+             *  Column with insertable primary key traits. Common case.
+             */
+            template<class T, class SFINAE = void>
+            struct is_column_with_insertable_primary_key : public std::false_type {};
```
fnc12, line 118: please add static tests for it (one at least)

dev/constraints.h
```diff
@@ -450,6 +451,23 @@ namespace sqlite_orm {
          */
         template<class... Cs>
         struct is_primary_key<constraints::primary_key_t<Cs...>> : public std::true_type {};
+
+        /**
+         * PRIMARY KEY INSERTABLE traits.
+         */
+        template<typename T>
+        struct is_primary_key_insertable {
```
fnc12, line 459: please add static tests for it

dev/tuple_helper.h
```diff
@@ -25,6 +27,24 @@ namespace sqlite_orm {
...
         using tuple_contains_type = typename has_type<T, Tuple>::type;
 
+        /**
+         *  HAS_SOME_TYPE type trait
+         */
+        template<template<class> class TT, typename Tuple>
+        struct has_some_type;
+
+        template<template<class> class TT>
+        struct has_some_type<TT, std::tuple<>> : std::false_type {};
```
fnc12, line 37: let's make some static tests for it please

fnc12 (changes requested):
please fix and/or reply to all comments before merging

dev/table.h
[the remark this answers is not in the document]
denzor200: I think that we should leave `table_without_rowid_t`, beacause _"a template bool arg in `table_t` class"_ may deliver some small problems for users. 
For example, lets see section "Slow & size effective"
https://github.com/fnc12/sqlite_orm/wiki/FAQ

tests/unique_cases/nonstandart_primary_key.cpp
denzor200: this is for control over clang-format

dev/column.h
```diff
@@ -121,6 +163,18 @@ namespace sqlite_orm {
         template<class O, class T, class... Op>
         struct is_column<column_t<O, T, Op...>> : public std::true_type {};
 
+        /**
+         *  Column with insertable primary key traits.
+         */
+        template<class T>
+        struct is_column_with_insertable_primary_key : public sfinae::is_column_with_insertable_primary_key<T> {};
```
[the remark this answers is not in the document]
denzor200: No. It can be removed in C++17, but we use C++14

dev/column.h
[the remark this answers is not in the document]
denzor200: No, it will break the logic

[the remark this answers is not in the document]
denzor200:
```
static_assert(
    count_tuple<columns_type, is_column_with_insertable_primary_key>::value <= 1,
    "Attempting to execute 'insert' request into an noninsertable table was detected. "
    "Insertable table cannot contain > 1 primary keys. Please use 'replace' instead of "
    "'insert', or you can use 'insert' with explicit column listing.");
static_assert(
    count_tuple<columns_type, is_column_with_noninsertable_primary_key>::value == 0,
    "Attempting to execute 'insert' request into an noninsertable table was detected. "
    "Insertable table cannot contain non-standard primary keys. Please use 'replace' instead "
    "of 'insert', or you can use 'insert' with explicit column listing.");
```
Second static_assert will be triggered for all normal columns, even if column is not a primary key

fnc12:
@denzor200 thanks

PR #652, opened 2021-01-26
made field_printer customizable via sfinae (fixes #649)

### LINKED ISSUE #649: Make field_printer customization via SFINAE avaliable
Now I have to write explicit specialization for every type, but I want to use one partial specialization which will be chosen via SFINAE.

For example, now I have to write something like this:
```c++
enum class E1 { ... };
enum class E2 { ... };
enum class E2 { ... };

template <typename T>
struct my_enums_field_printer
{
    using integral_type = std::underlying_type_t<T>;
    std::string operator()(const T& value) const
    {
        return field_printer<integral_type>()(static_cast<integral_type>(value));
    }
};

template <>
struct field_printer<E1> : my_enums_field_printer<E1> {};

template <>
struct field_printer<E2> : my_enums_field_printer<E2> {};

template <>
struct field_printer<E3> : my_enums_field_printer<E3> {};
```

I want to just write:
```c++
template <typename T>
struct field_printer<T, std::enable_if_t<std::is_enum_v<T>>>
{
    using integral_type = std::underlying_type_t<T>;
    std::string operator()(const T& value) const
    {
        return field_printer<integral_type>()(static_cast<integral_type>(value));
    }
};
```

It can be done if you provide to the field_printer second template parameter for SFINAE  like you did for other traits (type_printer, row_extractor, statement_binder).

I can make pull request for next week, if you don't mind these changes.

Overlordff:


fnc12:
@Overlordff thank you. Please make sure that enum binding examples (two items) compile well

Overlordff:
All examples compiles

fnc12:
@Overlordff thanks again!

Issue #654, opened 2021-01-28
Without rowid


denzor200:
The `_without_rowid` parameter only affects the `insert` method. This parameter does not apply to `insert_range`. Is this expected?

denzor200:
statement serializator for `insert`:
https://github.com/fnc12/sqlite_orm/blob/dev/include/sqlite_orm/sqlite_orm.h#L10710
statemente serializator for `insert_range`:
https://github.com/fnc12/sqlite_orm/blob/dev/include/sqlite_orm/sqlite_orm.h#L10832

fnc12:
It looks like a bug

denzor200:
Ok, i will fix it in separate PR

denzor200:
Fixed.
https://github.com/fnc12/sqlite_orm/pull/655

fnc12:
@denzor200  thanks. I've merged it a minute ago.

PR #655, opened 2021-01-30
#654 without rowid

denzor200:


denzor200:
I can't understand what's meaning this overload:
```
int insert(const O& o, columns_t<Cols...> cols);
```
Does it have any relevance to our topic?

fnc12:
it is `insert` with explicit columns. It is called like this:
```c++
storage.insert(myObject, columns(&Object::id, &Object::name));
```
It is used to specify columns explicitly in case we want to omit some of them in `INSERT INTO` query. It is not related to this PR. This PR is good. Thanks

PR #657, opened 2021-02-01
added nullopt support

fnc12:


dev/field_printer.h
```diff
@@ -85,7 +85,14 @@ namespace sqlite_orm {
             return "null";
         }
     };
-
+#ifdef SQLITE_ORM_OPTIONAL_SUPPORTED
+    template<>
+    struct field_printer<std::nullopt_t, void> {
+        std::string operator()(const std::nullopt_t&) const {
+            return "null";
```
Overlordff, line 92: You can now use this specialization for null values in `field_printer<std::optional<T>>` specialization instead of `std::nullptr_t`.

Overlordff (changes requested):
In general, everything is fine

PR #658, opened 2021-02-02
created distinct tests

fnc12:


tests/select_constraints_tests.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+            get<0>(statement) = 10;
+            expected.clear();
+            expected.push_back(10);
+            rows = storage.execute(statement);
+            REQUIRE(rows == expected);
+        }
+        SECTION("just names") {
+            auto rows = storage.select(as_optional(&Employee::name));
+            decltype(rows) expected;
```
Overlordff, line 138: I think we should check that `decltype(rows)::value_type` type is `optional` (`static_assert` or something like that) because now test will be passed if `as_optional` doesn't change result type.

fnc12: I thought about it cause you have already wrote it in your example code. But it turned out that it is over-engineering cause we already comparing vectors inside `REQUIRE` macro. It types are different the code will not compile.

PR #666, opened 2021-02-10
Bugfix/663 insert range

denzor200:


denzor200:
`execute` methods should also be collapsed into one

fnc12:
@denzor200 looks nice. Please fix windows build also

PR #680, opened 2021-03-15
Use constant references when iterating over ranges in replace_range

andrei-datcu:
This is required if one wants to use iterators that do not return references when are dereferenced.
Usually, this will happen whenever using something from `std::ranges`. Eg:

```c++
struct Message {
    int id;
    int code;
}
std::vector<int> messageCodes = {....};
int id = 0;
auto messages = messageCodes | std::views::transform([&id](int code) mutable { return Message {.id = id++, .code = code};});
storage.replace_range(messages.begin(), messages.end());

```


fnc12:
nice. Let's wait for builds

fnc12:
@andrei-datcu PRs must me merged into `dev` branch not `master`. I accidentally merged in and I am going to unmerge it right now. Please reopen it with a branch from `dev`

PR #685, opened 2021-03-20
Make iterator_t std::input_iterator compatible

andrei-datcu:
This PR makes the `iterator_t` class compatible with the new `std::input_iterator` concept. This way, the iterator can be used with all the standard ranges.

Things I've changed:
  * `iterator_t` is now default constructible
  * `const iterator_t` is now dereferenceable (added the constness). This means that the new values are generated during iterator increment
  * simplified a lot of stuff (including `statement_finalizer`)

Leftover questions:
 * How does one handle the amalgamation header? Are we supposed to re-generate it for every PR? If so, then it seems I'm missing something because for me running `python third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .` produces a lot of extra diff. It seems that the amalgamation header isn't really up to date. **Therefore, the diff I have in this PR is manually created**
 * `iterator_t` doesn't have to be copiable in order to satisfy `std::input_iterator`, so in theory we could drop all the `shared_pointer`s. All the tests would pass, but constructing some containers out of `iterator_t` would stop working (eg. `std::vector(view.begin(), view.end())`. Should I go for it and make it not copyable or do you reckon it would break a lot of client code? Iteration similar to the one performed in the tests would remain unaffected.

Hopefully, I got the branch right this time :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,7 @@ if (MSVC)
 	string(REGEX REPLACE "/RTC(su|[1su])" "" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")	
 	add_compile_options(/EHsc)
 	add_compile_options(/Zc:__cplusplus)
+	add_compile_options(/MP) # Allow multi parrallel build
 
 	if ("${CMAKE_GENERATOR}" MATCHES "(Win64|x64)")
 		message(STATUS "Add /bigobj flag to compiler")
--- a/dev/iterator.h
+++ b/dev/iterator.h
@@ -30,8 +30,10 @@ namespace sqlite_orm {
              *  call. When one finishes iterating it the pointer
              *  inside the shared_ptr is nulled out in all copies.
              */
-            std::shared_ptr<sqlite3_stmt*> stmt;
-            view_type& view;
+            std::shared_ptr<statement_finalizer> stmt;
+
+            // only null for the default constructed iterator
+            view_type* view;
 
             /**
              *  shared_ptr is used over unique_ptr here
@@ -39,11 +41,11 @@ namespace sqlite_orm {
              */
             std::shared_ptr<value_type> current;
 
-            void extract_value(std::unique_ptr<value_type>& temp) {
-                temp = std::make_unique<value_type>();
-                auto& storage = this->view.storage;
+            void extract_value() {
+                auto& storage = this->view->storage;
                 auto& impl = storage.template get_impl<value_type>();
-                object_from_column_builder<value_type> builder{*temp, *this->stmt};
+                this->current = std::make_shared<value_type>();
+                object_from_column_builder<value_type> builder{*this->current, this->stmt->get()};
                 impl.table.for_each_column(builder);
             }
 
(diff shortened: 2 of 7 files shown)


andrei-datcu:
I do have some a few unnecessary lines in the diff due to auto formatting (per .clang-format). I will fix those prior merging, but I want to understand the amalgamation problem first, to avoid manually duplicating the changes.

fnc12:
yes single header must be generated every time within every PR. Calling`python third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .` must work. This is how I do it.

Also `std::vector(view.begin(), view.end())` will break some client's code. Can we make it work as before?

andrei-datcu:
> yes single header must be generated every time within every PR. Calling`python third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .` must work. This is how I do it.

Well, then somebody has forgot to do it recently because, running it on a freshly checked-out `dev` branch it will generate a big diff:

```

PS C:\Users\Andrei\Documents\lucru\sqlite_orm> git checkout dev --
Switched to branch 'dev'
Your branch is up to date with 'origin/dev'.
PS C:\Users\Andrei\Documents\lucru\sqlite_orm> python third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .
Creating amalgamation:
 - processing "dev/start_macros.h"
 ...
 - processing "dev/get_prepared_statement.h"
...done!


PS C:\Users\Andrei\Documents\lucru\sqlite_orm> git status
On branch dev
Your branch is up to date with 'origin/dev'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   include/sqlite_orm/sqlite_orm.h

no changes added to commit (use "git add" and/or "git commit -a")
PS C:\Users\Andrei\Documents\lucru\sqlite_orm> git diff --stat
 include/sqlite_orm/sqlite_orm.h | 15398 +++++++++++++++++++++++++++-----------
 1 file changed, 10893 insertions(+), 4505 deletions(-)

```

> Also `std::vector(view.begin(), view.end())` will break some client's code. Can we make it work as before?

It's already working. This would happen if we were to make `iterator_t` non copyable, which I haven't done. So no worriers there, nothing's broken. I will keep it copyable. I'm eagerly looking forward to your review :)

fnc12:
@andrei-datcu ok nice. Big diff is ok. It is always big. Also what OS do you use?

andrei-datcu:
@fnc12 I'm on Win10, but I doubt it's OS specific. It's definitely somebody that forgot to do it in the PR. You can try yourself: pull the dev branch and run the amalgamation script. You should get no differences, yet you get 6000+ lines.

fnc12:
@andrei-datcu ok let's wait for unit test on appveyor finish. Also then I shall pull this branch and check everything on my mac. The bad thing is that travis ended up with working cause they want money. We need to add github actions for ubuntu and mac later.
Anyway thanks for contributing!

andrei-datcu:
@fnc12 You can start pulling, they're passing on my machine :)
The build is pending for the last 3 hours.

fnc12:
@andrei-datcu let me do it tomorrow. It is too late now in my city.

fnc12:
@andrei-datcu thanks

Issue #716, opened 2021-05-03
clang requires std:: namespace prefix for cstdlib imports


Ashoat:
The latest XCode (12.5) started giving compilation errors at callsites of functions imported from `cstdlib`, eg. `atoi`. It looks like the new version of `clang` expects the `std::` prefix to be prepended from `cstdlib` imports (but not `stdlib` imports).

Here is some background:
1. [Explanation](https://github.com/apple/swift-nio-ssl/issues/259#issuecomment-744623667) of what's going on by another team that encountered this issue
2. [Background on `cstdlib`](https://developers.redhat.com/blog/2016/02/29/why-cstdlib-is-more-complicated-than-you-might-think/) and why it is like this

Can be solved either by prefixing all of the calls with `std::`, or by importing from `stdlib` instead of `cstdlib`. (Using the namespace or declaring aliases for it are no-gos given the library is a header file.)

fnc12:
@Ashoat I do not get any errors but I use Xcode 12.5. Maybe you can make a PR or tell me exactly what I need to replace and I'll replace them.

Ashoat:
We were able to simply replace `#include <cstdlib>` with `#include <stdlib.h>` to solve the issue. In our case, we are using this library on iOS and Android, but I don't have enough context to be able to confirm that that's a good solution across all compilation environments that you folks support. That said, based on my understanding of the underlying libraries, I think it should work and is the "correct" thing to do given you're calling the methods without specifying a namespace.

> @Ashoat I do not get any errors but I use Xcode 12.5.

Oh hmm. Must be something else going on in our dev environment that is triggering this. As the issue author the right thing for me to do here would be to create a GitHub repo that reproduces the issue in an empty XCode project, but unfortunately I don't have enough time on my hands to do that right now :(

> Maybe you can make a PR

Happy to make a PR with this single-line change! But it might be difficult for you folks to merge without at least a repro case, right? If you can merge without a repro let me know and I'll submit the PR. If you need a repro before you can merge, no worries – feel free to either close or leave this issue around, and you can get to it when somebody else from the community comes in and provides a repro.

fnc12:
[...]

PR #717, opened 2021-05-03
import stdlib.h instead of cstdlib

Ashoat:
More details in the issue: https://github.com/fnc12/sqlite_orm/issues/716

Technically the headers in `cstdlib` define utilities like `atoi` under the std namespace, so they should be referenced as `std::atoi`. Most compilers don't care about this, but something changed recently in XCode 12.5 and it started complaining on our project.

I generated the amalgamated file by running `third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .`, but let me know if this is supposed to be handled by some build process instead and I can remove it.


fnc12:
@Ashoat thanks

PR #744, opened 2021-06-29
added column_name function

fnc12:


juandent:
how do you use it? Like so?
```C++
storage.column_name( &Claim::id )
```
is this feature new to dev branch? Do I need latest dev branch?

Thanks,
Juan

fnc12:
Yes like this. Yes download the latest dev branch please.

Challanger524:
why still no method like: `vector<string> colnames = storage.column_names<table>();`?

PR #748, opened 2021-07-06
Custom functions2

fnc12:


fnc12:
Finally it happened. This is the first working implementation of user defined functions. Right now on scalar function are implemented. Aggregate functions and windows functions got to be implemented also.

PR #750, opened 2021-07-10
Implemented generated

denzor200:
Without test, without examples


dev/constraints.h
```diff
@@ -75,6 +75,70 @@ namespace sqlite_orm {
             }
         };
 
+        struct generated_always_string {
+            operator std::string() const {
+                return "GENERATED ALWAYS";
```
fnc12, line 80: this struct is used at only one place so it can be removed and its code can be moved to the place of usage

tests/generated.cpp
```diff
@@ -0,0 +1,26 @@
+#include <sqlite_orm/sqlite_orm.h>
+#include <catch2/catch.hpp>
```
fnc12, line 2: please move this file to `tests/constraints`  folder to keep some order

dev/statement_serializator.h
```diff
@@ -812,6 +812,16 @@ namespace sqlite_orm {
             }
         };
 
+        template<class T>
+        struct statement_serializator<generated_always_as_t<T>, void> {
+            using statement_type = generated_always_as_t<T>;
+
+            template<class C>
+            std::string operator()(const statement_type& c, const C& context) const {
+                return static_cast<std::string>(c) + " " + serialize(c.expression, context);
```
fnc12, line 821: it would be great to add some `statement_serializator` tests for `generated_always_as_t`

fnc12:
closing this PR cause it is abandoned by author and cause there is an alternative GENERATED feature implementation

PR #798, opened 2021-10-04
Added the get table info function.

tstrutz:
I needed to query table information from some code I wrote for a project, and namely needed the number of columns present in a given mapped table. This was to populate a bulk insert statement from a template function. This code makes the `table_info` struct accessible from the storage object if needed.

I also added a move constructor to the prepared statement class as I wanted a way to cache PreparedStatements in a map, so I could create them just once. This wasn't working unless I did a move operation on the underlying connection and statement.

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -14634,7 +14634,7 @@ namespace sqlite_orm {
                 auto &tImpl = this->get_impl<object_type>();
                 sqlite3_reset(stmt);
                 for(auto it = statement.t.range.first; it != statement.t.range.second; ++it) {
-                    const auto &o = *it;
+                    auto &o = *it;
                     tImpl.table.for_each_column([&o, &index, &stmt, db](auto &c) {
                         if(!c.template has<constraints::primary_key_t<>>()) {
                             using column_type = typename std::decay<decltype(c)>::type;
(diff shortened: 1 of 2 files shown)


fnc12:
@tstrutz thank you for contributing to this lib. Please change destination branch from `master` to `dev` please. Also it would be better to make your branch from `dev` also

include/sqlite_orm/sqlite_orm.h
```diff
@@ -7186,6 +7186,12 @@ namespace sqlite_orm {
 
             prepared_statement_t(T t_, sqlite3_stmt *stmt_, connection_ref con_) :
                 prepared_statement_base{stmt_, std::move(con_)}, t(std::move(t_)) {}
+
```
fnc12, line 7189: please add this code into `dev/prepared_statement.h` file also. It is better to add new code inside `dev` sources and then call amalgamate script. To run amalgamate script use this command:
```
python third_party/amalgamate/amalgamate.py -c third_party/amalgamate/config.json -s .
```

fnc12 (commented):
please run a code format. You can do it using a command below after running amalgamate script:
```
clang-format -i -style=file include/sqlite_orm/*.h tests/*.cpp tests/*/*.cpp dev/*.h dev/*/*.h tests/*/*.h examples/*.cpp examples/*/src/*.cpp
```

tstrutz:
> @tstrutz thank you for contributing to this lib. Please change destination branch from `master` to `dev` please. Also it would be better to make your branch from `dev` also

Having issues with getting merge conflicts from `dev` so I'm going to remake this and resubmit starting from `dev`.

PR #831, opened 2021-11-10
normalized paths in amalgamation script

### LINKED ISSUE #830: Amalgamation script must normalize file paths
The amalgamation python script is keeping a list of already included files in order to determine whether it needs to paste an include into the target file _sqlite_orm.h_.
For this to function properly the script must normalize the paths, otherwise the key for the path list might differ during the amalgamation process even when referring to the same file. This leads to files being pasted multiple times.

Consider:
- Differing include paths depending on the directory structure (`#include "../serialize_result_type.h"` vs `#include "serialize_result_type.h"`)
- Combining the include with configured source/include paths leads to a mix of '/' and '\\' path separators on Windows.

Also, an early normalization leads to verbose output being more natural on Windows.

firedaemon-cto:
Fixes #830.

### DIFF
--- a/include/sqlite_orm/sqlite_orm.h
+++ b/include/sqlite_orm/sqlite_orm.h
@@ -153,7 +153,7 @@ namespace std {
 #include <tuple>  //  std::tuple, std::get, std::tuple_element, std::tuple_size
 #include <type_traits>  //  std::false_type, std::true_type
 
-// #include "static_magic.h"
+// #include "../static_magic.h"
 
 #include <type_traits>  //  std::false_type, std::true_type, std::integral_constant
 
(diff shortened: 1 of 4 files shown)


include/sqlite_orm/sqlite_orm.h
```diff
@@ -6161,22 +6161,6 @@ namespace sqlite_orm {
 
 // #include "../serialize_result_type.h"
 
-#ifdef SQLITE_ORM_STRING_VIEW_SUPPORTED
```
fnc12, line 6164: why this is disappeared?

fnc12 (changes requested):
please fix format

fnc12:
@firedaemon-cto thanks for improving this lib. Probably you'll be interested in this https://github.com/fnc12/sqlite_orm/wiki/Contribution-guide

include/sqlite_orm/sqlite_orm.h
firedaemon-cto: Because the incuded file serialize_result_type.h appears already on line [4120](https://github.com/fnc12/sqlite_orm/blob/1895b18832b4b36da82b14c3ddbcf532ba01f53a/include/sqlite_orm/sqlite_orm.h#L4120).

You just didn't notice until this PR :)

fnc12: oh nice. Thanks

PR #833, opened 2021-11-10
Support for 'pointer-passing interface'

trueqbit:
sqlite offers a powerful feature called ['pointer-passing interface'](https://sqlite.org/bindptr.html). It allows for functions to work with native datatypes, i.e. accept them as arguments or return them to other functions.

One example is to extract a ASN1_TIME object from a certificate as a 'pointer value', then send it to a comparison function.

This makes it very simple to write a domain-specific set of functions instead of writing a complete sqlite extension.

Not sure why github displays a diff for commits that where already merged via other PRs - here's a pristine diff: https://github.com/fnc12/sqlite_orm/compare/dev...FireDaemon:feature/pointer-values

### DIFF
--- a/dev/arithmetic_tag.h
+++ b/dev/arithmetic_tag.h
@@ -1,4 +1,5 @@
 #pragma once
+#include <type_traits>
 
 namespace sqlite_orm {
 
--- a/dev/carray.h
+++ b/dev/carray.h
@@ -14,15 +14,15 @@ namespace sqlite_orm {
     using carray_trule = pointer_trule<P, carray_pvt, D>;
 
     /**
-     *  SQL function that is a pass-through
-     *  for values (it returns a copy of its argument) but also saves the
+     *  SQL function that is a pass-through for values
+     *  (it returns its argument unchanged using move semantics) but also saves the
      *  value that is passed through into a C-language variable.
      */
     template<typename P>
     struct note_value_fn {
         using pointer_value_t = carray_value<P>;
 
-        P operator()(P value, pointer_value_t pv) const {
+        P operator()(P&& value, pointer_value_t pv) const {
             if (P* p = pv) {
                 *p = value;
             }
--- a/dev/carray.h
+++ b/dev/carray.h
@@ -9,23 +9,50 @@ namespace sqlite_orm {
 
     SQLITE_ORM_INLINE_VAR constexpr const char carray_pvt_name[] = "carray";
     using carray_pvt = std::integral_constant<const char*, carray_pvt_name>;
+
     template<typename P>
-    using carray_value = pointer_value<P, carray_pvt>;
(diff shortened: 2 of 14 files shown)


trueqbit:
I still would like to discuss the names that I have chosen, especially for:

1. `pointer_value`
2. `pointer_trule`

Their purpose:
1. `pointer_value` is intended to be used in functions that accept a 'pointer value' as an argument.
2. `pointer_trule` carries a deleter function, and is intended to be used ...
  a. for binding a 'pointer value' as part of a statement (to an sqlite placeholder) or ...
  b. for returning a 'pointer value' from a function.


In order to improve understanding:
1. `pointer_value` could be renamed to or at least be aliased by `pointer_argument`.
2. A better name for `pointer_trule` could be `bindable_pointer`.

fnc12:
`pointer_argument` and `bindable_pointer` are better cause `pointer_value` sounds very common and `pointer_trule` sounds like something unknown. Also I'd like you to provide a useful example of this. Also please review my comments about moving classes to `internal` namespace and adding builder functions in `sqlite_orm` namespace instead

trueqbit:
> [...] Also please review my comments about moving classes to `internal` namespace and adding builder functions in `sqlite_orm` namespace instead

Where do I find your comments?

fnc12:
> > [...] Also please review my comments about moving classes to `internal` namespace and adding builder functions in `sqlite_orm` namespace instead
> 
> Where do I find your comments?

in the code

trueqbit:
@fnc12 Thx for your feedback!

I have renamed `pointer_value` -> `pointer_arg` and `pointer_trule` -> `pointer_binding`, also I have introduced the builder functions `bindable_pointer()` and `statically_bindable_pointer()`.

However, I think that `pointer_binding` needs to remain part of the public interface:

1. `pointer_binding` will be returned from functions yielding a bindable 'pointer value'.
2. domain-specific code aliases `pointer_binding`. [see _carray.h_]
3. domain-specific scalar/aggregate functions are returning it. [see _carray.h_]

fnc12:
@trueqbit nice work! But IMHO `pointer_binding` can be moved to `internal` namespace

fnc12:
@trueqbit ok I see. Thank you!

fnc12:
I'd like to make a tweet about feature your made here https://twitter.com/sqlite_orm. If you have a tweeter account I can add @ with it.

trueqbit:
Thx a lot @fnc12, my twitter account is ktriendl, I just don't really use it :)

You were quick merging the PR! I still owe the library a good example.

fnc12:
I thought that it is time cause all checks are passed and now issues left during review. Anyway feel free to make a new PR right from this or another branch.
Also I've tweeted about this feature here https://twitter.com/sqlite_orm/status/1460158768885510145?s=21

PR #848, opened 2021-11-24
Taking bindable expressions in order-by clauses into consideration

trueqbit:


fnc12:
@trueqbit thanks! I appreciate your work

Issue #849, opened 2021-11-25
Allow for functions w/o arguments


trueqbit:
There's nothing in sql/sqlite or sqlite_orm that would prohibit functions w/o arguments.
But compilers generate at least a warning if not an error at the point of creating a function w/o arguments:

```c++
struct generate_value_fn {
    int64 operator()() const {
        return 42;
    }

    static const char* name() {
        return "generate_value";
    }
};

storage.create_scalar_function<generate_value_fn>();
```

> 1>sqlite_orm\dev\storage_base.h(217,71): warning C4309: 'specialization': truncation of constant value
> 1>main.cpp(98): message : see reference to function template instantiation 'void sqlite_orm::internal::storage_base::create_scalar_function<main::generate_value_fn>(void)' being compiled

PR follows 😃

fnc12:
never saw this kind of warning. What compiler and what OS do you use?

fnc12:
@trueqbit merged. Thanks!

trueqbit:
> never saw this kind of warning. What compiler and what OS do you use?

I am using Visual C++ 2019 16.11, but I also tried with the clang llvm toolset that ships with VC++. It probably depends on compiler flags.

PR #850, opened 2021-11-25
Allow for functions w/o arguments

trueqbit:
Addresses #849


fnc12:
Oh looks like it is `int` to `size_t` implicit conversion warning. Thank you

Issue #851, opened 2021-11-25
Validating pointer value types


trueqbit:
Now that sqlite_orm features the neat 'pointer-passing interface' a bonus would be to validate the bound pointer types against the function arguments.

In order to prevent pointer leaks, the sqlite engine itself checks whether a bound pointer's type string equals a function argument's pointer type string.

We can achieve this at compilation time when creating the `function_call<>` node.

fnc12:
@trueqbit is this issue actual?

trueqbit:
Since the PR has been merged we can close this issue, unless you have experienced other compilation issues...

PR #852, opened 2021-11-25
Validating pointer value types

trueqbit:
Adressing #851, this PR introduces validation of bound pointer values against the expected pointer value taken as function arguments.


trueqbit:
Sorry, wrong target branch.

PR #860, opened 2021-12-05
Amending pointer argument validation

trueqbit:
Addresses discussions/#857.

* Addresses warnings found with clang.
* Compilers seem to disagree about short-circuiting invocation of constexpr/consteval functions.

Additionally, I introduced 'polyfilling' useful generic type traits missing in previous C++ standards, currently in namespace `sqlite_orm::internal::polyfill`. @fnc12 Please review whether you are happy with this approach.

### DIFF
--- a/dev/cxx_polyfill.h
+++ b/dev/cxx_polyfill.h
@@ -1,0 +1,44 @@
+#pragma once
+
+#include <type_traits>
+
+#include "start_macros.h"
+
+namespace sqlite_orm {
+    namespace internal {
+        namespace polyfill {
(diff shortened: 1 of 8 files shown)


fnc12:
Doesn't compile. What is 'malloc.h'? Looks like Windows only header
Снимок экрана 2021-12-06 в 18 41 50

fnc12 (changes requested):
Cannot compile on mac using Xcode. Screenshot with error is attached in main thread

fnc12 (approved):

PR #877, opened 2022-01-12
added serialization support without question marks

fnc12:


tests/storage_non_crud_tests.cpp
```diff
@@ -602,13 +602,6 @@ TEST_CASE("Explicit insert") {
                 REQUIRE(visitFromStorage.usedId() == visit.usedId());
                 storage.remove<Visit>(visitFromStorage.usedId());
             }
-            SECTION("setter") {
```
fnc12, line 605: I know it is not a good practice. But I reckon that insert explicit is used by almost no one. Especially with setters. If someone does - sorry =)

PR #931, opened 2022-03-03
#929 Corrected serializing nullptr, nullopt, wide strings

trueqbit:

PR #934, opened 2022-03-05
Serializer for explicit column dosn't find correct column

### LINKED ISSUE #933: Serializer for explicit column dosn't find correct column
There's a line for the statement serializer of a column_pointer that tries to get the column name like this:
```c++
std::string statement_serializator<column_pointer<T, F>>::operator()(const column_pointer<T, F>& cp, const C& context) const {
    // ...
    if(auto columnNamePointer = context.impl.column_name_simple(cp.field)) {
    }
    // ...
}
```
This however is wrong for two reasons:
1. `storage_impl<>::column_name_simple()` doesn't walk the 'table storage chain'
2. the serializer should actually call `storage_impl<>::column_name(cp)`

I guess this was just an oversight in the first place as `column_name(column_pointer<>)` exists but remained unused, and the original intention was to use it.

trueqbit:
Fixes #933.


dev/type_traits.h
```diff
@@ -0,0 +1,43 @@
...
+        template<template<typename...> class Op, class... Args>
+        using satisfies_not = std::enable_if_t<std::negation<Op<Args...>>::value, bool>;
+
+        // enable_if for functions
+        template<class T, template<typename...> class Primary>
+        using satisfies_is_specialization_of = std::enable_if_t<polyfill::is_specialization_of_v<T, Primary>, bool>;
+    }
+
+    // type traits not quite polyfill, however used throughout the program
+    namespace internal {}
```
fnc12, line 36: what is the purpose of empty namespace?

fnc12:
Снимок экрана 2022-03-06 в 00 19 56
doesn't compile on my mac

fnc12 (changes requested):

dev/type_traits.h
trueqbit: This as a placeholder for upcoming things (like I have in my CTE branch).
Shall I remove it nevertheless?

fnc12: better be removed cause we don't pay for what we don't use. Please add it once it has any content

trueqbit:
> Снимок экрана 2022-03-06 в 00 19 56
> doesn't compile on my mac

Hmm, do you know where this comes from? It must have been already the case in the previously merged PR #931.

fnc12:
> > Снимок экрана 2022-03-06 в 00 19 56
> > doesn't compile on my mac
> 
> Hmm, do you know where this comes from? It must have been already the case in the previously merged PR #931.
> 
> Is it a real compiler error or just a failed intellisense check?

it is a compiler error.

fnc12:
@trueqbit try to add static unit tests for `is_bindable_v` and `is_printable_v`. Looks like that `statement_serializator<std::wstring>` is not instantiated

fnc12:
@trueqbit  do you have clang compiler on your machine?

fnc12:
@trueqbit same error appears in both cases when I add `SQLITE_ORM_OMITS_CODECVT` and when I remove it

trueqbit:
It happens only when you leave `SQLITE_ORM_OMITS_CODECVT ` undefined on non-Windows operating systems.

I know the reason - I was trying to be smart and make `statement_binder<const wchar_t*>` dependent on `sizeof(wchar_t)==2` instead of `SQLITE_ORM_OMITS_CODECVT `, in order to allow binding wide strings even if the STL converter isn't available. So if the `TEST_CASE("Wide string")` in _tests/tests3.cpp_ is enabled but the binder for wchar_t is disabled then there's no serializer.

I did this because I all the wide-character code in sqlite_orm assumes utf-16, however on non-Windows operating systems wchar_t is 32-bit and is assumed to be utf-32, usually.

I think I'll revert for this PR, but we should clarify the usage of wide characters, independent of SQLITE_ORM_OMITS_CODECVT.

PR #944, opened 2022-03-08
Amend serializing bindables

trueqbit:
Amends #931.

Checks whether an optional's underlying type is bindable.

* Added statically executed unit test for is_printable.
* Amended static unit test for is_bindable.
* Took care of optional's possibly const-qualified type (so, `std::unique_ptr<const int>` is possible).
* Removed binding an optional's underlying value as a return value (wrong and untested anyway).

@fnc12 Let me know what your code checker reports for
```c++
struct is_std_ptr<std::unique_ptr<T>> {
    static auto make(std::remove_cv_t<T>&& v) {
        return std::make_unique<T>(std::move(v));
    }
}
```


fnc12:
UT doesn't run well

trueqbit:
It's good to go now.

fnc12:
everything ran well. Merging. Thanks Klaus

PR #966, opened 2022-03-17
Order by kth column

trueqbit:
Addresses #950.


dev/conditions.h
```diff
@@ -1171,12 +1172,50 @@ namespace sqlite_orm {
...
     /**
      * ORDER BY column
-     * Example: storage.select(&User::name, order_by(&User::id))
+     * Examples:
+     * storage.select(&User::name, order_by(&User::id))
+     * storage.select(&User::name, order_by(1_nth_col))
      */
     template<class O>
     internal::order_by_t<O> order_by(O o) {
+        static_assert(!std::is_arithmetic<O>::value && !std::is_base_of<std::string, O>::value &&
```
fnc12, line 1204: actually one can bind any custom type using custom type binding technique. E.g. any custom type like `QInt`, `QString` or whatever but he/she will not be able to use these types because you prohibited it here. Assuming literals is not what ORM should do because it looks like an artificial prohibiting of some functional

fnc12 (changes requested):
I am not sure that `1_nth_col` must be added because when user wants to write `ORDER BY 1` using `sqlite_orm` then he/she will write `order_by(1)` by default. It is not clear for users why `order_by(1)` will not work but `order_by(1_nth_col)` will so it would be better to support `order_by(1)` instead

dev/conditions.h
trueqbit: No, the static assert does not prohibit custom types. Only arithmetic and strings.

fnc12 (changes requested):

dev/conditions.h
trueqbit: It should actually ensure it's not a bindable `static_assert(!is_bindable_v<O>, "")`.

fnc12:
I'll review once https://github.com/fnc12/sqlite_orm/issues/977 is closed

examples/column_aliases.cpp
```diff
@@ -0,0 +1,76 @@
...
+            cout << get<0>(row) << '\t' << get<1>(row) << '\n';
+        }
+    }
+    cout << endl;
+    {
+        //  SELECT name, instr(abilities, 'o')
+        //  FROM marvel
+        //  ORDER BY 2
+        auto rows = storage.select(columns(&MarvelHero::name, as<colalias_i>(instr(&MarvelHero::abilities, "o"))),
+                                   order_by(get<colalias_2>()));
```
fnc12, line 60: `order_by(2)` is more intuitive and available to implement. `get<colalias_2>()` looks weird for newbies and raises questions

fnc12 (changes requested):
Let's leave API like `order_by(2)`. Let's add an overload for `order_by` for bindables or types associated with `integer_printer` but let's leave API like this cause in SQL API is the same for both
1) `ORDER BY 2`
2) `ORDER BY CASE ... END`

trueqbit:
Alright, there's now an overload for `order_by()` for types associated with `integer_printer`.

fnc12:
thanks @trueqbit

PR #990, opened 2022-03-29
added three more APIs for transactions

fnc12:


juandent:
Hi!

I still see ```begin_transaction_internal``` not using its parameter ```query```.. so exclusive transaction is not yet implemented?

fnc12:
lol it is an oversight. I'll fix it right away

Issue #989, opened 2022-03-28
Quoting of table identifiers is wrong


trueqbit:
The SQLite grammar dictates that identifiers must be quoted using double-quotes; [reference](https://www.sqlite.org/lang_keywords.html).

This is done correctly for column identifiers but not for table identifiers or aliases.

Additionally:
- SQLite recommends quoting all identifiers, such that possible conflicts with current and future keywords are avoided.
- when dumping a statement we must properly escape single-quotes of strings. The serializer for 'bindables' currently doesn't do that [see e.g. C string](https://github.com/fnc12/sqlite_orm/blob/e43103aade8c3bcb693bf677b3a3cbf48bd1b8e5/dev/statement_serializator.h#L91).

An example out of unit tests:
```sql
REPLACE INTO users SELECT 'users_backup'."id", 'users_backup'."name" FROM 'users_backup'
```
must be corrected to
```sql
REPLACE INTO "users" SELECT "users_backup"."id", "users_backup"."name" FROM "users_backup"
```

While we are at it, I highly recommend using utility functions for quoting and escaping.

fnc12:
Yes Klaus you are right. Incorrect quotes is one more topic of legacy. It is got to be fixed. Utility functions is also a good idea.

PR #993, opened 2022-03-29
Corrected quoting of identifiers

trueqbit:
Addresses #989.

Corrected quoting column, index, table, alias identifiers, and serializing string literals, which must be escaped.

Additionally:
* simplified or streamlined joining multiple serialized arguments
* utility functions for quoting and escaping (I thought it's useful to provide them in the [public namespace](https://github.com/FireDaemon/sqlite_orm/blob/31d69b0684bbd282e940feec08d5efce63388d71/dev/util.h#L46))
* Fixed serializing `remove_all_t` conditions 

@fnc12 On another note, I wonder what the double-check of the collected table names in the 'update all' serializer should be corrected to; [see](https://github.com/fnc12/sqlite_orm/blob/45818cb388d6a57f49fcf97ef72cb6377d6ce0bc/dev/statement_serializer.h#L1251).
I simplified the nested if-statements a bit, which makes it easier to understand that the same condition leads to different exceptions, [see](https://github.com/FireDaemon/sqlite_orm/blob/31d69b0684bbd282e940feec08d5efce63388d71/dev/statement_serializer.h#L1281).

```c++
if(collector.table_names.empty()) {
    throw std::system_error{orm_error_code::no_tables_specified};
}
if(collector.table_names.size() < 1) {
    throw std::system_error{orm_error_code::incorrect_set_fields_specified};
}
```

Though this might be ignored in this PR, I recognized the oddity.


dev/statement_serializer.h
```diff
@@ -1241,42 +1270,39 @@ namespace sqlite_orm {
...
+                    auto leftContext = context;
+                    leftContext.skip_table_name = true;
+                    iterate_tuple(upd.set.assigns, [&context, &leftContext, &setPairs](auto& asgn) {
+                        std::stringstream sss;
+                        sss << serialize(asgn.lhs, leftContext);
+                        sss << ' ' << asgn.serialize() << ' ';
+                        sss << serialize(asgn.rhs, context);
+                        setPairs.push_back(sss.str());
+                    });
+                    ss << join_serialized_strings(setPairs);
```
fnc12, line 1300: let's avoid using of 'join' algorithm please

dev/statement_serializer.h
```diff
@@ -1290,20 +1316,20 @@ namespace sqlite_orm {
...
+                ss << "INSERT INTO " << quote_identifier(tImpl.table.name) << " ";
 
                 auto columnIndex = 0;
                 tImpl.table.for_each_column([&tImpl, &columnIndex, &ss](auto& column) {
                     using table_type = typename std::decay<decltype(tImpl.table)>::type;
                     if(table_type::is_without_rowid ||
                        (!column.template has<primary_key_t<>>() &&
                         !tImpl.table.exists_in_composite_primary_key(column) && !column.is_generated())) {
-                        if(0 == columnIndex) {
+                        if(columnIndex == 0) {
```
fnc12, line 1327: why?

dev/statement_serializer.h
```diff
@@ -1509,30 +1532,21 @@ namespace sqlite_orm {
...
+                ss << "INSERT INTO " << quote_identifier(tImpl.table.name) << " ";
                 std::vector<std::string> columnNames;
 
                 tImpl.table.for_each_column([&columnNames, &tImpl](auto& column) {
                     using table_type = typename std::decay<decltype(tImpl.table)>::type;
                     if(table_type::is_without_rowid ||
                        (!column.template has<primary_key_t<>>() &&
                         !tImpl.table.exists_in_composite_primary_key(column) && !column.is_generated())) {
-                        columnNames.emplace_back(column.name);
+                        columnNames.push_back(column.name);
```
fnc12, line 1543: why?

dev/statement_serializer.h
```diff
@@ -1748,13 +1745,7 @@ namespace sqlite_orm {
...
                     ss << static_cast<std::string>(distinct(0)) << " ";
                 }
-                auto columnNames = get_column_names(sel.col, context);
-                for(size_t i = 0; i < columnNames.size(); ++i) {
-                    ss << columnNames[i];
-                    if(i < columnNames.size() - 1) {
-                        ss << ", ";
-                    }
-                }
+                ss << join_serialized_strings(get_column_names(sel.col, context));
```
fnc12, line 1748: join here also got to be optimized

dev/statement_serializer.h
```diff
@@ -1767,24 +1758,12 @@ namespace sqlite_orm {
...
-                            auto& tableNamePair = tableNames[i];
-                            ss << "'" << tableNamePair.first << "'";
-                            if(!tableNamePair.second.empty()) {
-                                ss << " '" << tableNamePair.second << "'";
-                            }
-                            if(int(i) < int(tableNames.size()) - 1) {
-                                ss << ", ";
-                            }
-                        }
+                        ss << " FROM " << serialize_identifiers(collector.table_names);
```
fnc12, line 1766: I guess `for` with `stringstream` is more effective here

dev/statement_serializer.h
```diff
@@ -1856,12 +1835,7 @@ namespace sqlite_orm {
                         whereString = move(columnName);
                     }
                 });
-                for(size_t i = 0; i < columnNames.size(); ++i) {
-                    ss << columnNames[i];
-                    if(i < columnNames.size() - 1) {
-                        ss << ", ";
-                    }
-                }
+                ss << join_serialized_strings(columnNames);
```
fnc12, line 1838: join

dev/statement_serializer.h
```diff
@@ -2119,18 +2093,12 @@ namespace sqlite_orm {
...
                 });
-                ss << static_cast<std::string>(orderBy) << " ";
-                for(size_t i = 0; i < expressions.size(); ++i) {
-                    ss << expressions[i];
-                    if(i < expressions.size() - 1) {
-                        ss << ", ";
-                    }
-                }
-                ss << " ";
+                ss << static_cast<std::string>(orderBy) << " " << join_serialized_strings(expressions) << " ";
```
fnc12, line 2101: join also

dev/statement_serializer.h
```diff
@@ -2433,17 +2389,7 @@ namespace sqlite_orm {
...
-                    auto vectorSize = statement.vector.size();
-                    for(decltype(vectorSize) index = 0; index < vectorSize; ++index) {
-                        auto& value = statement.vector[index];
-                        ss << serialize(value, context);
-                        if(index < vectorSize - 1) {
-                            ss << ", ";
-                        }
-                    }
-                }
+                ss << "VALUES " << serialize_dynamic_args(statement.vector, context);
```
fnc12, line 2392: join algo is here also

dev/storage_base.h
```diff
@@ -149,9 +149,9 @@ namespace sqlite_orm {
                     sql.c_str(),
                     [](void* data, int argc, char** argv, char** /*columnName*/) -> int {
                         auto& tableNames_ = *(data_t*)data;
-                        for(int i = 0; i < argc; i++) {
+                        for(int i = 0; i < argc; ++i) {
                             if(argv[i]) {
-                                tableNames_.push_back(argv[i]);
+                                tableNames_.emplace_back(argv[i]);
```
fnc12, line 154: why? what's the difference?

dev/table_name_collector.h
```diff
@@ -33,16 +33,12 @@ namespace sqlite_orm {
 
             template<class F, class O>
             void operator()(F O::*, std::string alias = {}) const {
-                if(this->find_table_name) {
-                    table_names.insert(std::make_pair(this->find_table_name(typeid(O)), move(alias)));
-                }
+                table_names.emplace(this->find_table_name(typeid(O)), move(alias));
```
fnc12, line 36: best option is throwing an exception if `!this->find_table_name` cause if it is `null` then we get UB. We must prefer exceptions instead of UB

dev/table_name_collector.h
```diff
@@ -33,16 +33,12 @@ namespace sqlite_orm {
...
             }
 
             template<class T, class F>
             void operator()(const column_pointer<T, F> &) const {
-                if(this->find_table_name) {
-                    table_names.insert({this->find_table_name(typeid(T)), ""});
-                }
+                table_names.emplace(this->find_table_name(typeid(T)), "");
```
fnc12, line 41: same here with `this->find_table_name`

dev/table_name_collector.h
```diff
@@ -52,63 +48,44 @@ namespace sqlite_orm {
 
             template<class T>
             void operator()(const count_asterisk_t<T> &) const {
-                if(this->find_table_name) {
-                    auto tableName = this->find_table_name(typeid(T));
-                    if(!tableName.empty()) {
-                        table_names.insert(std::make_pair(move(tableName), ""));
-                    }
+                auto tableName = this->find_table_name(typeid(T));
```
fnc12, line 51: same

dev/table_name_collector.h
```diff
@@ -52,63 +48,44 @@ namespace sqlite_orm {
...
                 }
             }
 
             template<class T, satisfies_not<std::is_base_of, alias_tag, T> = true>
             void operator()(const asterisk_t<T> &) const {
-                if(this->find_table_name) {
-                    auto tableName = this->find_table_name(typeid(T));
-                    table_names.insert(std::make_pair(move(tableName), ""));
-                }
+                table_names.emplace(this->find_table_name(typeid(T)), "");
```
fnc12, line 59: same

dev/table_name_collector.h
```diff
@@ -52,63 +48,44 @@ namespace sqlite_orm {
...
-                    // note: not all alias classes have a nested A::type
-                    static_assert(polyfill::is_detected_v<type_t, T>,
-                                  "alias<O> must have a nested alias<O>::type typename");
-                    auto tableName = this->find_table_name(typeid(type_t<T>));
-                    table_names.insert(std::make_pair(move(tableName), alias_extractor<T>::get()));
-                }
+                // note: not all alias classes have a nested A::type
+                static_assert(polyfill::is_detected_v<type_t, T>,
+                              "alias<O> must have a nested alias<O>::type typename");
+                auto tableName = this->find_table_name(typeid(type_t<T>));
```
fnc12, line 67: you know about it 🙂. It is related to all functions in this file

dev/util.h
```diff
@@ -1,10 +1,53 @@
...
 
 namespace sqlite_orm {
 
+    /** 
+         *  Escape the provided character in the given string by doubling it.
+         *  @param str A copy of the original string
+         *  @param char2Escape The character to escape
+         */
+    inline std::string sql_escape(std::string str, char char2Escape) {
+        for(size_t pos = 0; (pos = str.find(char2Escape, pos)) != std::string::npos; pos += 2) {
```
fnc12, line 17: btw `std::string::npos` can be replaced with `str.npos`

dev/util.h
```diff
@@ -13,7 +56,7 @@ namespace sqlite_orm {
             }
         }
 
-        static void perform_void_exec(sqlite3* db, const std::string& query) {
+        inline void perform_void_exec(sqlite3* db, const std::string& query) {
```
fnc12, line 59: why did you change `static` to `inline`? `static` was placed to make this function file private

dev/util.h
trueqbit: Yes it can, but it doesn't have to :)

dev/statement_serializer.h
```diff
@@ -37,11 +38,71 @@
...
+                quoted = quote_identifier(qualifier) + "." + move(quoted);
+            }
+            if(!alias.empty()) {
+                quoted += " " + quote_identifier(alias);
+            }
+            return quoted;
+        }
+
+        inline std::string serialize_identifier(const std::string& identifier, const std::string& alias) {
+            std::string quoted = quote_identifier(identifier);
```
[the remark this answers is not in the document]
trueqbit: Yes, true, however it's a kind of a shortcut - why calling the other function with useless temporary objects?

dev/statement_serializer.h
```diff
@@ -231,20 +301,14 @@ namespace sqlite_orm {
...
                     args.push_back(serialize(v, context));
                 });
-                for(size_t i = 0; i < args.size(); ++i) {
-                    ss << args[i];
-                    if(i < args.size() - 1) {
-                        ss << ", ";
-                    }
-                }
-                ss << ")";
+                ss << "(" << join_serialized_strings(args) << ")";
```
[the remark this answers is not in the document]
trueqbit: Serialization isn't very allocation friendly anyway - if you look at a serialization callstack and the fact that `serialize()` always returns a string.

Anyway, you have a valid point, so just confirming whether this is important, as it is simple to change it.

dev/table_name_collector.h
trueqbit: No, calling an "empty" function is not UB, the function wrapper will throw an exception.
Otherwise there's no need checking `this->find_table_name == nullptr` all the time. It's an internal class anyway, and should be used correctly in the first place. An assertion would be more helpful than a repetitive runtime check that only grows your code size.
The only place where `table_name_collector` gets ever used directly (circumventing the collection callbacks) is in `serialize_get_all_impl`, which then also performs an AST iteration that goes into the void.

dev/util.h
trueqbit: I just aligned it to all other functions in sqlite_orm. `perform_step()` in the very same file is inline, as are all other free-standing functions.
Also because it's in the internal namespace there's no need to make it linker private?

So, counter question: why would you not use `inline`?

dev/statement_serializer.h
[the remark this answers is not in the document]
fnc12: because if you decide to improve you serialize algorithm inside these functions you'll need to improve it in several times. You'll do copy and paste code. But if you write one thing at one place then you'll never face copy and paste code

dev/statement_serializer.h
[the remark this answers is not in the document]
fnc12: serialization can be performed with only one allocation if we refactor algorithm some day. One day somebody can ask for this cause this is what C++ devs do sometimes - they try to improve algorithms to the highest performance.

dev/storage_base.h
fnc12: I bet it doesn't change the amount of heap allocations but ok

dev/table_name_collector.h
fnc12: ok I did not know about `[bad_function_call](https://www.cplusplus.com/bad_function_call)` exception. It is ok. Thanks

dev/util.h
fnc12: we can use `inline` it is ok. The question is about `static`. It is important to make this function file private to eliminate opportunities to run string queries for users. Private `internal` namespace is not private actually cause C++ doesn't have private namespaces. And C++ doesn't have private non-class functions. That is why this function is `static`

trueqbit: Oh I understand your point. So you want to make it private in regard to linking as well and prohibit the compiler from accidentally creating a full definition that the linker exports? Or otherwise what's your notion of "private", or "users"? Even a static function isn't private in C++ when not defined in a .cpp file.

fnc12: 1) "private" is the same as "private" in C#, Swift or any other popular languages.
2) "users" = developers who uses `sqlite_orm`
3) if static function in '.h' file is accessible then ignore my words. Then I forgot about it

dev/storage_base.h
trueqbit: Correct. It potentially eliminates move operations. It's one of those changes that I did while walking through all the code.

dev/statement_serializer.h
```diff
@@ -45,81 +44,222 @@ namespace sqlite_orm {
...
+        // serialize and stream a tuple of action expressions;
+        // space-separated
+        template<class T, class Ctx>
+        std::ostream& operator<<(std::ostream& ss, std::tuple<decltype((streaming_actions_tuple)), T, Ctx> tpl) {
+            const auto& actions = std::get<1>(tpl);
+            auto& context = std::get<2>(tpl);
+
+            bool first = true;
+            iterate_tuple(actions, [&ss, &context, &first](auto& a) {
+                constexpr const char* sep[] = {" ", ""};
```
fnc12, line 143: why don't you use `std::array`?

dev/statement_serializer.h
```diff
@@ -45,81 +44,222 @@ namespace sqlite_orm {
...
+        //
+        // comma-separated
+        template<class C>
+        std::ostream& operator<<(std::ostream& ss, std::tuple<decltype((streaming_identifiers)), C> tpl) {
+            const auto& identifiers = std::get<1>(tpl);
+
+            constexpr const char* sep[] = {", ", ""};
+            bool first = true;
+            for(auto& identifier: identifiers) {
+                ss << sep[std::exchange(first, false)];
```
fnc12, line 232: this is a very fucking elegant code

dev/statement_serializer.h
trueqbit: Built-in array is readily available and always has automatic size deduction. If you prefer I can change it.

dev/statement_serializer.h
```diff
@@ -45,81 +44,222 @@ namespace sqlite_orm {
...
+            expressions_tuple,
+            dynamic_expressions,
+            serialized,
+            identifier,
+            identifiers,
+            values_placeholders,
+        };
+
+        template<stream_as mode>
+        struct streaming_ {
```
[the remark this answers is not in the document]
trueqbit: You are asking tough questions that force me to verifying my own subtle choices 🤔 
Did you see the constexpr instantiations below, e.g. `streaming_expressions_tuple`? The underscore fits with the continuation in whatever words you append to it - `streaming_...`
Though I realize it might escape usual naming conventions. Do you have a better suggestion?

dev/statement_serializer.h
fnc12: it is not urgent. But if you are going to make any changes in this PR please prefer using objects instead of pure C types cause we use C++ with OOP

dev/statement_serializer.h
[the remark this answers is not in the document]
fnc12: yes: `streaming`. Underscore in the end is used in `sqlite_orm` only if omitting underscore will make a reserved C++ keyword. This is how it work in `boost` also

dev/statement_serializer.h
trueqbit: Thx! It's tight. The only little issue one might argue that the exchange operation involves a temporary variable.

dev/statement_serializer.h
```diff
@@ -1260,42 +1399,36 @@ namespace sqlite_orm {
...
-                        return ss.str();
-                    } else {
-                        throw std::system_error{orm_error_code::no_tables_specified};
-                    }
-                } else {
+
+                if(collector.table_names.empty()) {
+                    throw std::system_error{orm_error_code::no_tables_specified};
+                }
+                if(collector.table_names.size() < 1) {
```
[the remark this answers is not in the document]
fnc12: intention was to avoid calling `UPDATE` with several tables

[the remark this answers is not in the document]
fnc12: later it changed and calling `UPDATE` with 1+ tables became available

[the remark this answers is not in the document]
fnc12: current intention (from [previous code](https://github.com/fnc12/sqlite_orm/blob/45818cb388d6a57f49fcf97ef72cb6377d6ce0bc/dev/statement_serializer.h#L1251)) is designed to cover cases when expression has 0 tables mentioned. Theoretically it can happen. I wrote a code to handle such cases

dev/statement_serializer.h
```diff
@@ -52,6 +54,282 @@ namespace sqlite_orm {
...
+        inline void stream_identifier(std::ostream& ss,
+                                      const std::string& qualifier,
+                                      const std::string& identifier,
+                                      const std::string& alias) {
+            constexpr char quoteChar = '"';
+
+            if(!qualifier.empty()) {
+                ss << quoteChar << sql_escape(qualifier, quoteChar) << std::string{quoteChar, '.'};
+            }
+            ss << quoteChar << sql_escape(identifier, quoteChar) << quoteChar;
```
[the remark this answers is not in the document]
fnc12: Are you sure that identifiers can contain double quotes in SQLite?

tests/statement_serializer_tests/statements/remove.cpp
```diff
@@ -11,19 +11,20 @@ TEST_CASE("statement_serializer remove") {
...
     auto table = make_table("users", make_column("id", &User::id, primary_key()), make_column("name", &User::name));
     using storage_impl_t = internal::storage_impl<decltype(table)>;
-    auto storageImpl = storage_impl_t{table};
+    storage_impl_t storageImpl{table};
     using context_t = internal::serializer_context<storage_impl_t>;
     context_t context{storageImpl};
-    std::string value;
-    decltype(value) expected;
+
+    std::string value, expected;
```
fnc12, line 18: please avoid declaring more than one var within one statement - it is a very rare technique and it makes code less recognizable by most of developers

dev/statement_serializer.h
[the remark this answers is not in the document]
trueqbit: I understand. So `orm_error_code::no_tables_specified` is the exception to throw then.

tests/statement_serializer_tests/statements/remove.cpp
trueqbit: OK.
TBH I don't find it hard to read at all. It's even easier on the brain than all the auto and decltype(auto) variables. Developers know it from for-loops or from javascript, and all the template code is so complex that this simple thing doesn't really add any complexity.
Just stating my viewpoint.

dev/statement_serializer.h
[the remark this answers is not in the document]
trueqbit: Yes, they can.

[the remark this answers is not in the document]
fnc12: ok. If we add check whether identifiers contain a double-quote what if it turns out that it does?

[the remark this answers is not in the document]
trueqbit: Instead of `sql_escape()` we'll be using:
```c++
inline void stream_sql_escaped(std::ostream& os, const std::string& str, char char2Escape) {
    for (std::string::const_iterator it = str.cbegin(), next; true; it = next + 1) {
        next = std::find(it, str.cend(), char2Escape);
        os << std::string_view{ it, next };

        if (next == str.cend()) {
            break;
        }
        os << std::string(2, char2Escape);
    }
}
```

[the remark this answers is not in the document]
fnc12: BTW `std::string_view` is not available in C++14. Actually this code has a longer algorithm. I mean it will work longer at runtime. Let's leave it like that. I doubt that we will find a lot of people who uses double quotes in identifiers in their databases. So let's skip over-engineering. Probably we are afraid of a monster which doesn't exist. But of course if someone faces such an issue we'll upgrade the code with a snippet you provided. That do you think?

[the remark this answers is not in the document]
trueqbit: Hmm, why do you worry about runtime overhead? I wouldn't sacrifice accuracy for a few nanoseconds of runtime performance. There are so many other places that chew up runtime.
One could even argue about security - think of sql injection. You are absolutely right about the rareness of such identifiers, but I am not "afraid" of a monster, rather think about accuracy.

So I am not giving up easily :) it's just about finding creative ways...

What about less over-engineered:
```c++
inline void stream_sql_escaped(std::ostream& os, const std::string& str, char char2Escape) {
    for (char c : str) {
        if (c == char2Escape) {
            os << char2Escape;
        }
        os << c;
    }
}
```
Or even better:
```c++
inline void stream_sql_escaped(std::ostream& os, const std::string& str, char char2Escape) {
    if (str.find(char2Escape) == str.npos) [[likely]] {
        os << str;
    } else {
        for (char c : str) {
            if (c == char2Escape) {
                os << char2Escape;
            }
            os << c;
        }
    }
}
```

Otherwise, if you still think of it as a "no", we should prohibit double-quotes in table and column identifiers at storage construction time.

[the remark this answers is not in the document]
fnc12: it looks like ok. I agree with your words. Let's add such a check

fnc12 (approved):

fnc12:
thanks @trueqbit

dev/statement_serializer.h
```diff
@@ -54,18 +54,56 @@ namespace sqlite_orm {
...
+#if __cplusplus >= 202002L  // C++20 or later
+        inline void stream_sql_escaped(std::ostream& os, const std::string& str, char char2Escape) {
+            for(std::string::const_iterator it = str.cbegin(), next; true; it = next + 1) {
+                next = std::find(it, str.cend(), char2Escape);
+                os << std::string_view{it, next};
+
+                if(next == str.cend()) [[likely]] {
+                    break;
+                }
+                os << std::string(2, char2Escape);
```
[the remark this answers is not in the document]
trueqbit: I would prefer a simple array as well, since it saves a bit stack space and the constructor/destructor calls. What do you think?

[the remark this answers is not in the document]
fnc12: yes I forgot about small string optimization. So it doesn't matter array or string. Byte economy is not so important. So it is up to you

dev/statement_serializer.h
```diff
@@ -54,18 +54,56 @@ namespace sqlite_orm {
...
+            }
+            {
+                ss << quoteChar;
+                stream_sql_escaped(ss, identifier, quoteChar);
+                ss << quoteChar;
             }
-            ss << quoteChar << sql_escape(identifier, quoteChar) << quoteChar;
             if(!alias.empty()) {
-                ss << std::string{' ', quoteChar} << sql_escape(alias, quoteChar) << quoteChar;
+                ss << std::string{' ', quoteChar};
```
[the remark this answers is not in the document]
trueqbit: I would prefer a simple array as well, since it saves a bit stack space and the constructor/destructor calls. What do you think?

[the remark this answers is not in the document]
fnc12: same answer from the previous question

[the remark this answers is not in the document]
trueqbit: See [PR #998](https://github.com/fnc12/sqlite_orm/pull/998)

PR #1006, opened 2022-04-15
erased get_table_info() and replaced it with call to pragma.table_info()

juandent:
also did some refactoring in schema_status()
now we don't need to invoke pragma.table_info() from inside schema_status()!

### DIFF
--- a/dev/implementations/storage_impl_definitions.h
+++ b/dev/implementations/storage_impl_definitions.h
@@ -86,36 +86,6 @@ namespace sqlite_orm {
             return notEqual;
         }
 
-        inline std::vector<table_info> storage_impl_base::get_table_info(const std::string& tableName,
-                                                                         sqlite3* db) const {
-            std::vector<table_info> result;
-            auto query = "PRAGMA table_info(" + quote_identifier(tableName) + ")";
-            auto rc = sqlite3_exec(
-                db,
-                query.c_str(),
-                [](void* data, int argc, char** argv, char**) -> int {
-                    auto& res = *(std::vector<table_info>*)data;
-                    if(argc) {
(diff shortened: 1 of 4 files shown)


dev/implementations/storage_impl_definitions.h
```diff
@@ -154,7 +124,8 @@ namespace sqlite_orm {
         }
 
         template<class H, class... Ts>
-        sync_schema_result storage_impl<H, Ts...>::schema_status(sqlite3* db, bool preserve) const {
+        sync_schema_result
+        storage_impl<H, Ts...>::schema_status(sqlite3* db, bool preserve, std::vector<table_info>& dbTableInfo) const {
```
fnc12, line 128: why do you pass non const reference? better make it `const` cause `dbTableInfo` is never modified

fnc12 (changes requested):

dev/implementations/storage_impl_definitions.h
juandent: schema_status() changes parameter dbTableInfo in line 19496, in the call to `if(this->calculate_remove_add_columns(columnsToAdd, storageTableInfo, dbTableInfo)) {`
otherwise I would have passed `const reference`

juandent: please see `dbTableInfo.erase(dbColumnInfoIt);` inside `bool storage_impl_base::calculate_remove_add_columns`

fnc12: oh I see. Sorry. Everything is ok

fnc12 (approved):

fnc12:
thank @juandent

Issue #1007, opened 2022-04-16
Unconditionally undefining the "DELETE" preprocessor macro is bad


firedaemon-cto:
@DoozyX tried to fix build errors on Windows in commit 4a9084447de5 and unconditionally undefines the `DELETE` macro in _[dev/journal_mode.h](https://github.com/fnc12/sqlite_orm/blob/c009207a273a3137fff96f48f2be3d5ffcd75a0b/dev/journal_mode.h#L15)_. This is obviously a bad idea as it makes the inclusion of Windows and sqlite_orm headers order dependent.

Unfortunately, ALL CAPS case is an awkard problem in the Windows world, so a good solution would be to use snake case enum value identifiers in sqlite_orm, which would also adher to the STL style. This of course would be a breaking change, and I guess I am a bit late to the party with this suggestion?

fnc12:
yes it is a problem and I approved this solution assuming that it will make headers order dependent. Caps in `journal_mode` is made in `sqlite_orm` to repeat case from SQLite. Breaking changes is a bad idea. But we can try. Probably somebody will be angry but it is a holy light change 🙂. So why not.

trueqbit:
I completely understand the choice of uppercase enum values. Thinking about changes, turning it to lower-case wouldn't even help because as you mentioned in the enum's comment "delete" is a C/C++ keyword.

PR #1008 solves the include order dependency. To make life easier on Windows, can we introduce a dubber? I.e. extend the `journal_mode` enum class with `DISCARD = DELETE`?

In addition to the other enum values, we could basically provide both lowercase and uppercase variants, just to be on the safe side about any stupid macros. However, this is not required.
What do you think?

fnc12:
>In addition to the other enum values, we could basically provide both lowercase and uppercase variants, just to be on the safe side about any stupid macros. However, this is not required.

fnc12:
> In addition to the other enum values, we could basically provide both lowercase and uppercase variants, just to be on the safe side about any stupid macros. However, this is not required.

Sounds like over-engineering.

> PR https://github.com/fnc12/sqlite_orm/pull/1008 solves the include order dependency. To make life easier on Windows, can we introduce a dubber? I.e. extend the journal_mode enum class with DISCARD = DELETE?

Why `DISCARD`?

trueqbit:
[...]

PR #1008, opened 2022-04-16
Corrected undefining DELETE macro

trueqbit:
Addresses #1007.

Also I simplified pushing/undefining/popping macro definitions (Visual C++ doesn't complain if macro is undefined).


fnc12:
thanks @trueqbit

PR #1012, opened 2022-04-17
moved schema_status from storage_impl to storage_t and removed unused…

fnc12:
I am moving `schema_status` from `storage_impl` to `storage_t` cause `storage_impl` must not have member functions cause it must be refactored to `std::tuple` of `table_t`s


juandent (approved):
the original code is not satisfactory because:

`class storage_t` defines:       (==> means invokes):
 
`schema_status()` ==> `storage_impl::schema_status()`
`sync_table()` ==> `storage_impl::schema_status()`
`sync_schema_simulate()` ==> `storage_t ::schema_status` ==> `storage_impl::schema_status()`

`storage_impl::schema_status` requires the result of ` pragma.table_info(tablename)` so it is passed by the caller by parameter. Moving around state from object to object is a sign of low cohesion and strong coupling, in this case between `storage_t` and  `storage_impl<>`

moving `schema_status()` into `storage_t` and out of `storage_impl<>` not only increases cohesion, decreases coupling but also simplifies both types involved

juandent (approved):
the new design is simplified. `schema_status` is used twice from `witihin sync_table()` and `sync_schema_simulate()`, there is no more forwarding...
and there are only 2 definitions for `schema_status`...

trueqbit (approved):
Approving it as well.

PR #1014, opened 2022-04-18
Streamlined field value access

trueqbit:
It is advantageous to fold the object member pointer and member pointer function, which are stored in `column_t`, because it allows access to field values in a unified way and thus allows for streamlining and generalization of access to field values.

* Saves 2 pointers per column when using object member pointers
* Saves 1 pointer per column when using member pointer functions
* Accessing a data object's 'field value' was extremely simplified, as C++17's `std::invoke()` can usually be used instead
* Simplified comparing member pointers and finding columns


dev/statement_serializer.h
```diff
@@ -1298,11 +1301,11 @@ namespace sqlite_orm {
                         if(columnIndex > 0) {
                             ss << ", ";
                         }
-                        if(column.member_pointer) {
-                            ss << serialize(object.*column.member_pointer, context);
-                        } else {
-                            ss << serialize((object.*column.getter)(), context);
-                        }
+#if __cplusplus >= 201703L  // C++17 or later
+                        ss << serialize(std::invoke(column.member_pointer, object), context);
```
fnc12, line 1305: this API is very nice. I have one idea about reducing amount of `#ifdef`s. We can create `invoke` function which will call `std::invoke` with C++17 and higher and `access_field_value` with C++14. Then we can write such statements using one line not five. What do you think?

trueqbit: You are absolutely right about the ugly `#ifdef`s, I was expecting your feedback :)
In this sense it's best if we polyfill `std::invoke`? It's doable with either overloads or reusing `static_if`. The polyfill only needs to be good enough for internal purposes.

fnc12: I was thinking about an overload of `invoke` inside `sqlite_orm` namespace with one `#ifdef` inside but you can use polyfill also

dev/member_traits/is_field_member_pointer.h
```diff
@@ -1,17 +1,11 @@
...
 
-        template<class T, class SFINAE = void>
-        struct is_field_member_pointer : std::false_type {};
-
         template<class T>
-        struct is_field_member_pointer<T,
-                                       typename std::enable_if<std::is_member_pointer<T>::value &&
-                                                               !std::is_member_function_pointer<T>::value>::type>
-            : std::true_type {};
+        using is_field_member_pointer = std::is_member_object_pointer<T>;
```
trueqbit, line 9: @fnc12 Do we still want to keep `is_field_member_pointer<>`? It's only an alias.

fnc12: no

fnc12:
thanks @trueqbit

PR #1016, opened 2022-04-19
Providing NULLIF core function

trueqbit:
Additional versions of `nullif(), ifnull(), coalesce()` with automatic return type deduction for simple cases.


fnc12:
thanks @trueqbit

Issue #166, opened 2018-08-02
tests/test.cpp does not build in current master


unclevlad:
We were looking for a good example of the use of 'and' in a where clause, and tried building the latest test.cpp, which fails.

1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(920): error C2146: syntax error: missing ')' before identifier 'and'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(920): error C2146: syntax error: missing ';' before identifier 'and'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(919): error C2065: 'and': undeclared identifier
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(920): error C2146: syntax error: missing ';' before identifier 'greater_than'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(921): error C2146: syntax error: missing ';' before identifier 'and'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(920): error C2065: 'and': undeclared identifier
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(921): error C2146: syntax error: missing ';' before identifier 'lesser_than'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(922): error C2146: syntax error: missing ';' before identifier 'and'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(921): error C2065: 'and': undeclared identifier
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(922): error C2146: syntax error: missing ';' before identifier 'lesser_than'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(922): error C2059: syntax error: ')'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(923): error C2059: syntax error: ')'
1>c:\users\user\development\cppcommon\sqlite_orm\test\test.cpp(1164): warning C4018: '<': signed/unsigned mi

fnc12:
what platform? windows?

fnc12:
@unclevlad bump

unclevlad:
Sorry for the delay.

Windows / Visual Studio 2017 / Latest Update / 32- and 64-bit

(To be honest, we are having problems getting any of the examples with where clauses to build and/or work)

fnc12:
[...]

PR #1017, opened 2022-04-19
Updated the way feature macros are determined

trueqbit:
* `<ciso646>` was officially removed in C++20, hence include the C-header equivalent `<iso646.h>` in order to pull in alternative operator representations (the latter needs to be kept for non-conforming compilers, also see [#166](https://github.com/fnc12/sqlite_orm/issues/166))
* Rely on C++ feature test macros to determine availability of features instead of C++ standard version


fnc12:
thanks @trueqbit

PR #1018, opened 2022-04-20
Slim down `storage_impl<>`

trueqbit:
Removed last remaining storage_impl methods.

* Reusing already globally available `lookup_table_name()` instead of `storage_impl<>::find_table_name()`
* Using global `find_table_name(std::type_index)` where still necessary
* Moved `storage_impl<>::copy_table()` to `storage_t<>`

Additionally:
* Streamlined tuple filtering


trueqbit (commented):
@fnc12 I guess [`find_in_tuple`](https://github.com/fnc12/sqlite_orm/blob/5c1983acda5b59edb1344f50263df9d991ea384d/dev/tuple_helper/find_in_tuple.h#L10) can be thrown out as well? It's never used within sqlite_orm. There's only a [unit test file](https://github.com/fnc12/sqlite_orm/blob/dev/tests/static_tests/find_in_tuple.cpp) of the same name, which is however not included in CMakeLists.txt.

fnc12:
> @fnc12 I guess [`find_in_tuple`](https://github.com/fnc12/sqlite_orm/blob/5c1983acda5b59edb1344f50263df9d991ea384d/dev/tuple_helper/find_in_tuple.h#L10) can be thrown out as well? It's never used within sqlite_orm. There's only a [unit test file](https://github.com/fnc12/sqlite_orm/blob/dev/tests/static_tests/find_in_tuple.cpp) of the same name, which is however not included in CMakeLists.txt.

yes

fnc12:
Снимок экрана 2022-04-21 в 10 42 25
doesn't compile on my mac

fnc12:
thanks @trueqbit

PR #1019, opened 2022-04-20
Sync_schema() will not lose data anymore

juandent:
Simple migration API encapsulates storage_t::backup_table() so as to guarantee no loss of data upon schema changes


dev/implementations/storage_definitions.h
```diff
@@ -78,8 +78,9 @@ namespace sqlite_orm {
                             res = decltype(res)::new_columns_added_and_old_columns_removed;
                         }
                     } else if(schema_stat == sync_schema_result::dropped_and_recreated) {
-                        this->drop_table_internal(tImpl.table.name, db);
-                        this->create_table(db, tImpl.table.name, tImpl);
+                        this->backup_table(db, tImpl, {});
```
fnc12, line 81: this will break what `sync_schema` is designed for: schema synchronization. `sync_schema` must synchronize schema first, if data cannot be saved it will be deleted instead of operation abort. If we will abort schema synchronization in cases when data cannot be stored then user will be unable to sync schema in some circumstances. That is not right

dev/storage.h
```diff
@@ -149,14 +149,32 @@ namespace sqlite_orm {
...
+                    this->drop_table_internal(tableImpl.table.name, db);
 
-                tableImpl.rename_table(db, backupTableName, tableImpl.table.name);
+                    tableImpl.rename_table(db, backupTableName, tableImpl.table.name);
+
+                    this->commit_migration();
+
+                } catch(std::exception& ex) {
+                    std::string s = ex.what();  // 's' for debugging
+                    this->abort_migration();
```
fnc12, line 167: catching exceptions inside storage functions instead of allowing catching them by users is a bad idea

dev/storage.h
```diff
@@ -149,14 +149,32 @@ namespace sqlite_orm {
...
+                    throw;
+                }
+
+                // this->create_table(db, backupTableName, tableImpl);
+                //
+                // tableImpl.copy_table(db, backupTableName, columnsToIgnore);
+                //
+                // this->drop_table_internal(tableImpl.table.name, db);
+                //
+                // tableImpl.rename_table(db, backupTableName, tableImpl.table.name);
```
fnc12, line 177: BTW commented out code is prohibited

dev/storage_base.h
```diff
@@ -728,6 +728,20 @@ namespace sqlite_orm {
                 }
             }
 
+            // migration internal API
+            void start_migration() {
+                this->pragma.foreign_keys(false);
```
fnc12, line 733: there is no need to disable FK. We've told about already. You can migrate data using raw SQLite without disabling FK then it must work the same way in `sqlite_orm`

fnc12 (changes requested):

fnc12:
@juandent unit tests are failed. No PR can be merged with failing unit tests

dev/storage.h
trueqbit: It's rethrown properly just a line below.

dev/storage_base.h
juandent: Table Emp is the target of Department and thus we cannot drop the Emp with FK on

dev/storage.h
juandent: ok sorry about that

dev/storage.h
juandent: correct: it is rethown... the reason we have that code inside try/block is that iif the migration fails (as when adding a new column with no default value or not nullablem, then the add_column fails and we need to turn foeign keys ON again!!

fnc12: ok no problem

dev/storage_base.h
fnc12: don't drop it, rename it and create a new one

juandent: > there is no need to disable FK. 
look at the schema:

```C++
make_table("Dept",
                       make_column("deptno", &Department::m_deptno, primary_key(), autoincrement()),
                       make_column("deptname", &Department::m_deptname),
                       make_column("loc", &Department::m_loc),
                       make_column("mgr", &Department::m_manager),
                       foreign_key(&Department::m_manager).references(&Employee::m_empno)),
```

with this FK we cannot drop Employee table with FK ON!!
Why do you insist that this is possible?

juandent: please pay attention to my ideas - I am going to address your objections after my swimming 
please have an open mind
the goal with sync_schema is to not lose data while syncronizing , it should always be called inside try/block so user can see what existing data is not allowing the synchronization and correct it!

fnc12: > Why do you insist that this is possible?

Cause it is possible.

> with this FK we cannot drop Employee table with FK ON!!

don't drop it, rename it and create a new one.

dev/implementations/storage_definitions.h
fnc12: anyway you need to run `sync_schema` from within try/block cause it can throw just like any SQLite call. But it doesn't change my words

dev/storage_base.h
fnc12: > the goal with sync_schema is to not lose data while synchronizing 

why do you think so? One of scenarios of `sync_schema` is losing data, it is noted in a comment. `sync_schema` means `please synchronize schema in db just like I wrote in C++ inside make_storage call`. If it will not synchronize by artificial reasons then it means that function doesn't work. If someone is afraid to lose data there is an easy way to look into the future and to know what will happen if he/she will call `sync_schema` - to call `sync_schema_simulate` and check the result. What you are doing in this PR can already be achieved with different API. So please don't affect current `sync_schema` success results (BTW current unit tests failure tell about it). If you want to make life better for people who loses data during `sync_schema` calls then you need to
1) keep current `sync_schema` success results, skip changing algo to 'oops data is lost so no sync, here is an exception for even if you don't care about data';
2) add some API from side which will allow to handle cases when data is removed. You can place it under `if (preserve)` flag inside `sync_schema` cause `preserve` argument is designed for your case especially or make a new function

juandent: > > Why do you insist that this is possible?
> 
> Cause it is possible.
> 
> > with this FK we cannot drop Employee table with FK ON!!
> 
> don't drop it, rename it and create a new one.

I thought renaming was also prohibited if FK was ON!!
Now I know different! Will try to use this in backup_table

fnc12: the best way of adding functionality to `sqlite_orm` is to make it using raw SQLite first

dev/pragma.h
```diff
@@ -117,17 +117,18 @@ namespace sqlite_orm {
...
 
-            std::vector<sqlite_orm::table_info> table_info(const std::string& tableName) const {
+            // will include generated columns in response as opposed to table_info
+            std::vector<sqlite_orm::table_xinfo> table_xinfo(const std::string& tableName) const {
                 auto connection = this->get_connection();
                 auto db = connection.get();
 
-                std::vector<sqlite_orm::table_info> result;
-                auto query = "PRAGMA table_info(" + quote_identifier(tableName) + ")";
+                std::vector<sqlite_orm::table_xinfo> result;
```
trueqbit, line 125: Namespace qualifier can be omitted.

dev/pragma.h
```diff
@@ -117,17 +117,18 @@ namespace sqlite_orm {
                 return this->get_pragma<std::vector<std::string>>(oss.str());
             }
 
-            std::vector<sqlite_orm::table_info> table_info(const std::string& tableName) const {
+            // will include generated columns in response as opposed to table_info
+            std::vector<sqlite_orm::table_xinfo> table_xinfo(const std::string& tableName) const {
```
trueqbit, line 121: Namespace qualifier can be omitted.

dev/implementations/storage_definitions.h
```diff
@@ -73,14 +74,31 @@ namespace sqlite_orm {
 
                         if(schema_stat == sync_schema_result::new_columns_added_and_old_columns_removed) {
 
-                            // remove extra columns
+                            auto storageTableInfo = tImpl.table.get_table_info();
+                            add_generated_cols(columnsToAdd, storageTableInfo);  //
```
trueqbit, line 78: empty comment.

dev/storage.h
```diff
@@ -142,33 +142,82 @@ namespace sqlite_orm {
...
+            void drop_create_with_loss(const I& tImpl, sqlite3* db) {
+
+                try {
+                    this->start_migration();
+
+                    this->drop_table_internal(tImpl.table.name, db);
+                    this->create_table(db, tImpl.table.name, tImpl);
+
+                    this->commit_migration();
+                } catch(std::exception& ex) {
```
trueqbit, line 169: Please catch by const-reference, and e.g. comment out the variable name.

dev/storage.h
```diff
@@ -142,33 +142,82 @@ namespace sqlite_orm {
...
+                    this->drop_table_internal(tableImpl.table.name, db);
 
-                this->copy_table(db, backupTableName, tableImpl, columnsToIgnore);
+                    this->rename_table(db, backupTableName, tableImpl.table.name);
 
-                this->drop_table_internal(tableImpl.table.name, db);
+                    this->commit_migration();
 
-                tableImpl.rename_table(db, backupTableName, tableImpl.table.name);
+                } catch(std::exception& ex) {
```
trueqbit, line 208: Please catch by const-reference, and e.g. comment out the variable name.

dev/table_info.h
```diff
@@ -22,4 +22,24 @@ namespace sqlite_orm {
             name(move(name_)), type(move(type_)), notnull(notnull_), dflt_value(move(dflt_value_)), pk(pk_) {}
     };
 
+    struct table_xinfo {
```
trueqbit, line 25: Why don't you simply extend `table_info` above?

tests/sync_schema_tests.cpp
```diff
@@ -507,6 +507,7 @@ TEST_CASE("sync_schema with generated columns") {
         }
         SECTION("stored") {
             generatedAlwaysConstraint = generatedAlwaysConstraint.stored();
+            expectedUsers.push_back({5, 9});  // jdh added
```
trueqbit, line 510: Please never include any comments about who added a line.

trueqbit (changes requested):

dev/pragma.h
juandent: not really, colides with method name table_xinfo

dev/pragma.h
juandent: sorry, what do you mean?

dev/storage.h
```diff
@@ -142,33 +142,82 @@ namespace sqlite_orm {
                 perform_void_exec(db, ss.str());
             }
 #endif
+            inline void add_generated_cols(std::vector<table_xinfo*>& columnsToAdd,
```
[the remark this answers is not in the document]
juandent: ok removed

dev/storage.h
```diff
@@ -142,33 +142,82 @@ namespace sqlite_orm {
...
+
+                try {
+                    this->start_migration();
+
+                    this->drop_table_internal(tImpl.table.name, db);
+                    this->create_table(db, tImpl.table.name, tImpl);
+
+                    this->commit_migration();
+                } catch(std::exception& ex) {
+                    std::string s = ex.what();  // 's' for debugging
```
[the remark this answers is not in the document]
juandent: ok

dev/storage.h
```diff
@@ -142,33 +142,82 @@ namespace sqlite_orm {
...
 
-                this->copy_table(db, backupTableName, tableImpl, columnsToIgnore);
+                    this->rename_table(db, backupTableName, tableImpl.table.name);
 
-                this->drop_table_internal(tableImpl.table.name, db);
+                    this->commit_migration();
 
-                tableImpl.rename_table(db, backupTableName, tableImpl.table.name);
+                } catch(std::exception& ex) {
+                    std::string s = ex.what();  // 's' for debugging
```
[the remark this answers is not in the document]
juandent: ok

tests/statement_serializer_tests/schema/index.cpp
```diff
@@ -52,3 +52,22 @@ TEST_CASE("statement_serializer index") {
...
+    };
+    auto storage = make_storage("",
+                                make_index("user_name", &User::name),
+                                make_table("users",
+                                           make_column("id", &User::id, primary_key(), autoincrement()),
+                                           make_column("name", &User::name)));
+    storage.sync_schema(true);
+    storage.sync_schema_simulate(true);  // this will call in storage_t:
+
+    // template<class... Tss, class... Cols>
```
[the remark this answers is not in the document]
juandent: I removed the commented lines

tests/statement_serializer_tests/schema/index.cpp
```diff
@@ -52,3 +52,22 @@ TEST_CASE("statement_serializer index") {
...
+    // sync_schema_result schema_status(const storage_impl<index_t<Cols...>, Tss...>&, sqlite3*, bool) {
+    //     return sync_schema_result::already_in_sync;
+    // }
+}
```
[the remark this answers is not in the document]
juandent: ?? don't understand

tests/sync_schema_tests.cpp
```diff
@@ -529,7 +530,7 @@ TEST_CASE("sync_schema with generated columns") {
         }
         SECTION("stored") {
             generatedAlwaysConstraint = generatedAlwaysConstraint.stored();
-            //  expectedUsers.push_back({5, 9});    //  TODO: add sync_schema(true) support
+            expectedUsers.push_back({5, 9});  // JDH added
```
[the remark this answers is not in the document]
juandent: I removed my name

dev/implementations/storage_definitions.h
juandent: removed

dev/storage.h
juandent: ok done

dev/storage.h
juandent: done

dev/table_info.h
juandent: did not know if we were going to delete table_info now that it is obsolete!

tests/sync_schema_tests.cpp
juandent: ok removed

dev/implementations/storage_definitions.h
juandent: > if data cannot be saved it will be deleted instead of operation abort
the operation abort will give the programmer a chance to see what is preventing the migration and take action so it can

dev/table_info.h
trueqbit: That doesn't matter. You just implement the way it should be. There's no point in splitting off an entire struct/class, where the old one is obsoleted completely anyway.

dev/storage.h
```diff
@@ -142,33 +142,72 @@ namespace sqlite_orm {
                 perform_void_exec(db, ss.str());
             }
 #endif
+            void add_generated_cols(std::vector<table_xinfo*>& columnsToAdd,
+                                    std::vector<table_xinfo>& storageTableInfo) {
```
fnc12, line 146: if `storageTableInfo` is not modified within `add_generated_cols` please pass it as const ref instead

dev/storage.h
```diff
@@ -826,6 +865,41 @@ namespace sqlite_orm {
             }
 
           protected:
+            bool calculate_remove_add_columns(std::vector<table_xinfo*>& columnsToAdd,
+                                              std::vector<table_xinfo>& storageTableInfo,
+                                              std::vector<table_xinfo>& dbTableInfo);
+            inline void rename_table(sqlite3* db, const std::string& oldName, const std::string& newName) const {
```
fnc12, line 871: no need to `inline` member functions. Please remove `inline` - the code will be shorter with the same functionality

dev/storage.h
```diff
@@ -835,20 +909,20 @@ namespace sqlite_orm {
             sync_schema_result schema_status(const storage_impl<table_t<T, WithoutRowId, Cs...>, Tss...>& tImpl,
                                              sqlite3* db,
                                              bool preserve) {
-                auto dbTableInfo = this->pragma.table_info(tImpl.table.name);
+                auto dbTableInfo = this->pragma.table_xinfo(tImpl.table.name);  // should include generated
```
fnc12, line 912: such comments better place as comments inside PR not inside the code cause they are very specific

dev/storage.h
```diff
@@ -884,6 +958,7 @@ namespace sqlite_orm {
                                 } else {
                                     if(columnPointer->notnull && columnPointer->dflt_value.empty()) {
                                         gottaCreateTable = true;
+                                        res = decltype(res)::table_data_loss;
```
fnc12, line 961: I am still not sure that adding `table_data_loss` is required. `table_data_loss` is equal to dropped and recreated when `preserve = false` so it is redundant

dev/sync_schema_result.h
```diff
@@ -39,6 +39,11 @@ namespace sqlite_orm {
          *      4. data_type mismatch between table and storage.
          */
         dropped_and_recreated,
+
+        /*
+         * old table is dropped and new recreated but no effort at keeping data is made
```
fnc12, line 44: not sure that sync schema result my have a dedicated value about data status. Data is lost if result is dropped and recreated with `preserve = false`

fnc12 (changes requested):

fnc12:
@juandent this PR has very useful `storage_impl_base` change which I'd merge right now but it has a controversial change of `sync_schema` and `sync_schema_result`. Let's make a dedicated PR with `storage_impl_base` change and merge it separately to reduce diff and keep it thin

dev/pragma.h
```diff
@@ -116,7 +116,7 @@ namespace sqlite_orm {
                 oss << "integrity_check(" << n << ")";
                 return this->get_pragma<std::vector<std::string>>(oss.str());
             }
-
+#if 0  // JDH
             std::vector<sqlite_orm::table_info> table_info(const std::string& tableName) const {
```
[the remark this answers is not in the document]
juandent: re inserted table_info

dev/storage_impl.h
```diff
@@ -24,9 +24,13 @@ namespace sqlite_orm {
...
 
-            static bool calculate_remove_add_columns(std::vector<table_info*>& columnsToAdd,
-                                                     std::vector<table_info>& storageTableInfo,
-                                                     std::vector<table_info>& dbTableInfo);
+            static bool calculate_remove_add_columns(std::vector<table_xinfo*>& columnsToAdd,
+                                                     std::vector<table_xinfo>& storageTableInfo,
+                                                     std::vector<table_xinfo>& dbTableInfo);
+
+            static void
+            add_generated_cols(std::vector<table_xinfo*>& columnsToAdd,
```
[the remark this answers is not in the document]
juandent: you are not looking at latest changes!
currently storage_impl_base is empty

dev/storage.h
juandent: done

dev/storage.h
juandent: done

dev/storage.h
juandent: where in PR? I removed it

dev/storage.h
juandent: you don't understand my implementation.. table_data_loss is NOT equal to dropped and recreated!! Look at the code please
has nothing to do with preserve == false

juandent: in this special case:

```C++
if(columnPointer->notnull && columnPointer->dflt_value.empty()) {
                                        gottaCreateTable = true;
                                        res = decltype(res)::table_data_loss;
                                        break;
                                    }
```

the code to call WILL LOSE DATA by dropping and recreating without any backup, look at sync_table() in storage_definitions.h

```C++
                    } else if(schema_stat == sync_schema_result::table_data_loss) {
                        this->drop_create_with_loss(tImpl, db);
                        res = decltype(res)::table_data_loss;
                    }
```
this is very different from how drop and recreated is performed:

```C++
                    } else if(schema_stat == sync_schema_result::dropped_and_recreated) {
                        //  now get current table info from db using `PRAGMA table_xinfo` query..
                        auto dbTableInfo =
                            this->pragma.table_xinfo(tImpl.table.name);  // should include generated columns
                        auto storageTableInfo = tImpl.table.get_table_info();

                        //  this vector will contain pointers to columns that gotta be added..
                        std::vector<table_xinfo*> columnsToAdd;

                        calculate_remove_add_columns(columnsToAdd, storageTableInfo, dbTableInfo);

                        add_generated_cols(columnsToAdd, storageTableInfo);

                        this->backup_table(db, tImpl, columnsToAdd);
                        res = decltype(res)::dropped_and_recreated;
```
See?

dev/table_info.h
juandent: but @fnc12 wants table_info as part of the public API!

tests/statement_serializer_tests/schema/index.cpp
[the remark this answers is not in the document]
juandent: ok got it!

dev/sync_schema_result.h
juandent: no, if we wanted that then we could se the new enum value to express that! But I don't see the point.. however I can implement that if you insist
just let my migration API as it is because it works!!

tests/statement_serializer_tests/schema/index.cpp
```diff
@@ -52,3 +52,22 @@ TEST_CASE("statement_serializer index") {
     }
     REQUIRE(value == expected);
 }
+
+TEST_CASE("trying schema_status index") {
```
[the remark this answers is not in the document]
juandent: will remove it!!

tests/sync_schema_tests.cpp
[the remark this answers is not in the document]
juandent: done!

tests/sync_schema_tests.cpp
```diff
@@ -529,7 +530,7 @@ TEST_CASE("sync_schema with generated columns") {
...
         }
         auto storage2 = make_storage(storagePath,
                                      make_table("users",
```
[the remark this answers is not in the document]
juandent: generated_columns don't fail anymore! that was taken care in part by the use of table_xinfo!

dev/implementations/storage_definitions.h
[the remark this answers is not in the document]
juandent: done

dev/implementations/storage_definitions.h
```diff
@@ -126,5 +145,44 @@ namespace sqlite_orm {
...
+                    return ti.name == columnName;
+                });
+                if(dbColumnInfoIt != dbTableInfo.end()) {
+                    auto& dbColumnInfo = *dbColumnInfoIt;
+                    auto columnsAreEqual =
+                        dbColumnInfo.name == storageColumnInfo.name &&
+                        dbColumnInfo.notnull == storageColumnInfo.notnull &&
+                        (!dbColumnInfo.dflt_value.empty()) == (!storageColumnInfo.dflt_value.empty()) &&
+                        dbColumnInfo.pk == storageColumnInfo.pk &&
+                        (dbColumnInfo.hidden == 0) == (storageColumnInfo.hidden == 0);  // added
```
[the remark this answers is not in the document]
juandent: done!

dev/pragma.h
```diff
@@ -152,6 +154,13 @@ namespace sqlite_orm {
           private:
             friend struct storage_base;
 
+            bool foreign_keys() {
+                return this->get_pragma<bool>("foreign_keys");
```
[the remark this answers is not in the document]
juandent: yes! but only to be called inside functions inside sync_schema() call... that's why it is private - we don't want to expose the method!!

dev/pragma.h
```diff
@@ -152,6 +186,24 @@ namespace sqlite_orm {
...
+            }
+            void foreign_keys(bool value) {
+                /*
+                 * this define makes it impossible to change the value of FK checking if undefined!!
+                 */
+#ifdef FK_TOGGLE_ENABLE
```
fnc12, line 196: this is not acceptable cause if `FK_TOGGLE_ENABLE` is going to be used only by you then leave it in your local branch. No need to commit things which are never used. We don't pay for what we don't use.

dev/storage.h
```diff
@@ -868,7 +952,11 @@ namespace sqlite_orm {
                         }
                     }
                     if(gottaCreateTable) {
-                        res = decltype(res)::dropped_and_recreated;
+                        if(preserve) {
+                            res = decltype(res)::dropped_and_recreated;
+                        } else {
+                            res = decltype(res)::table_data_loss;
```
fnc12, line 958: as I said `table_data_loss` is returned when `dropped_and_recreated` returns in `dev` with `preserve = false`. Also `table_data_loss` sounds like a data state not schema sync result so I don't understand why you added it to `sync_schema_result`. If user want to know whether data was lost he/she can do it right now in `dev` but checking or `dropped_and_recreated` and `preserve == false`

dev/storage.h
fnc12: > has nothing to do with preserve == false
it has
Снимок экрана 2022-04-26 в 10 19 54

dev/storage_base.h
```diff
@@ -450,6 +451,18 @@ namespace sqlite_orm {
                 return this->connection->retain_count() > 0;
             }
 
+            /*
+             * checks whether there is a transaction in place
+             */
+            bool in_transaction() const {
```
fnc12, line 457: better call this function `get_autocommit` to make it recognizable by people who know `sqlite3_get_autocommit` API. This is how all other API designed in `storage_t`. Also use `get_connection` instead to obtain a connection for sure

dev/storage_base.h
```diff
@@ -728,6 +741,30 @@ namespace sqlite_orm {
...
+            }
+            void commit_migration() {
+                auto foreign_keys_guard = finally([&pragma = this->pragma]() {
+                    pragma.foreign_keys(true);
+                });
+                this->commit();
+            }
+            void abort_migration() {
+                auto foreign_keys_guard = finally([&pragma = this->pragma]() {
+                    pragma.foreign_keys(true);
```
fnc12, line 763: what if `pragma.foreign_keys(true);` returns non `SQLITE_OK` how user will get info about that?

tests/sync_schema_tests.cpp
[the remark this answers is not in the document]
fnc12: yes that it why we need to add a unit test for that - to make sure it will never fail in future

fnc12 (changes requested):

dev/pragma.h
juandent: I would like to keep this minimum support to FK toggling as part of the repo because we can document it as the configuration that tries to preserve data by making a backup when possible
It is not only for me, it is a configurable option

dev/storage.h
juandent: nit picking, the semantics of table_data_loss is a schema sync result just like dropped_and_recreated. It triggers a different treatment... see backup_table() and drop_create_with_loss() ... the latter doesn't backup and restore only drops/recreates if FK allows (rememeber FK is on). 
Actually because FK is on, then dropped and recreated only works when no  FK constraints are violated

dev/pragma.h
juandent: you might learn to be less selfish and reach a compromise with your collaborators who work very hard to make a better product...a very few lines makes the product configurable and would satisfy the people who want to save data when possible.

juandent: collaborating is making compromises, accepting others' ideas and letting those ideas enrich the product.

dev/storage_base.h
juandent: no problem with this!

dev/storage_base.h
juandent: this is a problem with set_pragma which is called by all the setting pragmas; thus I suggest this solution

```C++
            template<class T>
            void set_pragma(const std::string& name, const T& value, sqlite3* db = nullptr) {
                auto con = this->get_connection();
                if(!db) {
                    db = con.get();
                }
                std::stringstream ss;
                ss << "PRAGMA " << name << " = " << value;
#if 1   // JDH
                T result;
                auto rc = sqlite3_exec(db, ss.str().c_str(), nullptr, &result, nullptr);
                if (rc != SQLITE_OK) {
                    throw_translated_sqlite_error(db);
                }
#else
                internal::perform_void_exec(db, ss.str());
#endif
            }

            void set_pragma(const std::string& name, const sqlite_orm::journal_mode& value, sqlite3* db = nullptr) {
                auto con = this->get_connection();
                if(!db) {
                    db = con.get();
                }
                std::stringstream ss;
                ss << "PRAGMA " << name << " = " << internal::to_string(value);
#if 1   // JDH
                auto rc = sqlite3_exec(db, ss.str().c_str(), nullptr, nullptr, nullptr);
                if (rc != SQLITE_OK) {
                    throw_translated_sqlite_error(db);
                }
#else
                internal::perform_void_exec(db, ss.str());
#endif
            }
        };
```

Do you agree?

juandent: Actually perform_void_exec() checks return from sqlite3_exec as follows:

```C++
        inline void perform_void_exec(sqlite3* db, const std::string& query) {
            int rc = sqlite3_exec(db, query.c_str(), nullptr, nullptr, nullptr);
            if(rc != SQLITE_OK) {
                throw_translated_sqlite_error(db);
            }
        }
```

so nothing has to be changed!

dev/pragma.h
fnc12: I do reach compromise often enough. If I don't accept your feature it means that I don't accept your feature not I am selfish/compromise hater etc. Please look at the amount of PRs created by other people for the whole history of the lib. Most of opened PRs from other people were merged successfully. Please consider this before calling someone selfish

dev/storage.h
fnc12: I see the difference but table_data_loss has nothing common with schema. Schema is a list of objects of four types: tables, views, indexes and triggers. Please note that schema doesn't contain rows. No rows = no data. Every existing enum value of `sync_schema_result` a unique operation performed with schema. If you compare all enum value with each other you'll notice that there are no two equal enum values. But if you compare `table_data_loss` with `dropped_and_recreated` you'll find that `table_data_loss` is actually `dropped_and_recreated` in regards of schema. This is why `table_data_loss` is not a sync schema result. So it has nothing to do in `sync_schema_result` declaration. If someone need to know whether data (that is, rows) was dropped also he/she can check if result is `dropped_and_recreated` and `preserve` is `false`.

fnc12: if you think that I am nit picking you can make your custom fork where you can do whatever you want. Currently `sqlite_orm` has 200+ forks where people can commit what they want without `nit picking`. Actually you are the first who said that. Sometimes when you think that the world is wrong and you are right probably everything is reversed: the world is ok but you do something wrong

dev/storage_base.h
fnc12: alright

dev/storage.h
juandent: what if I eliminate that value from the enum? Will you be finally at peace with it?

dev/final_action.h
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+  private:
+    F f_;
+    bool invoke_ = true;
+};
+
+template<class F>
+auto finally(F&& f) noexcept {
```
fnc12, line 29: you use `finally` with lambda which throws. In such cases `std::terminate` gets called instead of allowing user to catch an exception. Library which may terminate the whole app by design is not what people want to use. Such code is not going to be merged into `sqlite_orm` ever.

fnc12 (changes requested):

dev/storage.h
fnc12: if you eliminate `table_data_loss` then the issue with `table_data_loss` will stop existing

dev/final_action.h
juandent: I removed this

dev/storage_base.h
```diff
@@ -25,6 +25,7 @@
 #include "function.h"
 #include "values_to_tuple.h"
 #include "arg_values.h"
+// #include "final_action.h"
```
fnc12, line 28: please don't commit commented out code. Just remove it. If we need to restore some code we can use git for it

dev/storage_base.h
fnc12: please rename it to `get_autocommit` or remove it if it is not used

dev/storage_base.h
```diff
@@ -450,6 +451,18 @@ namespace sqlite_orm {
                 return this->connection->retain_count() > 0;
             }
 
+            /*
+             * checks whether there is a transaction in place
+             */
+            bool in_transaction() const {
+                auto&& conn = this->connection;
```
fnc12, line 458: what is the purpose of right reference here?

dev/storage_base.h
```diff
@@ -728,6 +741,17 @@ namespace sqlite_orm {
...
 
+            // migration internal API
+            void start_migration() {
+                this->begin_exclusive_transaction();
+            }
+            void commit_migration() {
+                this->commit();
+            }
+            void abort_migration() {
+                this->rollback();
```
fnc12, line 752: just a question: why don't you call `rollback` instead of `abort_migration`? These functions are not migrations, these functions are aliases for transactions actually

tests/sync_schema_tests.cpp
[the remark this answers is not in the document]
fnc12: what unit test can repro https://github.com/fnc12/sqlite_orm/issues/1020?

fnc12 (changes requested):

fnc12:
@juandent you did a very important change in `sync_schema` - now `sync_schema` can call a transaction. It is a bad idea cause `sync_schema` is designed to avoid transaction calls to give user opportunity to call `sync_schema` within a transaction by himself. If we merge this code some users can have `sync_schema` not working anymore cause it will fire exceptions saying transaction was tried to be called inside a transaction

juandent:
what if we test to see if we are in a transaction at the entry to backup_table, and if it is in a transaction we do not start another... I understand what you say but it is necessary to make sure we cleanup -- well that could be accomplished with just the try catch
what about the tables that were created ... they will soon fill the database with backup tables that weren't removed by calliing rollback!

Have you thought about this?

dev/implementations/storage_definitions.h
```diff
@@ -126,5 +147,84 @@ namespace sqlite_orm {
             perform_void_exec(db, ss.str());
         }
 
+        template<class... Ts>
+        template<class I>
+        void storage_t<Ts...>::copy_table_from(
```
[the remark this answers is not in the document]
juandent: Work done!!

dev/storage.h
```diff
@@ -826,6 +875,41 @@ namespace sqlite_orm {
             }
 
           protected:
+            bool calculate_remove_add_columns(std::vector<const table_xinfo*>& columnsToAdd,
+                                              std::vector<table_xinfo>& storageTableInfo,
+                                              std::vector<table_xinfo>& dbTableInfo);
```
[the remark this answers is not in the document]
juandent: what do you mean by bump?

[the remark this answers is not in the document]
juandent: the function is now const!

dev/storage_base.h
juandent: ok, done!

dev/storage_base.h
juandent: well now they have a little more substance so I think we should keep them.. please check them out and look at backup_table()'s implementation...

juandent:
> what unit test can repro https://github.com/fnc12/sqlite_orm/issues/1020?
This is no longer an issue .. it has been fixed!!]

dev/storage.h
juandent: table_data_loss was removed! FK toggling option too!

dev/storage_base.h
juandent: Done!

dev/storage_base.h
juandent: auto&& because connection was a unique_ptr<> and I did not want to move from it

juandent: but i removed it
now I call get_connection()

dev/storage.h
juandent: removed!

jakemumu:
Excited to see this PR -- I've been watching this Repo but have been a bit scared of the no migrations approach of this framework -- although it is by far the best C++ SQL wrapper I've seen

juandent:
@jakemumu we have been working towards supporting migrations better and it is good to hear interest in our clients!
Do you know other potential clients that have this fear of using the tool?? if we can get a group of people interested in migrations then we could develop this area further!

We hope we can support better migrations - although already we have done some progress and we are documenting it. Visit our website at https://sqliteorm.com and see the Documentation tab, look for a PDF it has a section on migrations

image

jakemumu:
@juandent amazing! I had not seen this new website nor documentation. I am not aware of others unfortunately -- but my experience comes from the way Ruby on Rails handles migrations. If you haven't looked at how they support I recommend. 

I understand the nice simplistic support this framework offers currently, however it's also good to have a hard record of the individual changes which occurred. 

it would mean a schema declaration my look more like this on launch of a DB: 
```
make_storage(...)
```

Then a change occurs: 

```
// The Same Forever
make_storage(...)

// Date: XX/XX/XX - Adds X to Y Table
add_column(TABLE, COLUMN)

// Date: XX/XX/XX - Drops X from Y Table
drop_column(...)
```

Then instead of the framework tracking only the initial -- it would check each "migration" call. 

Just an idea! Perhaps this is somehow similar to how this already works on the sync schema -- I'm excited to review the new documentation! Nice work!

fnc12:
> what if we test to see if we are in a transaction at the entry to backup_table, and if it is in a transaction we do not start another... I understand what you say but it is necessary to make sure we cleanup --
what about the tables that were created ... they will soon fill the database with backup tables that could have been removed by the call to rollback.

> Have you thought about this? I understand your point completely... but there is the cleanup issue here which I think should be considered, IMHO

So cleanup will never happen if user called `sync_schema` within a transaction right?

@jakemumu thanks for the response. I helps a lot. We'll try to make it the best

dev/implementations/storage_definitions.h
```diff
@@ -126,5 +188,84 @@ namespace sqlite_orm {
...
+                ss << columnNames[i];
+                if(i < columnNamesCount - 1) {
+                    ss << ", ";
+                }
+            }
+            ss << " FROM " << quote_identifier(tableName);
+            perform_void_exec(db, ss.str());
+        }
+#endif
+        // jdh
```
fnc12, line 231: please don't commit comments saying about the author. We can inspect author of any line using `git blame` instead

dev/storage.h
```diff
@@ -124,16 +124,39 @@ namespace sqlite_orm {
                 }
                 perform_void_exec(db, ss.str());
             }
-
+#if 0  // jdh
```
fnc12, line 127: are you sure that this code is needed?

dev/storage.h
```diff
@@ -142,33 +165,79 @@ namespace sqlite_orm {
...
+                bool started_local_transaction = false;
+                try {
+                    started_local_transaction = this->start_migration();
+
+                    this->drop_table_internal(tImpl.table.name, db);
+                    this->create_table(db, tImpl.table.name, tImpl);
+
+                    this->commit_migration(started_local_transaction);
+                } catch(const std::exception&) {
+                    this->abort_migration(started_local_transaction);
```
fnc12, line 193: `abort_migration` will call `ROLLBACK`. Controlling transactions from inside a storage is not a good practice

dev/storage.h
```diff
@@ -826,6 +895,43 @@ namespace sqlite_orm {
             }
 
           protected:
+#if 0  // jdh  moved to storage_base.h
```
fnc12, line 898: please skip committing the same code under `#if 0` - just remove it

dev/storage.h
```diff
@@ -1014,7 +1125,8 @@ namespace sqlite_orm {
                 std::map<std::string, sync_schema_result> result;
                 auto db = con.get();
                 this->impl.for_each([&result, db, preserve, this](auto& tableImpl) {
-                    auto schemaStatus = this->schema_status(tableImpl, db, preserve);
+                    bool attempt_to_preserve = true;
+                    auto schemaStatus = this->schema_status(tableImpl, db, preserve, attempt_to_preserve);
```
fnc12, line 1129: `attempt_to_preserve` is passed here but never used. Better accept is as a pointer not ref to `schema_status` and pass `nullptr` when we don't need it

dev/storage.h
```diff
@@ -1027,9 +1139,12 @@ namespace sqlite_orm {
              */
             bool table_exists(const std::string& tableName) {
                 auto con = this->get_connection();
-                return this->impl.table_exists(tableName, con.get());
+                return this->table_exists(tableName, con.get());
             }
 
+            // jdh
```
fnc12, line 1145: please remove such comments. We use git especially for this - to know who wrote any line

dev/storage_base.h
```diff
@@ -450,6 +483,15 @@ namespace sqlite_orm {
                 return this->connection->retain_count() > 0;
             }
 
+            /*
+             * returning false when there is a transaction in place
+             * otherwise true; function is not const because it has to call get_connection()
+             */
+            bool get_autocommit() {  //jdh
```
fnc12, line 490: please remove `jdh`

fnc12:
> This is no longer an issue .. it has been fixed!!]

fnc12:
@jakemumu so you don't use `sqlite_orm` due to lack of migrations or because you are afraid that calling `sync_schema` can drop data? Of you use `sqlite_orm` but never call `sync_schema`?

fnc12 (changes requested):

dev/storage.h
fnc12: > What's wrong with using exceptions inside a library?? I don't see it..

As I said if you catch an exception inside storage you violate catching exception scenario for user in the client's code. Especially when you call another function inside `catch` which can also throw. There is no `try`/`catch` code inside the lib. This is what I am talking about

fnc12: > how would we access this rc and why, the rc has become an exception if it is not ok!

Then we need to return rc from thins function without throwing. Or just remove `try`/`catch` block to eliminate this problem. It is your PR and it is up to you

dev/implementations/storage_definitions.h
juandent: ok! it's just the way i mark my code where I have done changes so it is easy for me to jump around my changes but I will remove them!!

dev/storage.h
juandent: eliminated and no copy_table_from does not exist anymore - check the latest commit please!

dev/storage.h
juandent: Did you not see I was checking for an existing transaction in start_migration? if there is a transaction in place I don't begin another!

dev/storage.h
juandent: ok

dev/storage.h
juandent: go

dev/storage_base.h
```diff
@@ -728,6 +770,27 @@ namespace sqlite_orm {
...
+                return false;
+            }
+            void commit_migration(bool started_local_trans) {
+                if(started_local_trans) {
+                    this->commit();
+                }
+            }
+            void abort_migration(bool started_local_trans) {
+                if(started_local_trans) {
+                    this->rollback();
```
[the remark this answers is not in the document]
juandent: removed transactions everywhere

dev/storage_base.h
juandent: ok

dev/storage.h
```diff
@@ -142,33 +143,54 @@ namespace sqlite_orm {
...
-                        if(!tableImpl.table_exists(anotherBackupTableName, db)) {
+                        if(!table_exists(anotherBackupTableName, db)) {
                             backupTableName = anotherBackupTableName;
                             break;
                         }
                         ++suffix;
                     } while(true);
                 }
-
+                // eliminated all transaction handling
```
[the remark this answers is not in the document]
juandent: corrected

dev/storage.h
```diff
@@ -884,6 +911,11 @@ namespace sqlite_orm {
                                 } else {
                                     if(columnPointer->notnull && columnPointer->dflt_value.empty()) {
                                         gottaCreateTable = true;
+                                        // no matter if preserve is true or false, there is no way to preserve data, so we wont try!
+                                        if(attempt_to_preserve) {
+                                            *attempt_to_preserve = false;
+                                        };
+                                        // res = decltype(res)::dropped_and_recreated_with_loss;
```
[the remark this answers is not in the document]
juandent: corrected

dev/storage_base.h
```diff
@@ -728,6 +770,27 @@ namespace sqlite_orm {
                 }
             }
 
+            // migration internal API
+            // returns whether we started a transaction in this function
+            bool start_migration() {
```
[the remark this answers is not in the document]
juandent: no where -- I am sorry... I removed these functions

dev/implementations/storage_definitions.h
```diff
@@ -122,9 +144,47 @@ namespace sqlite_orm {
                     ss << ", ";
                 }
             }
-            ss << " FROM " << quote_identifier(tImpl.table.name);
+            ss << " FROM " << quote_identifier(sourceTableName);
             perform_void_exec(db, ss.str());
         }
 
+        inline bool storage_base::calculate_remove_add_columns(std::vector<const table_xinfo*>& columnsToAdd,
```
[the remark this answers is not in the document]
juandent: corrected

jakemumu:
> @jakemumu so you don't use `sqlite_orm` due to lack of migrations or because you are afraid that calling `sync_schema` can drop data? Of you use `sqlite_orm` but never call `sync_schema`?

Sorry for hijacking this thread! I don't yet use this library in any projects but I have created a prototype of a database I'd like to deploy this year. I made the prototype back before the new license and all the awesome work you guys are doing! 

To be clear I am planning to purchase a license this year and to start using this in production. I also watched your conference talks about the library. 

`you are afraid that calling `sync_schema` can drop data.` <- this is the case. Even after reading the explanation of sync_schema it feels difficult for me to have a clear understanding of situations where data could potentially be lost. I would think that the no matter what under any circumstances, data should never be dropped.

fnc12:
> Sorry for hijacking this thread!

You are always welcome!.

> you are afraid that calling sync_schema can drop data. <- this is the case. Even after reading the explanation of sync_schema it feels difficult for me to have a clear understanding of situations where data could potentially be lost.

Try to call `storage.sync_schema_simulate()` to see what will happen to your tables. Once you try to use it you'll see that it is easy to understand how it works

fnc12:
@juandent now once this PR becomes cleaner let's make a unit test which must fail without your update but doesn't fail with your update. It will show people how it is intended to work and will guarantee that your work will not be broken in future by someone else

juandent:
no need to add unit_tests since all generated columns unit tests in sync_schema_tests.cpp pass correctly with slight modification to sqlite_orm

juandent:
@jakemumu "you are afraid that calling sync_schema can drop data", in an upcoming release I will update the Guide and the Migration Data section and I will explain what you need to do to prevent data loss and so you know which changes are not guaranteed to lose data and which are.. Please be patient and look for v2.6 of the Guide in the Documentation tab of out website https::/sqliteorm.com. I am the author of that guide which you should also check out to learn how to use sqlite_orm if you think in SQL terms...

fnc12 (approved):

juandent:
@jakemumu concerning the issue of data preservation across schema migrations, please contact me at juandent@mac.com. I am the author of the guide in our website and am preparing more information that should interest you

PR #1022, opened 2022-04-25
Code quality

trueqbit:
* Used feature detection macros for polyfills
* Fixed spelling of "limit_accesor" -> "limit_accessor"
* Included dependent headers
* Removed some superfluous constructors that are just forwarding their arguments to initialize members
* Qualified enum values by their original enum class name instead of using `decltype`
* Spelled out type for SQLite database connection instead of using `auto` (I am not a friend of auto everywhere, because it makes code harder to understand)
* Moved `storage_t<>::drop_trigger()` to base class `storage_base`, changed signature to accept a `std::string`
* Addressed warning when streaming `conflict_action`
* Used `polyfill::negation` instead of `static_not`
* `mapped_type_proxy_t`
* Simplified member traits
* Gone is all the boiler-plate code when preparing, binding to, and executing statements


dev/storage_base.h
```diff
@@ -47,19 +45,24 @@ namespace sqlite_orm {
             void drop_index(const std::string& indexName) {
                 std::stringstream ss;
                 ss << "DROP INDEX " << quote_identifier(indexName);
-                perform_void_exec(get_connection().get(), ss.str());
+                perform_void_exec(this->get_connection().get(), ss.str());
+            }
+
+            void drop_trigger(const std::string& triggerName) {
```
trueqbit, line 51: @fnc12 This was a [templated method](https://github.com/fnc12/sqlite_orm/blob/451a81b774123d07f45413db4f4604aae0773fb1/dev/storage.h#L217) previously. Was there any reason for this?

dev/storage_base.h
```diff
@@ -328,7 +331,6 @@ namespace sqlite_orm {
                 std::stringstream ss;
                 ss << C::name();
                 auto name = ss.str();
-                ss.flush();
```
fnc12, line 331: removing this will flush in the end of the scope -> after `this->create_collation(name, move(func));`. Better to drop unused memory ASAP

dev/storage_base.h
```diff
@@ -360,7 +362,6 @@ namespace sqlite_orm {
...
```
fnc12, line 363: same here

dev/table_info.h
```diff
@@ -12,6 +14,7 @@ namespace sqlite_orm {
         std::string dflt_value;
         int pk = 0;
 
+#ifndef SQLITE_ORM_AGGREGATE_PAREN_INIT
```
fnc12, line 17: what is the purpose of `SQLITE_ORM_AGGREGATE_PAREN_INIT`? It covers all trivial constructors?

patch1.diff
```diff
@@ -0,0 +1,25 @@
...
+
+---
+ dev/statement_serializer.h | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/dev/statement_serializer.h b/dev/statement_serializer.h
+index 10c2a67..f2da5e5 100644
+--- a/dev/statement_serializer.h
++++ b/dev/statement_serializer.h
+@@ -1774,6 +1774,8 @@ namespace sqlite_orm {
```
fnc12, line 14: looks like a redundant file

fnc12:
@trueqbit please remove `.diff` files

fnc12 (changes requested):

dev/storage_base.h
trueqbit: I am just wondering then why the other methods don't accept "any thing"?

dev/cxx_polyfill.h
```diff
@@ -27,50 +45,31 @@ namespace sqlite_orm {
...
-
-            template<template<typename...> class Primary, class... Types>
-            SQLITE_ORM_INLINE_VAR constexpr bool is_specialization_of_v<Primary<Types...>, Primary> = true;
-
-            template<typename Type, template<typename...> class Primary>
-            struct is_specialization_of : bool_constant<is_specialization_of_v<Type, Primary>> {};
-
-            template<typename... T>
-            using is_specialization_of_t = typename is_specialization_of<T...>::type;
+#if 0  // __cpp_lib_detect >= 0L  //  library fundamentals TS v2, [meta.detect]
```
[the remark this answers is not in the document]
trueqbit: Because there's no known date yet, it's still experimental. There's `__cpp_lib_experimental_detect`, however I don't want to use experimental because usually it's going to be removed at some point in the future.

dev/cxx_polyfill.h
```diff
@@ -108,6 +107,24 @@ namespace sqlite_orm {
             SQLITE_ORM_INLINE_VAR constexpr bool is_detected_v = is_detected<Op, Args...>::value;
 #endif
 
+#if 1  // proposed but not pursued                                                                                     \
```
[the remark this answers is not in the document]
trueqbit: There's no feature detection macro, however I wanted to express that it's something to be filled in later.

dev/statement_serializer.h
```diff
@@ -1774,6 +1774,8 @@ namespace sqlite_orm {
                         return "IGNORE";
                     case conflict_action::rollback:
                         return "ROLLBACK";
+                    default:
```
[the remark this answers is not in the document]
trueqbit: Why exactly? Not having it also produces a warning with msvc.

dev/table_info.h
trueqbit: I am not sure I exactly understand what you mean by whether it covers all trivial constructors.
C++20 allows to initialize aggregates via parentheses, so e.g. `table_info()` works as well as `vector<table_info>::.emplace_back()`. So this constructor is no longer needed.

patch1.diff
trueqbit: gitkraken has the annoying feature of automatically staging new files.

dev/conditions.h
```diff
@@ -421,11 +421,6 @@ namespace sqlite_orm {
         struct order_by_base {
             int asc_desc = 0;  //  1: asc, -1: desc
             std::string _collate_argument;
-
-            order_by_base() = default;
```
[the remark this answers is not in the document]
trueqbit: I suggest not to keep legacy around? If it's a problem in a certain version of Visual C++ I rather surround it using `#if defined(_MSC_VER) && (_MSC_VER < 1920)` or alike. Only very old VC++ versions had some problems like with `is_base_of_template`.

dev/storage_base.h
fnc12: cause it is legacy.
yes `std::string_view` is a perfect solution but it is not available in C++14.

dev/cxx_polyfill.h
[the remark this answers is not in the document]
fnc12: so is it ok that we'll commit `#if 0` into `dev`?

dev/cxx_polyfill.h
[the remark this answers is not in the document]
fnc12: are you going to merge in as is into `dev` or is it still in progress?

dev/statement_serializer.h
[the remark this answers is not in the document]
fnc12: hm alright. Let it be that way cause VS is more popular then Xcode and I reckon I can disable this warning in Xcode

dev/storage_base.h
fnc12: not new, it is very old but yes it is not done consistently cause I added it only when I remembered about it.
> It's more code for little gain. 
this is how C and C++ work 🙂. It is not so critical cause no one asked for it yet. So it is up to you

dev/table_info.h
fnc12: I mean do we need to use `SQLITE_ORM_AGGREGATE_PAREN_INIT` for defining ctors in every struct?

patch1.diff
fnc12: try GitHub desktop (or git bash)

dev/conditions.h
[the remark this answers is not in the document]
fnc12: hm alright. Let's wait until one more developer comes with such an issue and try to fix it differently. Really I created and deleted such ctors several times before again in again.

dev/table_info.h
trueqbit: No, only those that will be constructed using `make_unique` or `emplace_back`, which call the parentheses constructor instead of the brace initializer. In this specific case, table_info gets emplaced into the container. Which in turn is the only one of that kind in sqlite_orm.

dev/storage_base.h
trueqbit: No worries, I made it state-of-the-art with a later commit in this PR.

dev/statement_serializer.h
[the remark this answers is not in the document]
trueqbit: Where do you get warnings? I'll try with clang. Let's not replace one warning with another.

dev/cxx_polyfill.h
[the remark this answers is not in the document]
trueqbit: It's going into dev like this - note that nothing changed.
However if you have a better suggestion, feel free to let me know :)

dev/cxx_polyfill.h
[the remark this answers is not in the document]
trueqbit: It's going into dev like this - note that nothing changed (it's just the inverse of the previous `#if 0`.
However if you have a better suggestion, feel free to let me know :)

dev/conditions.h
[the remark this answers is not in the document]
trueqbit: Yep, I have seen 3 tickets in regard to VC 2015 Update 3. For now I have reinstated those constructors using a proper feature detection macro (support for "Aggregate classes with default member initializers").

dev/storage_base.h
trueqbit: I assume that `drop_trigger` came after `drop_index` and `drop_table`? If so, can we just adjust `drop_trigger` to the other methods, and possibly make a feature request?

dev/statement_serializer.h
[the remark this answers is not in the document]
fnc12: on the `default` line saying 'this code will never be executed'

fnc12:
IDK why I cannot comment right inside conversation so I'll answer there with a quote.

> I assume that drop_trigger came after drop_index and drop_table? If so, can we just adjust drop_trigger to the other methods, and possibly make a feature request?

What do you mean by feature request? Feature request with what?

dev/storage_base.h
trueqbit: If somebody wants to pass a different thing than a std::string, that would be a feature request. Either we create that request ourselves or somebody in need.

fnc12: oh I see. Ok let's leave it like that. Probably we are discussing about something not used at all.

dev/implementations/storage_impl_definitions.h
```diff
@@ -21,11 +21,10 @@ namespace sqlite_orm {
             bool result = false;
             std::stringstream ss;
             ss << "SELECT COUNT(*) FROM sqlite_master WHERE type = " << quote_string_literal("table"s)
-               << " AND name = " << quote_string_literal(tableName);
-            auto query = ss.str();
+               << " AND name = " << quote_string_literal(tableName) << std::flush;
             auto rc = sqlite3_exec(
                 db,
-                query.c_str(),
+                ss.str().c_str(),
```
fnc12, line 27: tip: stopping keeping `query` as a dedicated variable makes it harder to debug

tests/static_tests/node_tuple.cpp
```diff
@@ -768,7 +768,7 @@ TEST_CASE("Node tuple") {
         }
 #endif
         SECTION("coalesce") {
-            auto f = char_(10, 20);
+            auto f = coalesce(10, 20);
```
fnc12, line 771: lol thanks

tests/storage_tests.cpp
```diff
@@ -259,7 +269,7 @@ TEST_CASE("issue880") {
                                 foreign_key(&X::fkey_Rendimiento).references(&Rendimiento::id)));
     storage.sync_schema();
 
-    Fondo fondo{5};
```
fnc12, line 262: why did you remove 5?

fnc12 (approved):
wow this is a lot of good job done here! I am impressed!

tests/storage_tests.cpp
trueqbit: What's the point?

tests/CMakeLists.txt
```diff
@@ -133,6 +133,11 @@ endif()
 
 if (MSVC)
 	add_compile_options(/MP) # Enable multi parrallel build
+	if (MSVC_VERSION LESS_EQUAL 1900)
```
trueqbit, line 136: I don't see the reason why appveyor doesn't pick up this condition and does not disable those specific warnings? [See appveyor build](https://ci.appveyor.com/project/fnc12/sqlite-orm/builds/43393529/job/r9kh99ga09hrqhm8).

tests/storage_tests.cpp
fnc12: 5 is checked at runtime. If you remove 5 it will be equal to 0. I know that the result is not checked and it will not alter UT results. But I don't see the reason to change this. Anyway it is just a question

appveyor.yml
```diff
@@ -27,13 +24,19 @@ platform:
   
 environment:
   matrix:
+   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
+     SQLITE_ORM_CXX_STANDARD: ""
+
    # using c++14
-   - SQLITE_ORM_CXX_STANDARD: "-DSQLITE_ORM_ENABLE_CXX_14=ON"
+   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022
```
trueqbit, line 31: There's still something wrong with the environment matrix: appveyor shouldn't build using Visual Studio 2022 with no `SQLITE_ORM_CXX_STANDARD` at all. [See appvyeor build](https://ci.appveyor.com/project/fnc12/sqlite-orm/builds/43393952).

fnc12: stranger things. Let's try to figure it out

tests/CMakeLists.txt
fnc12: Actually I am not a master in CMake and CI configs but let me try to read it and understand how it works

fnc12:
everything works on my mac but CI got to be fixed

tests/storage_tests.cpp
trueqbit: The reason was simply not having to provide a conditional constructor (due to a compiler's possibly missing default member initializer support).
But you have a valid point. Let me change it back.

tests/static_tests/node_tuple.cpp
trueqbit: one unit test more, yay!

dev/implementations/storage_impl_definitions.h
trueqbit: I thought that the code is simple enough. Note that the stream gets flushed, so you will see the full resulting string in the stringstream variable.

appveyor.yml
trueqbit: I am looking at appveyor's [build environment doco](https://www.appveyor.com/docs/build-environment/#using-multiple-images-for-the-same-build), which I would think is straightforward.

trueqbit: It's obvious once you see things through the eyes of a yaml parser: `- SQLITE_ORM_CXX_STANDARD` shouldn't be on the same level with a dash :roll_eyes:

dev/object_from_column_builder.h
```diff
@@ -27,10 +32,10 @@ namespace sqlite_orm {
                 object_from_column_builder_base{stmt_}, object(object_) {}
 
             template<class C>
-            void operator()(const C& c) const {
+            void operator()(const C& c) {
```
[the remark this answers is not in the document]
trueqbit: Yes, because a 'lambda' passed to the tuple iteration function doesn't need to be const anymore.

dev/backup.h
```diff
@@ -32,14 +33,12 @@ namespace sqlite_orm {
...
-                handle(other.handle), holder(move(other.holder)), to(other.to), from(other.from) {
-                other.handle = nullptr;
-            }
+                handle(std::exchange(other.handle, nullptr)), holder(move(other.holder)), to(other.to),
+                from(other.from) {}
 
             ~backup_t() {
                 if(this->handle) {
                     (void)sqlite3_backup_finish(this->handle);
-                    this->handle = nullptr;
```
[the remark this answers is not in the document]
trueqbit: I've never seen good code that relies on 'nulling' out members in a destructor, and I am not aware of any standard library implementer doing it. I have never needed this kind of debugging technique ever in my programmer life. If you want we can make it conditional via `#ifndef NDEBUG`?

dev/ast_iterator.h
```diff
@@ -39,18 +39,18 @@ namespace sqlite_orm {
              *  L is a callable type. Mostly is a templated lambda
              */
             template<class L>
-            void operator()(const T& t, const L& l) const {
-                l(t);
+            void operator()(const T& t, L& lambda) const {
```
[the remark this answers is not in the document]
trueqbit: Because here it just preserves the const/non-const qualification.

dev/static_magic.h
```diff
@@ -17,27 +17,24 @@ namespace sqlite_orm {
         }
 
         template<typename T, typename F>
-        decltype(auto) static_if(std::true_type, const T& t, const F&) {
-            return (t);
+        decltype(auto) static_if(std::true_type, T&& t, const F&) {
+            return static_cast<T&&>(t);
```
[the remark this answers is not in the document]
trueqbit: As written above, just for preserving the way it was passed in.

dev/table.h
```diff
@@ -116,10 +119,10 @@ namespace sqlite_orm {
              *  Calls **l** with every primary key dedicated constraint
              */
             template<class L>
-            void for_each_primary_key(const L& lambda) const {
+            void for_each_primary_key(L&& lambda) const {
```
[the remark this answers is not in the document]
trueqbit: Here we can do that, I updated it in line with the other 'lambda' changes, such that we don't care here what the function object does.

dev/backup.h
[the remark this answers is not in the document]
fnc12: ok let's leave it like this

dev/ast_iterator.h
```diff
@@ -683,8 +683,8 @@ namespace sqlite_orm {
             using node_type = collate_t<T>;
 
             template<class L>
-            void operator()(const node_type& node, const L& l) const {
-                iterate_ast(node.expr, l);
+            void operator()(const node_type& node, L& lambda) const {
```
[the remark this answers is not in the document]
fnc12: alright. I see the meaning. Let's leave it like this. Thanks

fnc12:
Снимок экрана 2022-05-05 в 11 29 55

fnc12 (approved):

trueqbit:
> C++14
Now I am surprised! Even VC 2015 already shipped `std::negate`.

fnc12:
Снимок экрана 2022-05-06 в 19 32 09
how it worked on CI?

trueqbit:
> Снимок экрана 2022-05-06 в 19 32 09
> how it worked on CI?

Because msvc doesn't yet report shadowing template paramters (I may have missed a compiler option).

fnc12:
thanks @trueqbit

PR #1026, opened 2022-05-05
Added first draft DbConnection.h and added an overload for make_storage()

juandent:
so the connection stays open  when made through DcConnection overload.

Please see changes to storage_base protected constructor: 

```C++
            storage_base(const DbConnection& con, int foreignKeysCount) :
                pragma(std::bind(&storage_base::get_connection, this)),
                limit(std::bind(&storage_base::get_connection, this)),
                inMemory(con.filename().empty() || con.filename() == ":memory:"),
                connection(std::make_unique<connection_holder>(con.filename())),
                cachedForeignKeysCount(foreignKeysCount) {
                this->connection->retain();  // make connection stay open!!! Look HERE!
                this->on_open_internal(this->connection->get());
            }
```

### DIFF
--- a/dev/DbConnection.h
+++ b/dev/DbConnection.h
@@ -1,0 +1,30 @@
+#pragma once
+#include <string>
+#include "connection_holder.h"
+
+namespace sqlite_orm {
+
+    struct DbConnection {
+        DbConnection(const std::string& filename) : m_filename(filename) {}
+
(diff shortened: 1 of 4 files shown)


juandent:
This PR was useful for me to experience what it would take to add a new class and a new overload to make_storage
I found a lot more about the design of sqlite_orm!!

PR #1030, opened 2022-05-06
Testing appveyor builds on linux

trueqbit:
Just a test PR for trying multi-platform appveyor builds.


trueqbit:
Appveyor build script and all fixes found compiling with gcc and clang are part of PR #1033.

PR #1034, opened 2022-05-11
Code quality 2

trueqbit:
* More appveyor test builds: clang - C++14, clang - C++17, gcc - C++14, gcc - C++17.
* Moved central C++ features on which sqlite_orm depends to universal header file.
* Moved symbols local to translation unit of unit tests into anonynous namespace.
* Replaced `assert`ions in unit tests with `REQUIRE`s (more test coverage!).
* Removed `const_cast` from generic `get<>(const prepared_statement_t<>)`.
* Unit test for generic `get<>(const prepared_statement_t<>)`
* Tuple iteration, filtering has been further simplified.
* Binding of object fields has been further simplified.
* `call_if_constexpr`, constexpr implementation for `static_if`
* Made some uses of `static_if` obsolete by filtering tuples before iteration
* Encapsulated result set iteration in a dedicated function


dev/implementations/storage_definitions.h
```diff
@@ -66,7 +66,7 @@ namespace sqlite_orm {
                         }
 
                         if(schema_stat == sync_schema_result::new_columns_added) {
-                            for(auto columnPointer: columnsToAdd) {
+                            for(auto* columnPointer: columnsToAdd) {
```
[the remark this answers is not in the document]
trueqbit: Just noting here: there's a few other places, which I'll change as well.

dev/functional/static_magic.h
```diff
@@ -35,6 +61,12 @@ namespace sqlite_orm {
         decltype(auto) static_if(T&& t) {
             return static_if(std::integral_constant<bool, B>{}, std::forward<T>(t), empty_callable());
         }
+
+        template<bool B, typename L, typename... Args>
+        void call_if_constexpr(L&& lambda, Args&&... args) {
+            static_if<B>(static_cast<L&&>(lambda))(static_cast<Args&&>(args)...);
```
fnc12, line 67: why don't you use `move` instead of `static_cast<T&&>`?

dev/storage.h
```diff
@@ -1304,99 +1273,75 @@ namespace sqlite_orm {
...
             template<class T, class... Ids>
             T execute(const prepared_statement_t<get_t<T, Ids...>>& statement) {
-                auto con = this->get_connection();
                 sqlite3_stmt* stmt = reset(statement.stmt);
-                auto& tImpl = this->get_impl<T>();
 
                 iterate_ast(statement.expression.ids, conditional_binder{stmt});
+
+#ifdef SQLITE_ORM_OPTIONAL_SUPPORTED
+                std::optional<T> res;
```
fnc12, line 1334: why that? are sure that we need one more branching here?

dev/storage_base.h
```diff
@@ -229,11 +231,11 @@ namespace sqlite_orm {
                     },
                     /* call = */
                     [](sqlite3_context* context, void* functionVoidPointer, int argsCount, sqlite3_value** values) {
-                        auto& functionPointer = *static_cast<F*>(functionVoidPointer);
+                        auto& fn = *static_cast<F*>(functionVoidPointer);
```
fnc12, line 234: better make less acronyms to make code more readable for newbies and us

dev/table.h
```diff
@@ -55,24 +56,35 @@ namespace sqlite_orm {
...
+            template<class M, satisfies<is_setter, M> = true>
+            const member_field_type_t<M>* object_field_value(const object_type& object, M memberPointer) const {
+                using F = member_field_type_t<M>;
                 const F* res = nullptr;
                 this->for_each_column_with_field_type<F>([&res, &memberPointer, &object](auto& column) {
                     if(res) {
                         return;
                     }
 
-                    if(compare_any(column.member_pointer, memberPointer) || compare_any(column.setter, memberPointer)) {
```
fnc12, line 75: this is pretty enough

dev/table.h
```diff
@@ -209,38 +220,36 @@ namespace sqlite_orm {
              *  Counts and returns amount of columns. Skips constraints.
              */
             constexpr int count_columns_amount() const {
-                return int(filter_tuple_sequence_t<elements_type, is_column>::size());
+                using col_seq = filter_tuple_sequence_t<elements_type, is_column>;
```
fnc12, line 223: same note about shortened identifiers: full name is better for readability

dev/storage_base.h
trueqbit: So, leave `functionPointer`?

dev/functional/static_magic.h
trueqbit: The arguments are just forwarded and comes w/o the cost of including <utility> for `std::forward`.
But it seems to cause confusion?

dev/storage.h
trueqbit: With branching I guess you mean the `#ifdef`? If reusing the `perform_steps()` function is desired then yes.

dev/table.h
trueqbit: Is it an appeal or an exclamation? :)

dev/storage_base.h
fnc12: yep

dev/functional/static_magic.h
fnc12: up to you

dev/storage.h
fnc12: ok. One more branch candidate for removal after C++14 support dropping appeared

dev/table.h
fnc12: an exclamation =)

fnc12 (commented):
let's rename shortened identifier back please

dev/functional/static_magic.h
trueqbit: Well, it's causing confusion, otherwise you wouldn't ask :) I just use the static cast because it's an internal implementation detail that even the STL guys at Microsoft use.
But you do have to know what the argument is, and `forward` expresses intent.

fnc12: I changed my mind. Let's place an intent here also. Many developers know about forward and move but not many developers know that these functions contain static cast inside

trueqbit: Agreed. I doubt the "many developers" part, because to know about forwarding you need to know about "universal references" and about how forwarding works. But there's no question that programming is about expressing intent, which makes things easier to read in an already complex piece of work!

dev/table.h
```diff
@@ -209,38 +220,36 @@ namespace sqlite_orm {
...
-                    static_if<is_column_v<element_type>>(lambda)(element);
-                });
+            void for_each_foreign_key(L&& lambda) const {
+                using fk_index_sequence = filter_tuple_sequence_t<elements_type, is_foreign_key>;
+                iterate_tuple(this->elements, fk_index_sequence{}, lambda);
             }
 
+            /**
+             *  Call passed lambda with all defined columns.
+             *  @param lambda Lambda to be called for each column. Must have signature like this [] (auto col) -> void {}
```
fnc12, line 239: nit: I'd change `auto col` to `auto& col`. Showing an example with redundant copying is a bad idea

tests/static_tests/functional/static_if_tests.cpp
```diff
@@ -64,4 +66,18 @@ TEST_CASE("static_if") {
             })();
         REQUIRE(called == -1);
     }
+    {  //  simple true
+        auto called = false;
+        call_if_constexpr<std::true_type::value>([&called] {
```
fnc12, line 71: if we have a bug when lambda is called twice this bug will never be detected with this UT. Better change `bool called` to `int called`, make `++called` inside lambda and `REQUIRED(called == 1)`

fnc12 (approved):

dev/table.h
trueqbit: Haha you are being funny! That was copy/paste from existing code (most likely from you).

fnc12: Lol ok let's leave it like that within this pr

tests/static_tests/functional/static_if_tests.cpp
trueqbit: I can change this one. Though it's not different from the other `static_if` tests above.

dev/table.h
trueqbit: I have changes there in another upcoming PR, so will update it there :)

fnc12:
thanks @trueqbit

PR #1042, opened 2022-05-29
Runner-up: tuple filter

trueqbit:
* Could get rid of 'constexpr if' logic in `storage_t<>::has_dependent_rows()`
* Simplified internal table interface
* Corrected comment on `pass_extracted_fn_to`


fnc12:
thanks @trueqbit

PR #1044, opened 2022-05-30
Iterate schema objects instead of 'storage impl' wrappers

trueqbit:
All runtime code parts now work with the schema objects themselves (i.e. tables, indexes, triggers) instead of the `storage_impl<>` wrapper. This lays the final groundwork for replacing `storage_impl` with a tuple of schema objects in the future (when it is decided that the wrapper is no longer useful).

This change also allows proper filtering of schema objects before iteration - e.g. `storage_t<>::has_dependent_rows()` may only operate on table definitions.

Fixes:
*  `storage_t<>::has_dependent_rows()` was previously choking on schema definitions that contained schema objects other than tables.
* Reversed iteration of tuples using fold expressions is now correctly taking the base index of a given index sequence into account.

### DIFF
--- a/dev/storage.h
+++ b/dev/storage.h
@@ -1491,8 +1491,8 @@ namespace sqlite_orm {
     }
 
     template<class... Ts>
-    internal::storage_t<Ts...> make_storage(const std::string& filename, Ts... tables) {
-        return {filename, internal::storage_impl<Ts...>{std::forward<Ts>(tables)...}};
+    internal::storage_t<Ts...> make_storage(std::string filename, Ts... tables) {
+        return {move(filename), internal::storage_impl<Ts...>{std::forward<Ts>(tables)...}};
     }
 
     /**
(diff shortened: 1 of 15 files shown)


dev/storage.h
```diff
@@ -999,9 +984,9 @@ namespace sqlite_orm {
             std::map<std::string, sync_schema_result> sync_schema(bool preserve = false) {
```
trueqbit, line 984: @fnc12 Now that it is possible to iterate specific schema objects (filter), shall we create tables first, then all other schema objects? This would make it possible to specify indexes and triggers anywhere in the storage definition.

trueqbit (commented):
It would be possible now to specify indexes and triggers anywhere in the storage definition.

fnc12 (approved):

fnc12:
thanks @trueqbit

PR #1052, opened 2022-06-13
Implemented sys_days as bindable type; a C++ 20 feature

juandent:


dev/date_binding.h
```diff
@@ -0,0 +1,127 @@
...
+    /**
+	 *  First of all is a type_printer template class.
+	 *  It is responsible for sqlite type string representation.
+	 *  We want SysDays to be `TEXT` so let's just derive from
+	 *  text_printer. Also there are other printers: real_printer and
+	 *  integer_printer. We must use them if we want to map our type to `REAL` (double/float)
+	 *  or `INTEGER` (int/long/short etc) respectively.
+	 */
+    template<>
+    struct type_printer<std::chrono::sys_days> : public text_printer {};
```
fnc12, line 70: what if user wants to store `sys_days` as integer?

dev/functional/cxx_core_features.h
```diff
@@ -70,3 +70,7 @@
 #if __cpp_concepts >= 201907L
 #define SQLITE_ORM_CONCEPTS_SUPPORTED
 #endif
+
+#if __cpp_lib_three_way_comparison >= 201907L
+#define SQLITE_ORM_THREE_WAY_COMPARISON_SUPPORTED
```
fnc12, line 75: what is three way comparison?

tests/chrono_binding.cpp
```diff
@@ -0,0 +1,41 @@
...
+    const std::string db_name = "sys_days.sqlite";
+    ::remove(db_name.c_str());
+
+    auto storage = make_storage(db_name,
+                                make_table("Persons",
+                                           make_column("id", &Person::id, primary_key(), autoincrement()),
+                                           make_column("name", &Person::name),
+                                           make_column("birthdate", &Person::birthdate)));
+
+    SECTION("YMD_sys_date") {
```
fnc12, line 25: nit: no need to add section if it is only one. Sections are helpful if you have to split the scenario

tests/statement_serializer_tests/schema/index.cpp
```diff
@@ -52,3 +52,22 @@ TEST_CASE("statement_serializer index") {
...
+    auto storage = make_storage("",
+                                make_index("user_name", &User::name),
+                                make_table("users",
+                                           make_column("id", &User::id, primary_key(), autoincrement()),
+                                           make_column("name", &User::name)));
+    storage.sync_schema(true);
+    storage.sync_schema_simulate(true);  // this will call in storage_t:
+
+    // template<class... Tss, class... Cols>
+    // sync_schema_result schema_status(const storage_impl<index_t<Cols...>, Tss...>&, sqlite3*, bool) {
```
fnc12, line 70: no comments

tests/statement_serializer_tests/schema/index.cpp
```diff
@@ -52,3 +52,22 @@ TEST_CASE("statement_serializer index") {
...
+        int id = 0;
+        std::string name;
+    };
+    auto storage = make_storage("",
+                                make_index("user_name", &User::name),
+                                make_table("users",
+                                           make_column("id", &User::id, primary_key(), autoincrement()),
+                                           make_column("name", &User::name)));
+    storage.sync_schema(true);
+    storage.sync_schema_simulate(true);  // this will call in storage_t:
```
fnc12, line 67: what this test tests?

fnc12 (changes requested):
this is good code but it is better to be placed inside `examples` folder cause the way you serialize/deserialize is not the only way people would like to use for date serialization. Some devs like to serialize data as integer. We cannot make them do it always like we like to do it.

dev/date_binding.h
juandent: he would have to provide his own implementation for sys_days binding...

dev/functional/cxx_core_features.h
juandent: a new operator in C++ 20:   <=>, which replaces the need to define <,<=,>,>=,!=,== if you provide it! The compiler can also provide it for you if you use =default.
It is a dependency of the chronos library and sys_day requires C++ 20, but there is no __cpp_lib_sysday so I used the feature testing 3 way comparison instead

tests/chrono_binding.cpp
juandent: I leave it open for extension that's why I put a SECTION()...

tests/statement_serializer_tests/schema/index.cpp
juandent: I did  not write this... ??

tests/statement_serializer_tests/schema/index.cpp
juandent: I did not write this test either!!

dev/date_binding.h
juandent: I chose as text because the you can see the date in a sql client... an integer would be like # of days since X for some X and it is not a meaningful representation

juandent: but I moved from unit test to example
and I modified CMakeLists.txt in examples folder so that you can build projects to compile the examples

fnc12: > he would have to provide his own implementation for sys_days binding...

fnc12: > he would have to provide his own implementation for sys_days binding...

No one cannot define specialization which is already defined

tests/statement_serializer_tests/schema/index.cpp
fnc12: r u sure? it exists in your diff

tests/statement_serializer_tests/schema/index.cpp
fnc12: it exists in your diff. It means you commit it

dev/date_binding.h
fnc12: > I chose as text because the you can see the date in a sql client... an integer would be like # of days since X for some X and it is not a meaningful representation

Sounds like a personal preferences. I prefer keeping data as integer in unix format cause it is comfortable to make operations on it. To see the date from unix time in SQL client you only need to add one function call. One more advantage: you don't need to keep date string format when you store date as integer

dev/date_binding.h
```diff
@@ -0,0 +1,127 @@
...
+
+// need support for #include <compare> for advanced std::chrono members
+#if defined(SQLITE_ORM_THREE_WAY_COMPARISON_SUPPORTED)
+
+#include <chrono>
+#include <sstream>
+#include <regex>
+#include <format>
+
+inline std::chrono::sys_days Today() {
```
fnc12, line 13: 1) this naming is unacceptable cause functions are named with camelCase inside lib
2) as I said before: please move this bindings from `dev` folder to example

dev/date_binding.h
fnc12: it is still here. IDK why you marked this conversation as resolved

fnc12 (changes requested):

examples/CMakeLists.txt
```diff
@@ -1,8 +1,63 @@
...
 
+# added by JDH
+set(sqlite_orm_examples_VERSION "1.8.0")
+set(PACKAGE_VERSION ${sqlite_orm_examples_VERSION})
+
+project("sqlite_orm_examples" VERSION ${PACKAGE_VERSION})
+
+# Handling C++ standard version to use
+option(SQLITE_ORM_ENABLE_CXX_20 "Enable C++ 20" OFF)
+option(SQLITE_ORM_ENABLE_CXX_17 "Enable C++ 17" OFF)
```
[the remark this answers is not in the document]
juandent: yes this is to be called like this from the command line:

`cmake  -G"Visual Studio 17 2022"   -Ax64  -DSQLITE_ORM_ENABLE_CXX_20=ON     -DSQLite3_INCLUDE_DIR=C:\Components\vcpkg\installed\x64-windows\include  -DSQLite3_LIBRARY=C:\Components\vcpkg\installed\x64-windows\lib\sqlite3.lib    -S . -B _binary2022plain_`

for instance for VS2022 and C++ 20

and for VS2015 it would be invoked like this:

`cmake  -G"Visual Studio 14 2015 Win64"   -DSQLite3_INCLUDE_DIR=C:\Components\vcpkg\installed\x64-windows\include  -DSQLite3_LIBRARY=C:\Components\vcpkg\installed\x64-windows\lib\sqlite3.lib    -S . -B _binary2015plain_`

so yes I think this way we can have the projects under one solution (that is Visual Studio solution) just like the unit tests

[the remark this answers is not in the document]
juandent: which is something that is lacking!!

dev/date_binding.h
juandent: ok no problem...

examples/CMakeLists.txt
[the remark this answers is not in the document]
fnc12: @trueqbit I need your review here cause I am not a pro in CMake

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#if defined(SQLITE_ORM_THREE_WAY_COMPARISON_SUPPORTED)
+
+#include "date_binding.h"
+
+using namespace sqlite_orm;
+using std::cout;
+using std::endl;
+
+// "sys_days_as_bindable_column_type"
```
fnc12, line 14: what does this comment mean?

examples/date_binding.h
```diff
@@ -0,0 +1,127 @@
...
+
+// need support for #include <compare> for advanced std::chrono members
+#if defined(SQLITE_ORM_THREE_WAY_COMPARISON_SUPPORTED)
+
+#include <chrono>
+#include <sstream>
+#include <regex>
+#include <format>
+
+inline std::chrono::sys_days today() {
```
fnc12, line 13: I'd move this code right inside example file (chrono_binding.h) just like it is done in `enum_binding.h` to make example self-sufficient

fnc12 (changes requested):

examples/date_binding.h
juandent: I understand but on the other hand it is good to have the binding in a separate header which is self-sufficient so people can just #include it if they want to have a basic implementation of sys_days binding.. What do you think about this?

examples/chrono_binding.cpp
juandent: I removed it

examples/date_binding.h
fnc12: > I understand but on the other hand it is good to have the binding in a separate header which is self-sufficient so people can just #include it if they want to have a basic implementation of sys_days binding.. What do you think about this?

People will not be able to include it cause it's not located in include dir. This header is located inside `tests` project which isn't available in public actually. And that's right cause example is not a thing to include

juandent: I have placed the header in examples folder - you are telling me that examples is not available to public too?
but its ok if it is not in the #include folder... it still can be copy paste the .h file to a folder of convenience to the developer
In this case the binding example IS a thing that can be included and we immediately have support for sys_days

juandent: the example is the CPP file

fnc12: > you are telling me that examples is not available to public too?

If you install the lib with any package manager then yes, you'll get only `include`(`dev`) contents, that's all.

>  it still can be copy paste the .h file to a folder of convenience to the developer

if this code is designed to be copied and pasted it can be copied and pasted from `.cpp` file the same way. I don't see any reason of having one example in two files

juandent: the reusable and self-sufficient is the date_binding.h file

fnc12: why do you think so? It has custom date to string converter. Different developers use different formats to serialize data to string. Some of them even serialize date to int as unix time (like me e.g.)

juandent: ok, not too important, it's okay will merge the 2 files

fnc12: thank you!

juandent: you are very welcome !!! I truly respect your opinions, BTW :-)

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,172 @@
...
+inline std::string sysDaysToString(std::chrono::sys_days pt) {
+    auto r = std::format("{:%F}", pt);
+    return r;
+}
+
+constexpr auto null_sys_day = std::chrono::sys_days{std::chrono::days{0}};
+
+/**
+ *  and sys_days from string. This function has nullable result cause
+ *  string can be neither `male` nor `female`. i have taken null_sys_day to
```
fnc12, line 38: what is `male` and `female`? Please make `i` capitalized

juandent: sorry... also changed response to std::optional<>

.gitignore
```diff
@@ -6,3 +6,14 @@ cmake-build-debug/
 .idea/
 
 /compile
+
+_binary2022_/
```
trueqbit, line 10: You will remove them from .gitignore before the final merge, right?

CMakeLists.txt
```diff
@@ -47,7 +47,8 @@ target_link_libraries(sqlite_orm INTERFACE SQLite::SQLite3)
 
 target_sources(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include/sqlite_orm/sqlite_orm.h>)
 
-target_include_directories(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
+###target_include_directories(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
+target_include_directories(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/not_single_header_include>)
```
trueqbit, line 51: Please remember to change it back again before the final merge.

examples/date_binding.h
```diff
@@ -0,0 +1,127 @@
...
+                throw std::runtime_error("incorrect date string (" + std::string(row_value) + ")");
+            }
+        }
+
+        std::chrono::sys_days extract(sqlite3_stmt* stmt, int columnIndex) const {
+            auto str = sqlite3_column_text(stmt, columnIndex);
+            return this->extract((const char*)str);
+        }
+        std::chrono::sys_days extract(sqlite3_value* row_value) const {
+            auto characters = reinterpret_cast<const char*>(sqlite3_value_text(row_value));
```
trueqbit, line 121: Rather use a C cast here just like above and all other places concerning row_extractor.

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,169 @@
...
+            if(sd) {
+                return sd.value();
+            } else {
+                throw std::runtime_error("incorrect date string (" + std::string(row_value) + ")");
+            }
+        }
+
+        std::chrono::sys_days extract(sqlite3_stmt* stmt, int columnIndex) const {
+            auto str = sqlite3_column_text(stmt, columnIndex);
+            return this->extract((const char*)str);
```
trueqbit, line 120: Beware: `sqlite3_column_text` might return null.

examples/date_binding.h
```diff
@@ -0,0 +1,127 @@
...
+            if(sd != null_sys_day) {
+                return sd;
+            } else {
+                throw std::runtime_error("incorrect date string (" + std::string(row_value) + ")");
+            }
+        }
+
+        std::chrono::sys_days extract(sqlite3_stmt* stmt, int columnIndex) const {
+            auto str = sqlite3_column_text(stmt, columnIndex);
+            return this->extract((const char*)str);
```
trueqbit, line 118: Beware: sqlite3_column_text might return null.

trueqbit (changes requested):

.gitignore
juandent: > You will remove them from .gitignore before the final merge, right?


ok I will

CMakeLists.txt
juandent: why? Its so useful here! During development

examples/chrono_binding.cpp
juandent: then, is this function a correct replacement?

```C++
std::chrono::sys_days extract(sqlite3_stmt* stmt, int columnIndex) const {
            auto str = sqlite3_column_text(stmt, columnIndex);
            if(str == nullptr)  str = "";
            return this->extract((const char*)str);
```

.gitignore
trueqbit: > > are you 100% sure that it is necessary to turn the examples subfolder into a cmake project?
> 
> Its convenient!! what's wrong with having a project for the examples??

I think you placed your comment at the wrong conversation :)

examples/date_binding.h
juandent: > sqlite3_column_text might return null.

When it is null, should we make str == "" ?

juandent: in which cases is this a null?

trueqbit: For a statement where SQLite returns null? Or am I wrong here?

juandent: taken care as we discussed!! Thanks @trueqbit

examples/chrono_binding.cpp
juandent: no - forget this one!!

examples/date_binding.h
juandent: why a C cast? MSVC does not allow a static_cast<> because the input is unsigned and the output signed that's why I had to use reinterpret_cast<> -- I always thought C++ casts should be preferred over C casts, am I wrong?

trueqbit: 3 lines up, you have a C cast, just like all the other line extractors use. sqlite_orm uses C casts or value casts throughout, so I would just be consistent.
As for preferring reinterpret_cast over C casts - I'm not in favor of always using reinterpret_cast (and often not even static_cast) simply because it's less readable. And a C cast is perfectly valid. Just my 2 cents :)

examples/chrono_binding.cpp
fnc12: right but please skip `if` operator without curly braces. It is a rule from the code style

juandent:
```C++
template<>
    struct row_extractor<std::chrono::sys_days> {
        std::chrono::sys_days extract(const char* row_value) const {
            if(row_value) {
                auto sd = sysDaysFromString(row_value);
                if(sd) {
                    return sd.value();
                } else {
                    throw std::runtime_error("incorrect date string (" + std::string(row_value) + ")");
                }
            } else {
                // ! row_value
                throw std::runtime_error("incorrect date string (nullptr)");
            }
        }
```

its done!

CMakeLists.txt
juandent: ok I changed it!

.gitignore
juandent: No actually there was no way to answer to your post regarding examples/CMakeLists.txt so I had to answer here!!
Why is that?

image

examples/chrono_binding.cpp
juandent: > BTW extract throws exception is argument is null
yes what should it return?

juandent: ok you hadn't seen this
we are ok now...

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <cassert>
+#include <string>
+#include <iostream>
+
+#if __cpp_lib_chrono >= 201907L
+
+///////////////////////////////
+/// sys_days binding as TEXT
+/// ///////////////////////////
+#if __cpp_lib_chrono >= 201907L
```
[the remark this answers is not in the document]
juandent: ok I have eliminated one of the #ifs --- it was to make the code for sys_days binding more self sufficient

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <format>
+
+inline std::chrono::sys_days today() {
+    const auto today = std::chrono::sys_days{floor<std::chrono::days>(std::chrono::system_clock::now())};
+    return today;
+}
+
+/**
+ *  This is the date we want to map to our sqlite db.
+  *  Let's make it `TEXT`
```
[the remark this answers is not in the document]
juandent: ok done

fnc12 (approved):

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,173 @@
+#include <sqlite_orm/sqlite_orm.h>
+#include <cassert>
+#include <string>
+#include <iostream>
+
+#if __cpp_lib_chrono >= 201907L
```
trueqbit, line 6: As you are also using Better test for `__cpp_lib_format`, too. You never know what the compiler really implements.
Additionally you need to first include the C++ header, then test for the library feature!

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+///////////////////////////////
+/// sys_days binding as TEXT
+/// ///////////////////////////
+
+#include <chrono>
```
trueqbit, line 12: Please include before the library feature test above.

examples/date_binding.h
trueqbit: This header file still exists?

fnc12: no. Just a comment

.gitignore
```diff
@@ -6,3 +6,4 @@ cmake-build-debug/
 .idea/
 
 /compile
+
```
trueqbit, line 9: Now you have introduced an extra newline...

examples/chrono_binding.cpp
juandent: done

.gitignore
juandent: done

tests/chrono_binding.cpp
```diff
@@ -0,0 +1,41 @@
+#include <sqlite_orm/sqlite_orm.h>
+#include <catch2/catch.hpp>
+
+using namespace sqlite_orm;
+
+#if defined(SQLITE_ORM_THREE_WAY_COMPARISON_SUPPORTED)
```
trueqbit, line 6: Need to test for `#if __cpp_lib_chrono >= 201907L && __cpp_lib_format >= 202110L` here, too.

juandent: but  I did already!

trueqbit: not in tests/chrono_binding.cpp. I think the diff is clearer than our conversation ;)

tests/statement_serializer_tests/schema/index.cpp
trueqbit: The diff doesn't lie, obviously.

tests/statement_serializer_tests/schema/index.cpp
trueqbit: The diff doesn't lie, obviously.

examples/chrono_binding.cpp
juandent: done

tests/chrono_binding.cpp
juandent: where are you seeing this?

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,173 @@
...
+    return today;
+}
+
+/**
+ *  This is the date we want to map to our sqlite db.
+ *  Let's make it `TEXT`
+ */
+
+//  also we need transform functions to make string from enum..
+inline std::string sysDaysToString(std::chrono::sys_days pt) {
```
trueqbit, line 28: I know you copied the function from the inital header file. However this function should be static instead of inline in the .cpp file.

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,173 @@
...
+/**
+ *  and sys_days from string. This function has nullable result cause
+ *  string can be an invalid sys_days. 
+ *  Of course we won't allow this to happen but as developers we must have
+ *  a scenario for this case.
+ *  These functions are just helpers. They will be called from several places
+ *  that's why I placed it separatedly. You can use any transformation type/form
+ *  (for example BETTER_ENUM https://github.com/aantron/better-enums)
+ */
+inline std::optional<std::chrono::sys_days> sysDaysFromString(const std::string& s) {
```
trueqbit, line 42: I know you copied the function from the inital header file. However this function should be static instead of inline in the .cpp file.

examples/chrono_binding.cpp
```diff
@@ -0,0 +1,173 @@
+#include <sqlite_orm/sqlite_orm.h>
+#include <cassert>
+#include <string>
+#include <iostream>
+#include <chrono>
+#include <format>
+
+#if __cpp_lib_chrono >= 201907L && __cpp_lib_format >= 202110L
```
trueqbit, line 8: Do you need to test `__cpp_lib_format` against 202110? I.e. does your example depend on locale handling in chrono formatters and non-const-formattable types? Or can it just be the text formatting initially introduced in 201907?

tests/chrono_binding.cpp
juandent: ok found'

examples/chrono_binding.cpp
juandent: why static? so it does not appear viewable from another translation unit?

juandent: done!

examples/chrono_binding.cpp
juandent: correct 201907 is enough

examples/chrono_binding.cpp
juandent: done

tests/chrono_binding.cpp
juandent: this file should be deleted!

trueqbit (approved):

fnc12 (approved):

fnc12:
thanks @juandent

PR #1063, opened 2022-07-08
Corrected storage lookup in regard to non-unique DB objects

trueqbit:
I assumed that 'DBOs' in a storage definition are of unique type, but this isn't the case for indexes created on a table using columns that have the same type.


tests/storage_tests.cpp
```diff
@@ -280,3 +280,41 @@ TEST_CASE("issue880") {
     Inversion inversion;
     storage.has_dependent_rows(inversion);
 }
+
+namespace {
+    class Record final {
```
fnc12, line 285: nit: one can place `Record` right inside the test in case other test will also have a different `Record` class

fnc12 (approved):

fnc12:
thanks @trueqbit

PR #1064, opened 2022-07-10
added on conflict for primary key

fnc12:


dev/constraints.h
```diff
@@ -465,6 +504,10 @@ namespace sqlite_orm {
         return {{}};
     }
 
+    /**
+     *  AUTOINCREMENT keyword. [Deprecation notice] Use `primary_key().autoincrement()` instead of using this function.
```
trueqbit, line 508: The standard attribute `[[deprecated]]` is available.

include/sqlite_orm/sqlite_orm.h
```diff
@@ -15704,18 +15745,55 @@ namespace sqlite_orm {
...
+                    case conflict_clause_t::abort:
+                        return "ABORT";
+                    case conflict_clause_t::fail:
+                        return "FAIL";
+                    case conflict_clause_t::ignore:
+                        return "IGNORE";
+                    case conflict_clause_t::replace:
+                        return "REPLACE";
+                }
+            }
```
trueqbit, line 15766: Please add a `return {};` in order to avoid warning "C4715 : not all control paths return a value".

trueqbit (changes requested):
Requesting changes.

dev/constraints.h
trueqbit, line 508: Standard attribute `[[deprecated]]` is available!

dev/statement_serializer.h
```diff
@@ -783,18 +783,55 @@ namespace sqlite_orm {
...
+                    case conflict_clause_t::abort:
+                        return "ABORT";
+                    case conflict_clause_t::fail:
+                        return "FAIL";
+                    case conflict_clause_t::ignore:
+                        return "IGNORE";
+                    case conflict_clause_t::replace:
+                        return "REPLACE";
+                }
+            }
```
trueqbit, line 804: Please add a `return {};` in order to avoid msvc warning "C4715: not all control paths return a value".

dev/constraints.h
trueqbit: Same would be nice on `having()`.

trueqbit (changes requested):

dev/statement_serializer.h
fnc12: done

dev/constraints.h
fnc12: done

trueqbit (approved):

PR #1126, opened 2023-01-28
Daisy chaining the concatenation operator

trueqbit:
It would be nice to be able to daisy-chain the concatenation operator, i.e. to use it several times in a row so that
```sql
select(c("first name") || " " || "last name");
```
works.


fnc12 (approved):

fnc12:
thanks @trueqbit

PR #1127, opened 2023-01-28
Runner-up: Daisy chaining the concatenation operator

trueqbit:
Updated example that daisy-chains the concatenation operator.


fnc12:
thanks @trueqbit

PR #1129, opened 2023-01-30
`EXISTS` must use parentheses

trueqbit:
To be independent of the current context, the serializer for `exist_t` should always parenthesize its subselect.


fnc12 (approved):

fnc12:
thanks @trueqbit

PR #1130, opened 2023-01-31
Fix warning of std::move

shuai132:
image

PR #1135, opened 2023-02-02
Qualifying `std::move()` 2

trueqbit:
Runner-up for shuai132's PR #1130, which qualifies the call to `std::move()` for all remaining occurrences.

### DIFF
--- a/dev/ast/set.h
+++ b/dev/ast/set.h
@@ -51,11 +51,11 @@ namespace sqlite_orm {
             }
 
             dynamic_set_t(dynamic_set_t&& other) :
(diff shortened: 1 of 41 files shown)


fnc12:
how it turned that ADL usage started producing warnings?

trueqbit:
> how it turned that ADL usage started producing warnings?

I was surprised too, but it seems to come from the WG21 recommendation, probably in conjunction with `views::move()`. Here is a [discussion ](https://reviews.llvm.org/D119670) about this new warning issued by the Clang compiler.

fnc12 (approved):

PR #1144, opened 2023-02-16
Working with aliases is a breeze

trueqbit:
C++20 allows us to introduce syntactic sugar, which I think is a huge improvement for readability and the natural use of table and column aliases, even at a conceptual level.

To see it in action, take a look:
[self_join.cpp](/FireDaemon/sqlite_orm/blob/2a1410e8a92ba5ae01de9a59822b17950ff185c4/examples/self_join.cpp#L202)
[subquery.cpp](/FireDaemon/sqlite_orm/blob/2a1410e8a92ba5ae01de9a59822b17950ff185c4/examples/subquery.cpp#L1349)
[exists.cpp](/FireDaemon/sqlite_orm/blob/2a1410e8a92ba5ae01de9a59822b17950ff185c4/examples/exists.cpp#L478)

This PR contains:
* Dedicated unit tests for aliases
* Concepts and type traits for aliases
* Syntactic sugar for easier creation and usage of aliased expressions (pointer-member-pointer operator, ad-hoc string aliases)
* Column aliases are usable on their own in expressions

Some codified updates:
* `as<colalias_i>(&User::id)` =>
  * `as<"i"_col>(&User::id)` or
  * `&User::id >>= "i"_col`
* `using als = alias_e<Employee>` =>
  * `constexpr auto e = "e"_alias.for_<Employee>()`
* `alias_column<als>(&Employee::id)` =>
  * `alias_column<e>(&Employee::id)` or
  * `e->*&Employee::id`

This of course works seamlessly with other functions like `from<e>()`, `asterisk<e>()`, `inner_join<e>()`, ...

Last but not least, the C++20-based way of working with aliases lays the foundation for the use of CTEs, which also reduces the changeset.


examples/custom_aliases.cpp
```diff
@@ -104,6 +104,35 @@ int main(int, char** argv) {
...
+        columns(c_als->*&Employee::id, c_als->*&Employee::name, c_als->*&Employee::age, d->*&Department::dept),
+        where(is_equal(c_als->*&Employee::id, d->*&Department::empId)));
+    assert(rowsWithTableAliases == simpleRows);
+
+    //  SELECT COMPANY.ID as COMPANY_ID, COMPANY.NAME AS COMPANY_NAME, COMPANY.AGE, DEPARTMENT.DEPT
+    //  FROM COMPANY, DEPARTMENT
+    //  WHERE COMPANY_ID = DEPARTMENT.EMP_ID;
+    auto rowsWithColumnAliases = storage.select(
+        columns(&Employee::id >>= empId, as<CompanyNameAlias>(&Employee::name), &Employee::age, &Department::dept),
+        where(is_equal(get<empId>(), &Department::empId)));
```
fnc12, line 122: question: can we replace `is_equal(get<empId>(), &Department::empId)` with `get<empId>() == &Department::empId`?

examples/custom_aliases.cpp
```diff
@@ -104,6 +104,35 @@ int main(int, char** argv) {
...
+    assert(rowsWithColumnAliases == rowsWithTableAliases);
+
+    //  SELECT C.ID AS COMPANY_ID, C.NAME AS COMPANY_NAME, C.AGE, D.DEPT
+    //  FROM COMPANY AS C, DEPARTMENT AS D
+    //  WHERE  C.ID = D.EMP_ID;
+    auto rowsWithBothTableAndColumnAliases =
+        storage.select(columns(c_als->*& Employee::id >>= empId,
```
fnc12, line 129: you have a redundant space right after `*&`

tests/operators/arithmetic_operators.cpp
```diff
@@ -35,11 +35,11 @@ TEST_CASE("Arithmetic operators") {
         }
     }
     {  //  +
-        auto rows = storage.select(columns(c(&Object::nameLen) + 1000));
+        auto rows = storage.select(c(&Object::nameLen) + 1000);
```
fnc12, line 38: after this change current test looks the same like the previous one. This test with a `column` with a single argument was created intentionally to cover corner case. Better bring it back

tests/statement_serializer_tests/select_constraints.cpp
```diff
@@ -71,6 +71,20 @@ TEST_CASE("statement_serializer select constraints") {
             value = serialize(expression, context);
             expected = R"(FROM "users" "u")";
         }
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+        {
+            SECTION("with alias 2") {
+                auto expression = from<alias_<'u'>.for_<User>()>();
```
fnc12, line 77: why don't we use `alias<'u'>` instead of `alias_<'u'>`?

appveyor.yml
```diff
@@ -25,12 +25,15 @@ environment:
       CC: clang
       CXX: clang++
       SQLITE_ORM_CXX_STANDARD: "-DSQLITE_ORM_ENABLE_CXX_14=ON"
+      cmake_build_parallel: --parallel
```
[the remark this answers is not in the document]
trueqbit: Unfortunately not with the standard appveyor subscription.
Might keep it anyway?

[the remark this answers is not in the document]
fnc12: yeah why not

dev/alias.h
```diff
@@ -40,20 +63,39 @@ namespace sqlite_orm {
             column_type column;
         };
 
+        /*
+         * Encapsulates extracting the alias identifier of a non-alias.
+         */
         template<class T, class SFINAE = void>
         struct alias_extractor {
-            static std::string get() {
+            static std::string extract() {
```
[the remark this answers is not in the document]
trueqbit: Both are used.

If you are asking this in the context of `extract()` + `as_alias()`:

1. I just don't like `get()`. Everything's called 'get' these days and one doesn't recognize anymore what it does. So I figured a better name is 'extract'.
2. `extract()` returns the name unconditionally for columns - when collecting serialized column expressions ('column names getter') and when introducing an alias for them (`as<>()`). `as_alias()` is used for aliasing and referring to aliased tables.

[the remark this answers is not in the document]
fnc12: > I just don't like get(). Everything's called 'get' these days and one doesn't recognize anymore what it does. So I figured a better name is 'extract'.

That is good!

> extract() returns the name unconditionally for columns - when collecting serialized column expressions ('column names getter') and when introducing an alias for them (as<>()). as_alias() is used for aliasing and referring to aliased tables.

Got it. Thank you!

dev/alias.h
```diff
@@ -70,42 +112,100 @@ namespace sqlite_orm {
...
+    template<orm_column_alias auto als, class E>
+    auto as(E expression) {
+        return internal::as_t<std::remove_const_t<decltype(als)>, E>{std::move(expression)};
+    }
+
+    /** 
+     *  Alias a column expression.
+     */
+    template<orm_column_alias A, class E>
+    internal::as_t<A, E> operator>>=(E expression, const A&) {
```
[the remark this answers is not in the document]
trueqbit: I was just looking for a natural 'as' operator. But I don't want to exaggerate, so you'd better stop me if you're not sure.

dev/type_traits.h
```diff
@@ -58,4 +61,9 @@ namespace sqlite_orm {
         template<typename T>
         using on_type_t = typename T::on_type;
     }
+
+#ifdef SQLITE_ORM_CPP20_CONCEPTS_SUPPORTED
+    template<class T>
+    concept orm_names_type = requires { typename T::type; };
```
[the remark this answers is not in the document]
trueqbit: I think it should remain in the public namespace because it is part of the public requirements [`orm_column_alias`](https://github.com/FireDaemon/sqlite_orm/blob/d87938ef9b48204465030466b10955bd1b7dbb7f/dev/alias_traits.h#L66) and [`orm_recordset_alias`](https://github.com/FireDaemon/sqlite_orm/blob/d87938ef9b48204465030466b10955bd1b7dbb7f/dev/alias_traits.h#L75).

dev/alias.h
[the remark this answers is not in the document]
fnc12: I don't want to stop something new. Let's leave it as is. It looks nice. Yes it is unusual but everything seemed unusual initially. Anyway we can remove it before the next release in case of something happens

examples/column_aliases.cpp
```diff
@@ -58,12 +58,26 @@ void marvel_hero_ordered_by_o_pos() {
         }
     }
     cout << endl;
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+    {
+        //  SELECT name, instr(abilities, 'o') i
+        //  FROM marvel
+        //  WHERE i > 0
+        //  ORDER BY i
+        auto rows = storage.select(columns(&MarvelHero::name, as<"i"_col>(instr(&MarvelHero::abilities, "o"))),
```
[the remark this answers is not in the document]
trueqbit: You are right, yes of course.

examples/custom_aliases.cpp
trueqbit: It can be rewritten as `get<empId>() == c(&Department::empId)`.

examples/custom_aliases.cpp
trueqbit: That's the work of the stupid clang formatter.

tests/operators/arithmetic_operators.cpp
trueqbit: Sorry about that!

tests/statement_serializer_tests/select_constraints.cpp
trueqbit: Because I thought it resembles the current predefined table aliases. `alias_<'u'>` resembles `alias_u`.
However, the question is whether we want to have this syntax in the library at all.

The other available option is to use `"u"_alias`:
1. `constexpr auto u = "u"_alias.for_<User>()`
2. `constexpr auto u = alias_<'u'>.for_<User>()` - yes/no?

examples/custom_aliases.cpp
fnc12: ok not our business =)

tests/statement_serializer_tests/select_constraints.cpp
fnc12: `constexpr auto u = "u"_alias.for_<User>()` - nice
`constexpr auto u = alias<'u'>.for_<User>()` also but without redundant underscore as I wrote it. Underscore is a service symbols and shouldn't be used if it is not required. If you see that `alias_<'u'>` replaces `alias_u` then you can see it like underscore is replaced with templates brackets

dev/alias.h
```diff
@@ -70,42 +112,100 @@ namespace sqlite_orm {
...
             using type = T;
+
+            alias_holder() = default;
         };
+
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+        template<char A, char... C>
+        struct table_alias_builder {
+            static_assert(sizeof...(C) == 0 && ((A >= 'A' && 'Z' <= A) || (A >= 'a' && 'z' <= A)),
+                          "Table alias identifiers shall consist of a single alphabetic character, in order to evade "
```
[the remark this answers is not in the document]
trueqbit: Well, I derived it from the existing built-in table (and column) aliasing, which has been artificially limited to one character as well.

There's the other way around, where I currently require CTE aliases to [consist of 2 characters or more](https://github.com/FireDaemon/sqlite_orm/blob/f3d9ff985e44f855b531317f37e9c8ff312a331a/dev/cte.h#L39).

As I had to [lift that limitation](https://github.com/FireDaemon/sqlite_orm/blob/d87938ef9b48204465030466b10955bd1b7dbb7f/dev/alias.h#L46) anyway, I agree that it should be up to the programmer, and shouldn't be an issue when properly defining aliases.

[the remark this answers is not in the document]
fnc12: > Well, I derived it from the existing built-in table (and column) aliasing, which has been artificially limited to one character as well.

It is not limited cause it doesn't limit creating new aliases with label at any length. It is implemented for single chars cause implementing it for every character combination is unreal. But `static_assert` you placed is a real limitation. That's the case I am talking about.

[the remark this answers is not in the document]
trueqbit: [It was limited](https://github.com/fnc12/sqlite_orm/blob/d42981421c0121fe2c87d915fac2190c4dd909c8/dev/alias.h#L23).
`template<class T, char A> struct table_alias;`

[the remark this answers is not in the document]
fnc12: no no no, I understand but the approach is not limited. Here I see that you limit the approach. Maybe I am wrong but it looks like it is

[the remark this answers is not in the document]
trueqbit: Well, it probably depends on from the angle you look at it. I think I understand what you mean. One isn't supposed to use `internal::table_alias<>` directly, so you can only meaningfully define public types having one character. But limiting `""_alias` to a single character is a real artificial handicap.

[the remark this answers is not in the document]
fnc12: I'd allow user to write the code like "temp"_alias cause it can be written with raw SQL on purpose

dev/conditions.h
```diff
@@ -1015,21 +1028,49 @@ namespace sqlite_orm {
         return {std::move(o)};
     }
 
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+    template<orm_recordset_alias auto als, class O>
+    auto left_join(O o) {
```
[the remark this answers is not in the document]
trueqbit: It's now declared as
```c++
template<orm_recordset_alias auto alias, class On>
auto left_join(On on);
```

appveyor.yml
```diff
@@ -25,16 +25,19 @@ environment:
...
       CXX: g++
       SQLITE_ORM_CXX_STANDARD: "-DSQLITE_ORM_ENABLE_CXX_14=ON"
+      # gcc was stuck with a parallel build
+      cmake_build_parallel: ""
 
     # Representative for C++14
-    - job_name: Visual Studio 2022, x64, C++14
-      appveyor_build_worker_image: Visual Studio 2022
+    - job_name: Visual Studio 2019, x64, C++14
+      appveyor_build_worker_image: Visual Studio 2019
```
trueqbit, line 40: @fnc12 There's a little stupid problem with unit tests on Windows.

The Visual Studio 2019 build worker image only has CMake 3.16.
With this PR I introduced a dependency on CMake 3.24, because of the `FIND_PACKAGE_ARGS` feature (which prefers an already installed Catch2 library over building it from source, which takes quite some build time).

Contrary to Visual C++ 2019, Visual C++ 2017 has still broken variadic template pack expansion, which we treat as a compiler quirk in sqlite_orm.
But of course it would be useful to have the representative C++14 test on Windows run with VC++ 2017.

dev/conditions.h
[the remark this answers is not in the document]
fnc12: BTW there is no rule for template identifier name length. So you can use it with any length. Just please make it capitalized please as it is done everywhere in this project

trueqbit:
Also in this PR:

- CMake BUILD_EXAMPLES option corrected.
- Corrected handling of CMake dependencies
  - especially up to CMake 3.24, dependencies cannot be made globally available from a subdirectory
  - CMake now prefers an already installed Catch2 v3 library instead of building it from source code
- The Appveyor build using the Clang C++20 environment now also builds SQLiteORM examples

@fnc12 It's been a bit of a journey to update the CMake thing. I can open a new PR for a clean commit history if you like.

fnc12:
@trueqbit thank you for such a great job! No need to refactor commits. I am a fan or original commit history cause it shows how every goes in fact

dev/column_result.h
```diff
@@ -24,6 +24,8 @@ namespace sqlite_orm {
     namespace internal {
 
         /**
+         *  Obains the result type of expressions that form the columns of a select statement.
```
fnc12, line 27: typo: `Obtains` instead of `Obains`

dev/conditions.h
```diff
@@ -1015,21 +1035,49 @@ namespace sqlite_orm {
         return {std::move(o)};
     }
 
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+    template<orm_recordset_alias auto alias, class On>
```
fnc12, line 1039: nit for the future: there is no requirement of making template names longer than one symbols, only argument/variable/constant names

dev/serializer_context.h
```diff
@@ -19,6 +19,9 @@ namespace sqlite_orm {
             serializer_context(const db_objects_type& dbObjects) : db_objects{dbObjects} {}
         };
 
+        //template<class DBOs>
+        //serializer_context(const DBOs&) -> serializer_context<DBOs>;
```
fnc12, line 23: why did you leave a comment? we'd better skip commented out code cause it is useless during code reading and we can use version control system if we want to keep some unused code

fnc12 (approved):

dev/conditions.h
trueqbit: Understood. Though TBH I find it also nicer to read when template variables are spelled out, too.

dev/serializer_context.h
trueqbit: Thanks for catching this as well!

dev/conditions.h
fnc12: agree. It is up to you

fnc12 (approved):

PR #1143, opened 2023-02-13
Ability to select all columns of multiple tables

trueqbit:
Addresses #1106.

Occasionally it can be useful to select all columns of several tables, and for CTEs this could be a nice improvement:
```c++
select(columns(asterisk<X>(), asterisk<Y>()), ...);
```

Additionally, this PR contains:
* Streamlined collection of serialized column expressions
* Column expression collector doesn't modify serialization context anymore
* Serializing `select_t` always sets the context such that it qualifies table columns
* RIP `join_iterator`

PR #1153, opened 2023-03-13
Tests for constrained join clauses with regard to table collection

trueqbit:
I thought I have broken the serialization of cross and natural joins with commit 34afdf3005172761bd3b9f43c9c8669ff231d578 (RIP `join_iterator`) that was part of PR #1143.

This PR introduces dedicated unit tests for serializing table joins when not using an explicit `from<>()`, showing that fortunately I didn't break anything.

### DIFF
--- a/tests/statement_serializer_tests/select_constraints.cpp
+++ b/tests/statement_serializer_tests/select_constraints.cpp
@@ -86,6 +86,51 @@ TEST_CASE("statement_serializer select constraints") {
         }
 #endif
     }
+    // tests whether the statement serializer for a select with joins
+    // properly deduplicates the table names when no explicit from is used
+    SECTION("deduplicated table names") {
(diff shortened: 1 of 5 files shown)


tests/statement_serializer_tests/select_constraints.cpp
```diff
@@ -86,6 +86,51 @@ TEST_CASE("statement_serializer select constraints") {
         }
 #endif
     }
+    // tests whether the statement serializer for a select with joins
+    // properly deduplicates the table names when no explicit from is used
+    SECTION("deduplicated table names") {
```
fnc12, line 91: good job! There is no shame in unit tests!

fnc12 (approved):

fnc12 (approved):

PR #1159, opened 2023-03-18
Operators should be in internal namespace

### LINKED ISSUE #1047: Operators should be in internal namespace
Operators defined [here](https://github.com/fnc12/sqlite_orm/blob/fdcc1da46fbd90feb886c0588462a62d29eb5a06/include/sqlite_orm/sqlite_orm.h#L3344) should be in the `internal` namespace (the same one as `expression`) so they can be found by ADL.
Or in the global namespace if you prefer.

Goldbolt if you want to play with it: https://godbolt.org/z/9ofP3rvd4.

Thanks for the awesome library !

trueqbit:
Fixes #1047.

Deliberately put operators for `expression_t` into the internal namespace to facilitate ADL (Argument Dependent Lookup).


tests/operators/adl.cpp
```diff
@@ -0,0 +1,64 @@
...
+using sqlite_orm::internal::lesser_or_equal_t;
+using sqlite_orm::internal::lesser_than_t;
+using sqlite_orm::polyfill::is_specialization_of_v;
+
+TEST_CASE("ADL and expression operators") {
+    struct User {
+        int id;
+    };
+
+    STATIC_REQUIRE(is_specialization_of_v<decltype(c(&User::id) < 42), lesser_than_t>);
```
fnc12, line 19: this test file better be moved to `static_tests` folder

fnc12 (approved):

Issue #1160, opened 2023-03-19
Compilation time analysis via ClangBuildAnalyzer


vittorioromeo:
Trying to speed up compilation times a bit. The first step is an analysis using [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer), which I ran over the entire repository, with tests and examples enabled.

ClangBuildAnalyzer output here:
https://gist.github.com/vittorioromeo/ab81ce00927149737398f97d3948b62e

Most of the time is spent in these instantiations, would be nice to try and optimize those:

```
**** Template sets that took longest to instantiate:
 71289 ms: sqlite_orm::internal::iterate_tuple<$> (2459 times, avg 28 ms)
 60327 ms: sqlite_orm::internal::serialize<$> (1782 times, avg 33 ms)
 51614 ms: sqlite_orm::internal::storage_t<$>::sync_schema (162 times, avg 318 ms)
 46721 ms: sqlite_orm::internal::storage_t<$>::sync_table<$> (270 times, avg 173 ms)
 36778 ms: sqlite_orm::internal::storage_t<$>::prepare<$> (797 times, avg 46 ms)
 36467 ms: sqlite_orm::internal::storage_t<$>::prepare_impl<$> (797 times, avg 45 ms)
 24957 ms: sqlite_orm::internal::storage_t<$>::create_table<$> (246 times, avg 101 ms)
 24890 ms: sqlite_orm::internal::storage_t<$>::select<$> (346 times, avg 71 ms)
 24793 ms: std::forward_as_tuple<$> (1914 times, avg 12 ms)
 21827 ms: std::tuple<$> (3978 times, avg 5 ms)
 18537 ms: std::__and_<$> (10567 times, avg 1 ms)
 11696 ms: std::_TupleConstraints<$>::__is_implicitly_constructible<$> (5385 times, avg 2 ms)
 10446 ms: sqlite_orm::internal::table_t<$>::for_each_column<$> (972 times, avg 10 ms)
```

trueqbit:
Thank you, @vittorioromeo, any investigation, insights, ideas, and tests are appreciated!

Additionally to the [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer), is the `chrome://tracing` tool helpful as well?

[...]

PR #1161, opened 2023-03-19
Simplify reverse tuple iteration

vittorioromeo:
Should help a bit with #1160. Would also be nice to see how it interacts with #1055

### DIFF
--- a/dev/functional/index_sequence_util.h
+++ b/dev/functional/index_sequence_util.h
@@ -3,9 +3,6 @@
 #include <utility>  //  std::index_sequence, std::make_index_sequence
 
 #include "../functional/cxx_universal.h"
-#ifdef SQLITE_ORM_RELAXED_CONSTEXPR_SUPPORTED
-#include <array>
-#endif
 
 namespace sqlite_orm {
     namespace internal {
--- a/dev/functional/index_sequence_util.h
+++ b/dev/functional/index_sequence_util.h
@@ -17,37 +17,6 @@ namespace sqlite_orm {
             return I;
         }
 
-#ifdef SQLITE_ORM_RELAXED_CONSTEXPR_SUPPORTED
-        /**
-         *  Reorder the values of an index_sequence according to the positions from a second sequence.
-         */
-        template<size_t... Value, size_t... IdxOfValue>
-        SQLITE_ORM_CONSTEVAL auto reorder_index_sequence(std::index_sequence<Value...>,
-                                                         std::index_sequence<IdxOfValue...>) {
-            constexpr std::array<size_t, sizeof...(Value)> values{Value...};
-            return std::index_sequence<values[sizeof...(Value) - 1u - IdxOfValue]...>{};
-        }
-
-        template<size_t Value, size_t IdxOfValue>
-        SQLITE_ORM_CONSTEVAL std::index_sequence<Value> reorder_index_sequence(std::index_sequence<Value>,
-                                                                               std::index_sequence<IdxOfValue>) {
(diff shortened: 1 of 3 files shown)


fnc12:
@vittorioromeo thank you for this PR. Please resolve the conflicts to be able to run CI tests

vittorioromeo:
> @vittorioromeo thank you for this PR. Please resolve the conflicts to be able to run CI tests

Should be fine now :)

fnc12:
@vittorioromeo did you test compilation time with your branch? I am actually not sure whether reverse tuple iteration is used anywhere so I am not sure whether your fix will make much effort

vittorioromeo:
> @vittorioromeo did you test compilation time with your branch? I am actually not sure whether reverse tuple iteration is used anywhere so I am not sure whether your fix will make much effort

I saw a very small improvement of average 24ms per instantiation of `iterate_tuple` from 28ms, but hard to tell if it's impactful in any way. There are a few calls to `iterate_tuple<true>` in the codebase. Regardless, I think that the PR is an improvement overall because of the removal of unnecessary code.

dev/functional/index_sequence_util.h
trueqbit, line 1: Could you please also remove
```c++
#ifdef SQLITE_ORM_RELAXED_CONSTEXPR_SUPPORTED
#include <array>
#endif
```
from the top of the header?

dev/tuple_helper/tuple_iteration.h
```diff
@@ -33,7 +33,8 @@ namespace sqlite_orm {
         template<bool reversed = false, class Tpl, size_t... Idx, class L>
         void iterate_tuple(const Tpl& tpl, std::index_sequence<Idx...>, L&& lambda) {
             if constexpr(reversed) {
-                iterate_tuple(tpl, reverse_index_sequence(std::index_sequence<Idx...>{}), std::forward<L>(lambda));
+                int sink;
```
trueqbit, line 36: Please put a comment above similar to:
`// nifty fold expression trick: make use of guaranteed right-to-left evaluation order when folding over operator=`.

Even seasoned C++ programmers will find it helpful :) - Even I had to try it out and find out about it.

trueqbit:
This is awesome C++! Thanks @vittorioromeo for thinking along.

I doubt you will see any overall improvement in compilation times.
But it simplifies the code, and that was what I was originally looking for when I updated the tuple iteration but couldn't find it!

By the way, there is good reading about this fold expression trickery:
* https://www.foonathan.net/2020/05/fold-tricks
* https://quuxplusone.github.io/blog/2020/05/07/assignment-operator-fold-expression/

trueqbit (changes requested):

dev/functional/index_sequence_util.h
trueqbit, line 1: BTW, we might implement `first_index_sequence_value()` as follows. What do you think?

```c++
template<size_t... Idx>
SQLITE_ORM_CONSTEVAL size_t first_index_sequence_value(std::index_sequence<Idx...>) {
    static_assert(sizeof...(Idx) > 0);
    size_t result;
    ((result = Idx, true) || ...);
    return result;
}
```

vittorioromeo: Compared to the current implementation, I'm not sure it's going to be faster. I would recommend a benchmark on https://www.build-bench.com/

trueqbit (approved):

fnc12 (approved):

fnc12:
@vittorioromeo thanks!

PR #1171, opened 2023-04-01
Streamline operators for expressions

trueqbit:
Runner-up for #1047.

* Provide comparison operators for the case where both arguments are a type wrapped in `c()`.
* Moved all operators into the internal namespace in order to facilitate ADL
* Streamlined operators that involve expressions wrapped in `c()` or arithmetic expressions

Contains also:
* Renamed `lesser` comparison functions to `less`
  * "Less" refers to quantity, "lesser" refers to quality
  * Deprecated `lesser_than()` and `less_than`


dev/conditions.h
```diff
@@ -1215,22 +1116,40 @@ namespace sqlite_orm {
     }
 
     template<class L, class R>
-    internal::lesser_than_t<L, R> lesser_than(L l, R r) {
+    internal::less_than_t<L, R> less_than(L l, R r) {
+        return {std::move(l), std::move(r)};
+    }
+
+    /**
+     *  [Deprecation notice] This function is deprecated and will be removed in v1.9. Use the accurately named function `less_than(...)` instead.
```
fnc12, line 1124: Let's deprecate it in 1.10. If we break back compatibility better give two versions to switch instead of 1. This is how it was done before by me every time.

PR #1190, opened 2023-06-06
add runsettings file

juandent:
to allow Catch2 to find tests. It is also necessary to set in Visual Studio 2022 Test\Configure Run Settings as shown in this image:

image


fnc12:
@juandent what problems it solves?

juandent:
Without  the runsettings, the tests are not found... at least not in my machine. It could be you have a global runsettings and so you do not need it, I don't know. Please check

fnc12:
@juandent I use Xcode or raw cmake on macOS. Is it something related to Visual Studio only?

juandent:
could be 
image

fnc12:
why Windows works well on CI machines without this file?

juandent:
but additionally VS2022 must be made to point to it

juandent:
who configured VS2022 in the CI mnachines?

trueqbit:
sqlite_orm comes with CMake build scripts, which in turn uses the specified compiler. CMake doesn't use Visual Studio per se, and even if they are kind of tied together, Visual Studio is not the msvc compiler, but ships with the msvc compiler.

What you are referring to is when you open a "solution" in the Visual Studio IDE.

CMakeLists.txt
```diff
@@ -74,6 +74,8 @@ if (MSVC)
         message(STATUS "Add /bigobj flag to compiler")
         add_compile_options(/bigobj)
     endif()
+    # For Visual Studio users
+    file(COPY_FILE ${PROJECT_SOURCE_DIR}/CMake/.runsettings ${PROJECT_BINARY_DIR}/.runsettings )
```
trueqbit, line 78: Need to conditionally copy the file: `if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21)`

cmake/.runsettings
trueqbit, line 1: I'd choose a better name for the file, e.g. _vs.runsettings_.

trueqbit (changes requested):

CMakeLists.txt
juandent: Ok I conditioned the call

PR #1223, opened 2023-10-05
added skip_types_and_constraints

fnc12:


dev/statement_serializer.h
```diff
@@ -998,12 +998,18 @@ namespace sqlite_orm {
...
-                   << " "
-                   << streaming_column_constraints(
-                          call_as_template_base<column_constraints>(polyfill::identity{})(column),
-                          column.is_not_null(),
-                          context);
+                ss << streaming_identifier(column.name);
+                if(!context.skip_types_and_constraints) {
+                    ss << " " << type_printer<field_type_t<column_type>>().print();
+                    const bool columnIsNotNull = column.is_not_null();
+                    auto constraintsTuple = streaming_column_constraints(
```
trueqbit, line 1005: No wonder you've got UB with gcc!
`streaming_column_constraints` isn't supposed to be used that way, in the sense that you shouldn't capture the returned tuple.
For starters, `streaming_column_constraints()` doesn't return you a constraints tuple, but a special tuple whose streaming operator is overloaded.
Second, the problem is that `column.is_not_null()` returns a temporary bool, which is captured by const-reference into the `constraintsTuple` variable; but at the end of that expression the boolean value is gone, and the `constraintsTuple` contains a dangling reference.

As in all other places, use it as an argument to the streaming operator:
```c++
                    ss << streaming_column_constraints(
                        call_as_template_base<column_constraints>(polyfill::identity{})(column),
                        column.is_not_null(),
                        context);
```

PR #1226, opened 2023-10-07
added list support

fnc12:


dev/serializing_util.h
```diff
@@ -197,8 +197,9 @@ namespace sqlite_orm {
             auto& context = get<2>(tpl);
 
             constexpr std::array<const char*, 2> sep = {", ", ""};
-            for(size_t i = 0, first = true; i < args.size(); ++i) {
-                ss << sep[std::exchange(first, false)] << serialize(args[i], context);
+            bool first = true;
+            for(auto it = args.begin(); it != args.end(); ++it) {
```
trueqbit, line 201: Why not use `for(const auto& arg : args)`?

tests/statement_serializer_tests/logical_operators.cpp
```diff
@@ -75,8 +75,26 @@ TEST_CASE("statement_serializer logical operators") {
...
         }
         SECTION("dynamic in") {
-            auto inValue = in(&User::id, {1, 2, 3});
-            stringValue = internal::serialize(inValue, context);
+            SECTION("initializer list") {
+                auto inValue = in(&User::id, {1, 2, 3});
+                stringValue = internal::serialize(inValue, context);
+            }
+            SECTION("vector") {
+                std::vector<int> vector;
```
trueqbit, line 83: Why not directly initialize `std::vector<int> vector{1, 2, 3}`?

tests/statement_serializer_tests/logical_operators.cpp
```diff
@@ -75,8 +75,26 @@ TEST_CASE("statement_serializer logical operators") {
...
+                vector.push_back(1);
+                vector.push_back(2);
+                vector.push_back(3);
+                auto inValue = in(&User::id, vector);
+                stringValue = internal::serialize(inValue, context);
+            }
+            SECTION("list") {
+                std::list<int> list;
```
trueqbit, line 91: Why not directly initialize `std::list<int> list{1, 2, 3}`?

trueqbit (approved):

dev/serializing_util.h
fnc12: cause I have weird habits left from old C++ standards. Just fixed. Thanks

tests/statement_serializer_tests/logical_operators.cpp
fnc12: fixed

tests/statement_serializer_tests/logical_operators.cpp
fnc12: fixed

PR #1230, opened 2023-10-11
added explicit null and not null

fnc12:
1. Added `null()` and `not_null()` public API for specifying explicitly how you want the column to be serialized during `sync_schema` call;
2. changed README according to the changes
3. covered with unit tests all the cases of column serialization with nullable and non-nullable types
4. changed serialization algo for `column_t`

### DIFF
--- a/dev/constraints.h
+++ b/dev/constraints.h
@@ -412,6 +412,9 @@ namespace sqlite_orm {
             }
         };
 
+        struct null_t {};
+
+        struct not_null_t {};
     }
 
     namespace internal {
@@ -440,6 +443,12 @@ namespace sqlite_orm {
         template<class T>
         SQLITE_ORM_INLINE_VAR constexpr bool is_generated_always_v = is_generated_always<T>::value;
 
+        template<class T>
+        using is_null = std::is_same<null_t, T>;
+
+        template<class T>
+        using is_not_null = std::is_same<not_null_t, T>;
+
         /**
          * PRIMARY KEY INSERTABLE traits.
          */
@@ -458,6 +467,8 @@ namespace sqlite_orm {
         using is_constraint =
             mpl::instantiate<mpl::disjunction<check_if<is_primary_key>,
                                               check_if<is_foreign_key>,
+                                              check_if<is_null>,
+                                              check_if<is_not_null>,
                                               check_if_is_template<unique_t>,
                                               check_if_is_template<default_t>,
                                               check_if_is_template<check_t>,
@@ -535,5 +546,13 @@ namespace sqlite_orm {
     template<class T>
     internal::check_t<T> check(T t) {
         return {std::move(t)};
+    }
+
+    inline internal::null_t null() {
+        return {};
+    }
+
+    inline internal::not_null_t not_null() {
+        return {};
     }
 }
--- a/dev/constraints.h
+++ b/dev/constraints.h
@@ -443,12 +443,6 @@ namespace sqlite_orm {
         template<class T>
         SQLITE_ORM_INLINE_VAR constexpr bool is_generated_always_v = is_generated_always<T>::value;
 
-        template<class T>
-        using is_null = std::is_same<null_t, T>;
-
-        template<class T>
-        using is_not_null = std::is_same<not_null_t, T>;
-
         /**
          * PRIMARY KEY INSERTABLE traits.
          */
@@ -467,8 +461,8 @@ namespace sqlite_orm {
         using is_constraint =
             mpl::instantiate<mpl::disjunction<check_if<is_primary_key>,
                                               check_if<is_foreign_key>,
-                                              check_if<is_null>,
-                                              check_if<is_not_null>,
+                                              check_if_is_type<null_t>,
+                                              check_if_is_type<not_null_t>,
                                               check_if_is_template<unique_t>,
                                               check_if_is_template<default_t>,
                                               check_if_is_template<check_t>,
--- a/dev/serializing_util.h
+++ b/dev/serializing_util.h
@@ -372,12 +372,11 @@ namespace sqlite_orm {
                 ss << sep[std::exchange(first, false)];
                 ss << serialize(constraint, context);
             });
-            using ConstraintsTuple = decltype(column.constraints);
-            constexpr bool constraintsHaveNullConstraint =
-                mpl::invoke_t<check_if_tuple_has_type<null_t>, ConstraintsTuple>::value;
-            constexpr bool constraintsHaveNotNullConstraint =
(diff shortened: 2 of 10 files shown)


dev/statement_serializer.h
```diff
@@ -886,6 +886,26 @@ namespace sqlite_orm {
             }
         };
 
+        template<>
+        struct statement_serializer<null_t, void> {
+            using statement_type = null_t;
+
+            template<class Ctx>
+            std::string operator()(const statement_type& statement, const Ctx& context) const {
+                return "NULL";
```
fnc12, line 895: I think we should remove `std::string` cast operators from AST nodes cause they don't help us but just spread serialization between `statement_serializer` and node's type. That's why I added the whole serialization literals right here

dev/statement_serializer.h
```diff
@@ -1031,13 +1051,11 @@ namespace sqlite_orm {
...
                     const bool columnIsNotNull = column.is_not_null();
-                    auto constraintsTuple = streaming_column_constraints(
-                        call_as_template_base<column_constraints>(polyfill::identity{})(column),
-                        columnIsNotNull,
-                        context);
-                    if(std::tuple_size<decltype(constraintsTuple)>::value > 0) {
-                        ss << " " << constraintsTuple;
-                    }
+                    ss << " "
+                       << streaming_column_constraints(
```
fnc12, line 1055: I put `streaming_column_constraints` back into `operator<<` statement not variable declaration to make it right. But `const bool columnIsNotNull = column.is_not_null();` is left as a constant in stack cause it has to have real memory address

trueqbit: It already has a "real memory address", in the form of an rvalue, so there's no need for another variable.
Like I mentioned in my [comment](https://github.com/fnc12/sqlite_orm/pull/1223/files#r1353270989), the only problem is if you would capture that rvalue in a variable that outlives the temporary rvalue.
That very rvalue will live for the whole streaming expression.

dev/constraints.h
```diff
@@ -440,6 +443,12 @@ namespace sqlite_orm {
         template<class T>
         SQLITE_ORM_INLINE_VAR constexpr bool is_generated_always_v = is_generated_always<T>::value;
 
+        template<class T>
```
trueqbit, line 446: Let's remove those traits as they are currently superfluous? Keep in sync with other constraints that are only used in a single place like `unique_t` or `collate_constraint_t`.

dev/constraints.h
```diff
@@ -458,6 +467,8 @@ namespace sqlite_orm {
         using is_constraint =
             mpl::instantiate<mpl::disjunction<check_if<is_primary_key>,
                                               check_if<is_foreign_key>,
+                                              check_if<is_null>,
```
trueqbit, line 470: If traits above are removed then replace with `check_if_is_type<null_t>`

dev/constraints.h
```diff
@@ -458,6 +467,8 @@ namespace sqlite_orm {
...
+                                              check_if<is_not_null>,
```
trueqbit, line 471: If traits above are removed then replace with `check_if_is_type<not_null_t>`

dev/statement_serializer.h
```diff
@@ -886,6 +886,26 @@ namespace sqlite_orm {
             }
         };
 
+        template<>
+        struct statement_serializer<null_t, void> {
+            using statement_type = null_t;
+
+            template<class Ctx>
+            std::string operator()(const statement_type& statement, const Ctx& context) const {
```
trueqbit, line 894: Please comment out unused parameters `statement` and `context` in order to avoid compiler warnings.

dev/statement_serializer.h
```diff
@@ -886,6 +886,26 @@ namespace sqlite_orm {
...
+                return "NULL";
+            }
+        };
+
+        template<>
+        struct statement_serializer<not_null_t, void> {
+            using statement_type = not_null_t;
+
+            template<class Ctx>
+            std::string operator()(const statement_type& statement, const Ctx& context) const {
```
trueqbit, line 904: Please comment out unused parameters `statement` and `context` in order to avoid compiler warnings.

dev/statement_serializer.h
trueqbit: It already has a "real memory address", in the form of an rvalue, so there's no need for another variable.
Like I mentioned in my [comment](https://github.com/fnc12/sqlite_orm/pull/1223/files#r1353270989), the only problem is if you would capture that rvalue in a variable that outlives the temporary rvalue.
That very rvalue will live for the whole streaming expression.

trueqbit (changes requested):
Please also add further unit tests to the "column is_not_null" test case in _tests/schema/column_tests.cpp_, which assert that `column_field::is_not_null()` returns the type-dependent nullable flag also when null/not null are specified explicitly.

dev/constraints.h
fnc12: done

dev/constraints.h
fnc12: already done

dev/constraints.h
fnc12: already done

dev/statement_serializer.h
fnc12: done

dev/statement_serializer.h
fnc12: done

dev/serializing_util.h
```diff
@@ -366,11 +366,24 @@ namespace sqlite_orm {
...
-                ss << serialize(constraint, context) << ' ';
+            auto first = true;
+            constexpr std::array<const char*, 2> sep = {" ", ""};
+            iterate_tuple(column.constraints, [&ss, &context, &first, &sep](auto& constraint) {
+                ss << sep[std::exchange(first, false)];
+                ss << serialize(constraint, context);
             });
-            if(isNotNull) {
-                ss << "NOT NULL";
+            using ConstraintsTuple = decltype(column.constraints);
```
trueqbit, line 375: Simplified:
```c++
            using constraints_tuple = decltype(column.constraints);
            constexpr bool hasExplicitNullableConstraint =
                mpl::invoke_t<mpl::disjunction<check_if_tuple_has<is_null>, check_if_tuple_has<is_not_null>>,
                              constraints_tuple>::value;
            if(!hasExplicitNullableConstraint) {
```

trueqbit (changes requested):
I am absolutely certain that I had requested another change, but it got lost somehow... Sorry for the inconvenience!

dev/serializing_util.h
fnc12: done

trueqbit (approved):

PR #1237, opened 2023-10-20
Operators for alias columns, alias columns from column pointers

trueqbit:
This PR further simplifies working with alias columns.

* Alias columns can be made from a derived object member pointer without forming an explicit column pointer. E.g. instead of writing `alias_column<alias_a<Derived>>(column<Derived>(&Object::field))` one can simply write `alias_column<alias_a<Derived>>(&Object::field)`.
* The convenience pointer-to-member operator `operator->*()` now also checks whether object type of a member pointer is the same as the aliased type, just like the named function `alias_column()`.
E.g. this means that `"my"_alias.for_<Object>()->*&Object::field` works, but `"xyz"_alias.for_<Something>()->*&Object::field` does not.
* Overloaded operators for alias columns. One can write e.g. `alias_column<myalias>(&Object::id) == 4` without "lifting" `4` into an expression using `c(4)`, which improves legibility.

Additionally:
- Column pointers and alias columns can now be used as constant expressions. This IMO improves the ability to create short definitions for long expressions that can be reused afterwards.
E.g. `constexpr auto reuseColumn = column_pointer<Derived>(&Object::id);` or `constexpr auto aliasedColumn = alias_column<alias_a<Derived>>(&Object::id);`.
- Updated vcpkg test environment on appveyor to tag "2023.10.19".


dev/schema/table.h
```diff
@@ -277,30 +277,32 @@ namespace sqlite_orm {
...
+#ifndef SQLITE_ORM_AGGREGATE_BASES_SUPPORTED
             virtual_table_t(std::string name, module_details_type module_details) :
                 basic_table{std::move(name)}, module_details{std::move(module_details)} {}
+#endif
 
-            /**
-             *  Call passed lambda with columns not having the specified constraint trait `OpTrait`.
-             *  @param lambda Lambda called for each column.
+            /** 
+             *  Call passed lambda with columns not having the specified constraint trait `OpTrait`.
```
fnc12, line 286: why do we need a pace after the line?

dev/statement_serializer.h
```diff
@@ -163,7 +163,7 @@ namespace sqlite_orm {
             using statement_type = current_time_t;
 
             template<class Ctx>
-            std::string operator()(const statement_type& statement, const Ctx& context) {
+            std::string operator()(const statement_type& /*statement*/, const Ctx& /*context*/) {
```
fnc12, line 166: 👍

fnc12 (approved):

dev/schema/table.h
trueqbit: A copy/paste accident!

PR #1241, opened 2023-10-25
Compound statements with variable number of select statements

trueqbit:
This PR extends the operators for compound statements (i.e. `union_, union_all, intersect, except`)
-    to accept a variable number of single select statements, which is especially useful for `UNION ALL`,
-    to allow the select statements to return a common type.

Additionally:
- Simplified code for the calculation of node tuples.
- Improved unit tests for node tuples.

My motivating use case is to be able to handpick multiple value ranges:

```c++
struct Object {
    int64 id;
    int64 difference;   // diff value in days
};

union_all(
    select(cast<int64>(min(&Object::difference))),
    select(count(), where(c(&Object::difference) < 0)),
    select(count(), where(c(&Object::difference) >= 0 and c(&Object::difference) < 7)),
    select(count(), where(c(&Object::difference) >= 7 and c(&Object::difference) < 15)),
    ...
);
```

### DIFF
--- a/dev/column_result.h
+++ b/dev/column_result.h
@@ -1,14 +1,15 @@
 #pragma once
 
 #include <type_traits>  //  std::enable_if, std::is_same, std::decay, std::is_arithmetic, std::is_base_of
-#include <tuple>  //  std::tuple
 #include <functional>  //  std::reference_wrapper
 
-#include "functional/cxx_universal.h"
+#include "functional/cxx_universal.h"  //  ::nullptr_t
+#include "functional/cxx_type_traits_polyfill.h"
 #include "functional/mpl.h"
 #include "tuple_helper/tuple_traits.h"
 #include "tuple_helper/tuple_fy.h"
 #include "tuple_helper/tuple_filter.h"
+#include "tuple_helper/tuple_transformer.h"
 #include "tuple_helper/same_or_void.h"
 #include "type_traits.h"
 #include "member_traits/member_traits.h"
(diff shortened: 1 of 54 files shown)


dev/tuple_helper/tuple_iteration.h
```diff
@@ -35,7 +35,7 @@ namespace sqlite_orm {
 
 #if defined(SQLITE_ORM_FOLD_EXPRESSIONS_SUPPORTED) && defined(SQLITE_ORM_IF_CONSTEXPR_SUPPORTED)
         template<bool reversed = false, class Tpl, size_t... Idx, class L>
-        void iterate_tuple(const Tpl& tpl, std::index_sequence<Idx...>, L&& lambda) {
+        void iterate_tuple(Tpl& tpl, std::index_sequence<Idx...>, L&& lambda) {
```
fnc12, line 38: why did you remove `const`? if we don't modify the tuple then we don't ned it as mutable

tests/prepared_statement_tests/update_all.cpp
```diff
@@ -39,15 +39,6 @@ TEST_CASE("Prepared update all") {
 
     {  //  by val
         auto statement = storage.prepare(update_all(set(assign(&User::name, conc(&User::name, "_")))));
-        using Statement = decltype(statement);
```
fnc12, line 42: why did you remove these static parts of ut?

dev/column_result.h
```diff
@@ -220,20 +227,18 @@ namespace sqlite_orm {
         struct column_result_t<DBOs, column_pointer<T, F>, void> : column_result_t<DBOs, F> {};
 
         template<class DBOs, class... Args>
-        struct column_result_t<DBOs, columns_t<Args...>, void> {
-            using type = tuple_cat_t<tuplify_t<column_result_of_t<DBOs, std::decay_t<Args>>>...>;
-        };
+        struct column_result_t<DBOs, columns_t<Args...>, void>
+            : conc_tuple<tuplify_t<column_result_of_t<DBOs, std::decay_t<Args>>>...> {};
```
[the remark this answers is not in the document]
trueqbit: 🙏

tests/prepared_statement_tests/update_all.cpp
trueqbit: They are covered by the tests already in tests/bindables.cpp and the new "set assign" section in [tests/node_tuple.cpp](https://github.com/fnc12/sqlite_orm/pull/1241/files/9813ec16425e3e4f070a7c22ddafb5cfa16d792f?diff=split&w=0#diff-567fe4ab260b49c101b491d47de299ea1da4e62338918924d86e98c91b8d51fb).

dev/tuple_helper/tuple_iteration.h
trueqbit: `compound_operator`'s contructor [sets](https://github.com/fnc12/sqlite_orm/blob/9813ec16425e3e4f070a7c22ddafb5cfa16d792f/dev/select_constraints.h#L121) the `highest_level` flag of the single select-statements.

fnc12 (approved):

PR #1244, opened 2023-11-11
Uniform syntax for column pointer expressions through table references

trueqbit:
Since the advent of a new syntax for alias columns in the form of

```c++
constexpr auto z_alias = "z"_alias.for_<Object>();
select(z_alias->*&Object::id);
```

and the upcoming CTE feature that uses the same notation:

```c++
constexpr auto xaxis = "xaxis"_cte;
select(xaxis->*x);
```

it would be very convenient and again a huge improvement to the legibility of SQL statement expressions if column pointers can be formed using the same syntax.

```c++
struct Object { int64 id; };
struct Derived : Object {};
constexpr auto derived = c<Derived>();
select(derived->*&Derived::id);
```

Additionally, this PR contains the following improvements:

- A lot more expressions are now available for column pointers (e.g. `storage.count(column<Derived>(&Derived::id))` works)
- A myriad more unit tests for column pointers
- More unit tests for alias column pointers

It must be clarified how a "table reference" can be created. Currently available:
```c++
// 1.
constexpr auto derived = c<Derived>();
// 2.
constexpr auto derived = column<Derived>();
```

### DIFF
--- a/dev/alias_traits.h
+++ b/dev/alias_traits.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <type_traits>  //  std::remove_const, std::is_base_of, std::is_same
+#include <type_traits>  //  std::remove_const, std::is_base_of, std::is_same, std::type_identity
 #ifdef SQLITE_ORM_WITH_CPP20_ALIASES
 #include <concepts>
 #endif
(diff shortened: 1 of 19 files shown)


fnc12 (approved):

fnc12:
thanks @trueqbit

PR #1248, opened 2023-11-15
Permanent memory space for a user-defined function

trueqbit:
Instead of recreating a user-defined function on the "free store" each time it is used for calculations, the memory space is now allocated once together with the "proxy" object registered with SQLite, so that the user-defined function only needs to be reconstructed.

Additional features in this PR:
- User-defined functions participate in operator overloading, just like expressions, colum pointers, and column aliases.
- Possibility of passing arguments when constructing the user-defined function

Additional improvements in this PR:
- Reorganized code around user-defined functions.
- Simplified creating a tuple of arguments for user-defined function calls from a SQL query.
- Replaced custom `call()` function by `std::apply()`.


dev/udf_proxy.h
```diff
@@ -0,0 +1,124 @@
...
+
+            struct destruct_only_deleter {
+                template<class F>
+                void operator()(F* f) const noexcept {
+                    f->~F();
+                }
+            };
+
+            std::string name;
+            int argumentsCount;
```
fnc12, line 37: better default values be added like this `int argumentsCount = 0;`, `bool constructed = false;` and `void* const udfHandle = nullptr;` to be sure we never face trash values there

dev/udf_proxy.h
```diff
@@ -0,0 +1,124 @@
...
+
+        /*
+         *  A veneer to `udf_proxy` that provides memory space for a user-defined function object.
+         *  
+         *  Note: it must be a veneer, i.e. w/o any non-trivially destructible member variables.
+         */
+        template<class UDF>
+        struct udf_proxy_veneer : udf_proxy {
+            // allocated memory for user-defined function
+            alignas(UDF) char storage[sizeof(UDF)];
```
fnc12, line 57: from the first sight `storage` reminds of `storage_t` but it is a whole different entity type. Probably we better rename it to something that has no existing associations in the lib like `bytes`

dev/udf_proxy.h
```diff
@@ -0,0 +1,124 @@
...
+            }
+            {}
+        };
+
+        inline void scalar_function_callback(sqlite3_context* context, int argsCount, sqlite3_value** values) {
+            udf_proxy* proxy = static_cast<udf_proxy*>(sqlite3_user_data(context));
+            if(proxy->argumentsCount != -1 && proxy->argumentsCount != argsCount) {
+                throw std::system_error{orm_error_code::arguments_count_does_not_match};
+            }
+            const std::unique_ptr<void, xdestroy_fn_t> udfHandle{proxy->constructAt(proxy->udfHandle), proxy->destroy};
```
fnc12, line 96: what if this callback is called from several threads simultaneously? Looks like it will produce data races cause we have only one pack of bytes for function object but several callers. Current `dev` implementation doesn't have this issue

dev/udf_proxy.h
trueqbit: The main reason in my head was not to duplicate the constructor for not fully C++14 compliant compilers (or certain versions thereof).

dev/udf_proxy.h
trueqbit: I have changed the way to allocate memory anyway, with commit [2c23449](https://github.com/fnc12/sqlite_orm/pull/1248/commits/2c234491ccfe05111656110bb06c1e0306a438f9).

dev/udf_proxy.h
trueqbit: Your concern is absolutely valid, and I have dealt with this issue of user-defined functions (aka "application-defined SQL functions") myself.

There are 2 things to consider:

1. if you think about it, you come to the conclusion that if SQLite were to impose thread-safety on a callback, it would be an almost impossible requirement to fulfill. This would mean that sqlite_orm would have to document that a user-defined function must be programmed to be thread-safe (we never know what the user-defined function is doing). Also, the SQLite documentation does not mention that a callback must be inherently thread-safe.
1. the callback of the user-defined function is part of a database connection. Therefore, there is a separate "proxy" object for each connection.
  Also, I checked the callstack, and the call to a user-defined function is part of evaluating a prepared statement, i.e. the generation of result rows with `sqlite3_step()`. The `sqlite3_step()` function in turn locks a database mutex, as you can see in the attached screenshot of the callstack. From another perspective, the step function itself would not create multiple threads, but multiple threads could be created to call the step function.

The conclusion is that calling a user-defined function as part of a step during the generation of result rows is inherently concurrency-safe.

image

fnc12 (approved):

fnc12 (approved):

dev/storage_base.h
```diff
@@ -248,53 +258,46 @@ namespace sqlite_orm {
...
-                        values_to_tuple{}(values, argsTuple, argsCount);
-                        auto result = call(udf, std::move(argsTuple));
-                        statement_binder<return_type>().result(context, result);
-                    }));
-
-                if(this->connection->retain_count() > 0) {
-                    sqlite3* db = this->connection->get();
-                    try_to_create_function(db, static_cast<scalar_udf_proxy&>(*this->scalarFunctions.back()));
-                }
+                this->create_scalar_function_impl<F>(/* constructAt */ [args...](void* location) {
```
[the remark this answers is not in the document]
trueqbit: The closure (lambda) requires a copy of the arguments, as it outlasts them.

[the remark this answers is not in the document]
trueqbit: So, the point is: the lambda must store a copy of the arguments, because it needs to construct the application-defined function object later, and does so repeatedly. Movable-only types cannot be used for this, obviously.

dev/storage_base.h
```diff
@@ -846,8 +871,8 @@ namespace sqlite_orm {
             std::map<std::string, collating_function> collatingFunctions;
             const int cachedForeignKeysCount;
             std::function<int(int)> _busy_handler;
-            std::vector<std::unique_ptr<udf_proxy_base>> scalarFunctions;
-            std::vector<std::unique_ptr<udf_proxy_base>> aggregateFunctions;
+            std::vector<std::unique_ptr<udf_proxy>> scalarFunctions;
```
trueqbit, line 874: @fnc12 Since `udf_proxy` is now a complete structure itself (without derived classes), do you mind using an `std::list<udf_proxy>` instead of the `unique_ptr` indirection?

fnc12: yes it is a good call

PR #1258, opened 2023-12-24
Improving parenthesizing of binary expressions

trueqbit:
The serialization of binary expressions currently suffers from two problems:

1. The statement serializer too eagerly puts parentheses around every built-in function and binary expression (comparison, logic, arithmetic, bitwise operators) and is additionally dependent on a flag provided by the outer context.
2. Subqueries in binary expressions in columns currently only work because the outer expressions do not change the `serializer_context<>::use_parentheses=true` flag. As soon as it is changed, as in CTEs, the serialization goes wrong.

1) can be improved and 2) remedied in the following ways:

* Binary operators/conditions: Parentheses are only needed to preserve precedence by following simple logic: only "subtree" expressions need to be parenthesized, "leaf" expressions do not.
* No parentheses around built-in functions necessary - they are already independent expressions (and the outer expression already takes over the parentheses, e.g. `check()`, `default()`).
* Establishing a parenthesized subcontext for potential subqueries in binary expressions, such that `select(&Employee::salary >= select(avg(&Employee::salary)))` works.

This leads to a simplified approach, as the serializer is independent of the context.
In addition, fewer characters are serialized with simple binary expressions, which also improves the readability of statements when they are examined.

Also in this PR:
* Allow conditional expressions to be equality-comparable, so expressions like `select(c(5) > 3 == true)` are possible instead of `select(c(5) > 3 == c(true))`
* Merged handling of binary conditions and operators (AST iteration, serialization)
* Appveyor vcpkg environmnent updated to 2023.12.12.
* Fixed "left_and_inner_join" example

### DIFF
--- a/examples/left_and_inner_join.cpp
+++ b/examples/left_and_inner_join.cpp
@@ -60,6 +60,7 @@ inline auto initStorage(const std::string& path) {
 int main(int, char**) {
 
     auto storage = initStorage("chinook.db");
+    storage.sync_schema();
 
     using namespace sqlite_orm;
 
--- a/examples/left_and_inner_join.cpp
+++ b/examples/left_and_inner_join.cpp
@@ -60,7 +60,6 @@ inline auto initStorage(const std::string& path) {
 int main(int, char**) {
 
     auto storage = initStorage("chinook.db");
-    storage.sync_schema();
 
     using namespace sqlite_orm;
 
--- a/examples/subquery.cpp
+++ b/examples/subquery.cpp
@@ -1306,8 +1306,8 @@ int main(int, char**) {
         }
     }
     {
-        //  SELECT employee_id,first_name,last_name,salary,avg(salary),salary>avg(salary)
-        //  FROM employees;
+        //  SELECT employee_id, first_name, last_name, salary, (SELECT AVG(salary) FROM employees), salary > (SELECT AVG(salary) FROM employees)
+        //  FROM "employees";
         auto rows = storage.select(columns(&Employee::id,
                                            &Employee::firstName,
                                            &Employee::lastName,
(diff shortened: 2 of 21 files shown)


examples/left_and_inner_join.cpp
```diff
@@ -60,6 +60,7 @@ inline auto initStorage(const std::string& path) {
 int main(int, char**) {
 
     auto storage = initStorage("chinook.db");
+    storage.sync_schema();
```
fnc12, line 63: no no no, here you have to download `chinook.db` instead of syncing it from here https://www.sqlitetutorial.net/sqlite-sample-database/ (text search `Download SQLite sample database`)
- [ ] remove `sync_schema`  call

examples/subquery.cpp
```diff
@@ -1305,6 +1305,22 @@ int main(int, char**) {
                  << endl;
         }
     }
+    {
+        //  SELECT employee_id,first_name,last_name,salary,avg(salary),salary>avg(salary)
+        //  FROM employees;
+        auto rows = storage.select(columns(&Employee::id,
```
fnc12, line 1311: I bet SQL here is different than
```sql
SELECT employee_id,first_name,last_name,salary,avg(salary),salary>avg(salary)
FROM employees;
```
. I am sure it is
```sql
SELECT employee_id, first_name, last_name, salary, (SELECT avg(salary) FROM employees),salary > (SELECT avg(salary) FROM employees)
FROM employees;
```
Is it right? If it is let's change the comment to make it clear for future readers

fnc12 (changes requested):

fnc12:
@trueqbit everything is brilliant (as always) except two items

examples/left_and_inner_join.cpp
trueqbit: Oh oh. I see the file comment now! I was simply running through all examples... Thx

PR #1278, opened 2024-05-03
Support for selecting multiple objects and unmapped but named structs

trueqbit:


dev/object_from_column_builder.h
```diff
@@ -44,5 +48,33 @@ namespace sqlite_orm {
...
+         *  
+         *  This plays together with `column_result_of_t`, which returns `object_t<O>` as `table_referenece<O>`
+         */
+        template<class O, class DBOs>
+        struct struct_extractor<table_reference<O>, DBOs> {
+            const DBOs& db_objects;
+
+            O extract(const char* columnText) const = delete;
+
+            O extract(sqlite3_stmt* stmt, int&& /*nextColumnIndex*/) const {
```
fnc12, line 63: why `&&`?

dev/row_extractor.h
```diff
@@ -445,5 +422,88 @@ namespace sqlite_orm {
...
+         */
+        template<class O, class... Args, class DBOs>
+        struct struct_extractor<structure<O, std::tuple<Args...>>, DBOs> {
+            const DBOs& db_objects;
+
+            O extract(const char* columnText) const = delete;
+
+            O extract(sqlite3_stmt* stmt, int&& /*nextColumnIndex*/ = 0) const {
+                int columnIndex = -1;
+                return O{make_row_extractor<Args>(this->db_objects).extract(stmt, ++columnIndex)...};
```
fnc12, line 498: this is good!

fnc12 (approved):

dev/object_from_column_builder.h
trueqbit: For the overload resolution, discerning it from the other overload taking `int&`.
It's because of a few things that play together.

There are two contexts in which row extractors are called:

1. as top-level extractor.
2. as embedded/nested extractor, i.e. as part of extracting other columns.

.
1. `storage_t::_execute_select()` is a generic implementation and always passes `0` to any top-level row extractor. It needs to do that because the value extractors always expect the column index, so we need a default argument for the structure extractors. So the argument is a temporary int [I just forgot the default value 0], which the structure extractors just discard, assuming they are only called from the top level.
2. `struct_extractor<tuple<...>>` in turn carries on the column index across multiple columns. In this case the other overload taking an `int&` is called.

I made a note on all the extraction methods.

dev/row_extractor.h
trueqbit: It's a thing of beauty, indeed :)

fnc12 (approved):

PR #1282, opened 2024-05-05
Added prefix with unit tests

fnc12:


dev/serializing_util.h
```diff
@@ -422,11 +422,15 @@ namespace sqlite_orm {
                     }
                 }
             } else {
-                constexpr bool hasUnindexedOption =
-                    mpl::invoke_t<check_if_has_type<unindexed_t>, constraints_tuple>::value;
-                if SQLITE_ORM_CONSTEXPR_IF(hasUnindexedOption) {
-                    ss << " UNINDEXED";
-                }
+                iterate_tuple(column.constraints, [&ss, &context](auto& constraint) {
```
trueqbit, line 425: It's easier to get a filtered index sequence:
```c++
                using fts5_index_sequence = filter_tuple_sequence_t<
                    constraints_tuple,
                    mpl::disjunction<check_if_is_type<unindexed_t>, check_if_is_template<prefix_t>>::template fn>;
                iterate_tuple(column.constraints, fts5_index_sequence{}, [&ss, &context](auto& constraint) {
                    ss << ' ' << serialize(constraint, context);
```

dev/constraints.h
```diff
@@ -433,6 +445,8 @@ namespace sqlite_orm {
 #else
             false;
 #endif
+        template<class T>
```
trueqbit, line 448: We don't need that. See _serializing_util.h_.

include/sqlite_orm/sqlite_orm.h
```diff
@@ -1796,12 +1796,24 @@ namespace sqlite_orm {
...
 
         struct unindexed_t {};
 
+        template<class T>
+        struct prefix_t {
+            using value_type = T;
+
+            value_type value;
+
+#ifndef SQLITE_ORM_AGGREGATE_BASES_SUPPORTED
```
trueqbit, line 1812: This constructor isn't necessary in all cases.

trueqbit (changes requested):

include/sqlite_orm/sqlite_orm.h
fnc12: Снимок экрана 2024-05-06 в 11 07 51
are you sure?

trueqbit: Yes, I am absolutely certain. There are no aggregate base classes involved. The error happens because you removed the `unique_t` constructor - the [branch build](https://ci.appveyor.com/project/fnc12/sqlite-orm/builds/49752631/job/w6wdg9utb2gf2e3h) of your previous commit has failed, too.

fnc12: oh now I see how it works. Thanks

dev/serializing_util.h
trueqbit: @fnc12 II have an even better suggestion. The question is, why do we need to differentiate between constraint types here at all?
You can never add regular column constraints to FTS5 columns and vice versa. In any case, it is better to prevent this when creating the column, not when serializing it.

So, this implementation is much simpler:
```c++
template<class... Op, class Ctx>
std::ostream& operator<<(std::ostream& ss,
                         std::tuple<const streaming<stream_as::column_constraints>&,
                                    const column_constraints<Op...>&,
                                    const bool&,
                                    Ctx> tpl) {
    const auto& column = std::get<1>(tpl);
    const bool& isNotNull = std::get<2>(tpl);
    auto& context = std::get<3>(tpl);

    using constraints_tuple = decltype(column.constraints);
    iterate_tuple(column.constraints, [&ss, &context](auto& constraint) {
        ss << ' ' << serialize(constraint, context);
    });
    // add implicit null constraint
    if(!context.fts5_columns) {
        constexpr bool hasExplicitNullableConstraint =
            mpl::invoke_t<mpl::disjunction<check_if_has_type<null_t>, check_if_has_type<not_null_t>>,
                          constraints_tuple>::value;
        if SQLITE_ORM_CONSTEXPR_IF(!hasExplicitNullableConstraint) {
            if(isNotNull) {
                ss << " NOT NULL";
            } else {
                ss << " NULL";
            }
        }
    }

    return ss;
}
```

fnc12: @trueqbit the case is that SQLite FTS docs say that one can add regular column constraints to FTS columns but those constraints will be ignored by SQLite engine. That's why I decided to remove them from serialization. But let's try your idea

trueqbit: Interesting - I haven't read all the details... I would say that in this case it would even argue for not checking the restrictions on FTS5 columns.

PR #1286, opened 2024-05-09
Removed `prefix` from column constraints

trueqbit:
Introduces the checking of elements and constraints that are passed to the table factory functions.

### DIFF
--- a/dev/constraints.h
+++ b/dev/constraints.h
@@ -3,7 +3,7 @@
 #include <system_error>  //  std::system_error
 #include <ostream>  //  std::ostream
 #include <string>  //  std::string
-#include <tuple>  //  std::tuple, std::make_tuple
+#include <tuple>  //  std::tuple
 #include <type_traits>  //  std::is_base_of, std::false_type, std::true_type
 
 #include "functional/cxx_universal.h"
@@ -356,7 +356,7 @@ namespace sqlite_orm {
 
             template<class... Rs>
             foreign_key_t<std::tuple<Cs...>, std::tuple<Rs...>> references(Rs... refs) {
-                return {std::move(this->columns), std::make_tuple(std::forward<Rs>(refs)...)};
+                return {std::move(this->columns), {std::forward<Rs>(refs)...}};
             }
         };
 #endif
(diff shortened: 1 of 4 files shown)


dev/constraints.h
```diff
@@ -481,7 +481,6 @@ namespace sqlite_orm {
                                                              check_if_is_type<null_t>,
                                                              check_if_is_type<not_null_t>,
                                                              check_if_is_type<unindexed_t>,
-                                                             check_if_is_template<prefix_t>,
```
fnc12, line 484: are you sure about that? the template is named `is_constraint` not `is_column_constraint`. There are two types of constraints: table constraints and column constraints. E.g. FOREIGN KEY is table constraint and it exists in this template. So I doubt we need to remove `prefix_t` from this list

trueqbit: I had in mind that `is_constraint` is only ever used to check column constraints in `make_column()`.
However, your point is fair enough. See next commit.

dev/constraints.h
```diff
@@ -501,32 +499,35 @@ namespace sqlite_orm {
...
     template<class... Cs>
     internal::foreign_key_intermediate_t<Cs...> foreign_key(Cs... columns) {
-        return {std::make_tuple(std::forward<Cs>(columns)...)};
+        return {{std::forward<Cs>(columns)...}};
     }
 #endif
 
     /**
-     *  UNIQUE constraint builder function.
+     *  UNIQUE table constraint builder function.
```
fnc12, line 515: `UNIQUE` is also can be applied at table level. I'd remove `table` word from here. More info [here](https://www.sqlite.org/lang_createtable.html)

dev/schema/table.h
```diff
@@ -402,11 +412,17 @@ namespace sqlite_orm {
 
     template<class... Cs, class T = typename std::tuple_element_t<0, std::tuple<Cs...>>::object_type>
     internal::using_fts5_t<T, Cs...> using_fts5(Cs... columns) {
+        static_assert(polyfill::conjunction_v<internal::is_table_element_or_constraint<Cs>...>,
```
fnc12, line 415: these checks are really useful. Thanks

dev/constraints.h
trueqbit: There are two overloads, see below.

fnc12: oh I see. Pardon

fnc12 (approved):

PR #1308, opened 2024-06-09
added various transation guard API to storage_base

fnc12:


tests/transaction_tests.cpp
```diff
@@ -102,20 +102,60 @@ TEST_CASE("Transaction guard") {
 
     SECTION("insert, call make a storage to call an exception and check that rollback was fired") {
         auto countBefore = storage.count<Object>();
-        try {
-            auto guard = storage.transaction_guard();
+        SECTION("transaction_guard") {
+            try {
```
trueqbit, line 106: Can we use a dedicated Catch2 test macro for this, please?
```c++
REQUIRE_THROWS_WITH(storage.transaction([&] {
    storage.insert(Object{0, "John"});
    storage.get<Object>(0);
    return true;
}),
                    "Not found");
```

tests/transaction_tests.cpp
```diff
@@ -102,20 +102,60 @@ TEST_CASE("Transaction guard") {
...
+                auto guard = storage.transaction_guard();
 
-            storage.insert(Object{0, "John"});
+                storage.insert(Object{0, "John"});
 
-            storage.get<Object>(-1);
+                storage.get<Object>(-1);
```
trueqbit, line 111: Also, the object ID to retrieve should be _**0**_, no?

trueqbit (changes requested):

tests/transaction_tests.cpp
fnc12: can be 0, but any value below 0 is also suitable for non-existing object

fnc12: but I confess this code was written by me during epoch when I was sure that `0` is a valid ID in SQLite

tests/transaction_tests.cpp
fnc12: let me try

fnc12: actually I'd like to test exception's error code more than its description string. But I don't see any option of doing it with `REQUIRE_THROWS_WITH` so I'd like to keep the things as they are

fnc12: done. Please check

tests/transaction_tests.cpp
```diff
@@ -100,22 +100,71 @@ TEST_CASE("Transaction guard") {
 
     storage.insert(Object{0, "Jack"});
 
-    SECTION("insert, call make a storage to call an exception and check that rollback was fired") {
-        auto countBefore = storage.count<Object>();
-        try {
-            auto guard = storage.transaction_guard();
+    struct ExceptionMatcher : Catch::Matchers::MatcherGenericBase {
+        std::string what;
```
trueqbit, line 104: Checking the error code alone would be enough, we don't care about the exact message string at this point...

tests/transaction_tests.cpp
```diff
@@ -100,22 +100,71 @@ TEST_CASE("Transaction guard") {
...
+            std::stringstream ss;
+            const std::string message = get_orm_error_category().message(this->errorCode.value());
+            ss << "Exception with code: " << message << ", what: " << this->what;
+            return ss.str();
+        }
+    };
 
-            REQUIRE(countBefore == countNow);
+    const ExceptionMatcher notFoundExceptionMatcher("Not found", make_error_code(orm_error_code::not_found));
+    SECTION("insert, call make a storage to call an exception and check that rollback was fired") {
```
trueqbit, line 124: BTW, this section within this test case replaces the same section in test case "transaction_rollback" some lines above, so the old one can be removed AFAIK?

trueqbit (approved):

tests/transaction_tests.cpp
fnc12: fixed

tests/transaction_tests.cpp
fnc12: agree, I removed `TEST_CASE("transaction_rollback")`

trueqbit (approved):

PR #1310, opened 2024-06-10
Ability to specify a table reference for the object iteration interface

trueqbit:
Originally, I wanted to extend the ‘table reference’ feature to the iteration of objects:
```c++
struct Object {};
constexpr auto object_table = c<Object>();
auto storage = make_storage("", make_table<object_table>());
storage.iterate<object_table>();
```

This PR also includes changes and improvements to the object iterator:
* Disentangled iterator from its object view: an iterator doesn't need to access the object view itself nor the storage. All it needs is access to the 'db objects'.
* Made all of `iterator_t` private, it's currently not designed for derivation.
* Removed redundant access checks.
* Ensure that `iterator_t` is a `LegacyInputIterator`
* Turned the object view into a so-called 'borrowed view', which guarantees that a function can take it by value and return iterators obtained from it without danger of dangling.

### DIFF
--- a/dev/iterator.h
+++ b/dev/iterator.h
@@ -1,14 +1,13 @@
 #pragma once
 
 #include <sqlite3.h>
-#include <cassert>  //  assert
 #include <memory>  //  std::shared_ptr, std::make_shared
 #include <utility>  //  std::move
 #include <iterator>  //  std::input_iterator_tag
 #include <system_error>  //  std::system_error
 #include <functional>  //  std::bind
 
-#include "functional/cxx_universal.h"
+#include "functional/cxx_universal.h"  //  ::ptrdiff_t
 #include "statement_finalizer.h"
 #include "error_code.h"
 #include "object_from_column_builder.h"
@@ -18,17 +17,26 @@
 namespace sqlite_orm {
     namespace internal {
 
+        /*  
+         *  (Legacy) Input iterator over a result set for a mapped object.
+         */
         template<class O, class DBOs>
-        struct iterator_t {
-            using value_type = O;
+        class iterator_t {
+          public:
             using db_objects_type = DBOs;
+
+            using iterator_category = std::input_iterator_tag;
+            using difference_type = ptrdiff_t;
+            using value_type = O;
+            using reference = O&;
+            using pointer = O*;
 
           private:
             /**
-                pointer to the view's db objects member variable.
+                pointer to the db objects.
                 only null for the default constructed iterator.
              */
-            const db_objects_type** db_objects = nullptr;
+            const db_objects_type* db_objects = nullptr;
 
             /**
              *  shared_ptr is used over unique_ptr here
@@ -45,8 +53,8 @@ namespace sqlite_orm {
             void extract_object() {
                 this->current = std::make_shared<value_type>();
                 object_from_column_builder<value_type> builder{*this->current, this->stmt.get()};
-                assert(*this->db_objects);
-                pick_table<value_type>(**this->db_objects).for_each_column(builder);
+                auto& table = pick_table<value_type>(*this->db_objects);
+                table.for_each_column(builder);
             }
 
             void step() {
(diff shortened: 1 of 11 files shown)


dev/iterator.h
```diff
@@ -16,45 +16,49 @@
...
              */
             std::shared_ptr<value_type> current;
 
-            void extract_value() {
-                auto& dbObjects = obtain_db_objects(this->view->storage);
+            void extract_object() {
                 this->current = std::make_shared<value_type>();
                 object_from_column_builder<value_type> builder{*this->current, this->stmt.get()};
-                pick_table<value_type>(dbObjects).for_each_column(builder);
+                assert(*this->db_objects);
```
fnc12, line 48: `assert` performs nothing in release build. Let's replace `assert` with exception of something else which works in both debug and release build.

fnc12 (changes requested):

dev/iterator.h
trueqbit: That's my intention. The assert is just there to help in debugging of something you are never supposed to do in the first place (i.e. using an invalid iterator), and in release builds it would dereference a nullptr, which is good, and much better than the dangling pointer that we had before.

fnc12: > in release builds it would dereference a nullptr

that leads to UB, let's avoid spawning UB please. Originally there are no places in the lib which perform dereferencing before changing for `nullptr` to reduce UB which is the reason why C++ is hated a lot. Libs which spawn UB a lot are hated more

trueqbit: It's not UB - it's a well-defined null pointer exception (because `~view_t()` sets it to a `nullptr`.
If you compare the diff, the iterator had a pointer to the view. If the view went out of scope and an iterator would have been accessed, then you hit UB. So contrary to what you say, there is a place that performs unchecked dereferencing. And it was never checked.

I stand by my argument that this is much worse than a controlled and well-defined null pointer exception, which - again - should not occur.
It's a programming error of the programmer still using an iterator after the lifetime of a view. That's BTW no different than other iterators in general.

trueqbit: But I have a completely different suggestion: we don't need the check at all. The iterator exists perfectly without the view, it is self-sufficient, so to speak, and the ‘storage’ together with the ‘db objects’ is a global instance anyway.

fnc12: simple dereferencing a null pointer never leads to exception, it produces a reference which equals `nullptr` which is UB. Try it by yourself.
```c++
int *pointer = nullptr;
int &reference = *pointer;  // no exception is thrown here
```

trueqbit: It will be thrown once you access `reference`.
image

fnc12: yes but it is not guaranteed by C++ standard. C++ standard only guarantees UB in that case, but segfault or any exception is compiler specific thing. This is what I am talking about. Of course in 99% cases compilers will fire segfault in case of memory violation but it is result of UB which I am talking about. If we can skip UB and prevent it with regular C++ exception which can be caught in contrast to segfault better prevent it

fnc12: > But I have a completely different suggestion: we don't need the check at all. The iterator exists perfectly without the view, it is self-sufficient, so to speak, and the ‘storage’ together with the ‘db objects’ is a global instance anyway.

Please apply your suggestion - it sounds good

dev/iterator.h
```diff
@@ -1,101 +1,120 @@
...
-            bool operator==(const iterator_t& other) const {
-                return this->current == other.current;
+            friend bool operator==(const iterator_t& lhs, const iterator_t& rhs) {
+                return lhs.current == rhs.current;
             }
 
-            bool operator!=(const iterator_t& other) const {
-                return !(*this == other);
+#ifndef SQLITE_ORM_DEFAULT_COMPARISONS_SUPPORTED
+            friend bool operator!=(const iterator_t& lhs, const iterator_t& rhs) {
```
fnc12, line 114: Are you sure we need `friend` here as well? We don't access `current` field here so we can skip `friend`

fnc12 (approved):

dev/iterator.h
trueqbit: Yes, we do need `friend` as it is a global operator...

PR #1334, opened 2024-07-14
Return underlying 64-bit integer value for last inserted row id

trueqbit:
For the high-level insertion API:
* instead of narrowing the actual return value of `sqlite3_last_insert_rowid()` to `int`, `int64` is now returned.
* inserting a range of objects also returns the last insert rowid, or 0 if the range is empty.

### DIFF
--- a/dev/storage.h
+++ b/dev/storage.h
@@ -4,7 +4,7 @@
 #include <memory>  //  std::unique_ptr/shared_ptr, std::make_unique/std::make_shared
 #include <system_error>  //  std::system_error
 #include <string>  //  std::string
-#include <type_traits>  //  std::remove_reference, std::is_base_of, std::decay, std::false_type, std::true_type
+#include <type_traits>  //  std::remove_reference, std::is_base_of, std::remove_cvref, std::decay, std::false_type, std::true_type
 #include <functional>  //   std::identity
 #include <sstream>  //  std::stringstream
 #include <map>  //  std::map
@@ -117,7 +117,7 @@ namespace sqlite_orm {
 
             template<class Table>
             void create_table(sqlite3* db, const std::string& tableName, const Table& table) {
-                using table_type = std::decay_t<decltype(table)>;
+                using table_type = polyfill::remove_cvref_t<decltype(table)>;
                 using context_t = serializer_context<db_objects_type>;
 
                 context_t context{this->db_objects};
@@ -802,7 +802,7 @@ namespace sqlite_orm {
                 std::stringstream ss;
                 ss << "{ ";
                 table.for_each_column([&ss, &object, first = true](auto& column) mutable {
-                    using column_type = std::decay_t<decltype(column)>;
+                    using column_type = polyfill::remove_cvref_t<decltype(column)>;
                     using field_type = typename column_type::field_type;
                     constexpr std::array<const char*, 2> sep = {", ", ""};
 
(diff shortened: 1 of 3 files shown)


dev/storage.h
```diff
@@ -852,11 +852,11 @@ namespace sqlite_orm {
             }
 
             template<class O, class... Cols>
-            int insert(const O& o, columns_t<Cols...> cols) {
-                static_assert(cols.count > 0, "Use insert or replace with 1 argument instead");
+            int64 insert(const O& o, columns_t<Cols...> cols) {
```
fnc12, line 855: I'd leave it like that cause changing return type of public API which exists 6+ years will spawn a lot of warnings

dev/storage.h
```diff
@@ -865,11 +865,11 @@ namespace sqlite_orm {
              *  @return id of just created object.
              */
             template<class O>
-            int insert(const O& o) {
+            int64 insert(const O& o) {
```
fnc12, line 868: I'd leave it like that cause changing return type of public API which exists 6+ years will spawn a lot of warnings. `int` is the most popular type and the first type every developer uses in the first C++ program. And also for 7+ years nobody complained about `int` here. Who needs the highest precision just uses `insert` with `last_insert_rowid` calls together. Now one person decided it is wrong and we are adding a lot of warning to all other projects. This is a bad practice. The better solution is to leave this public API as is and tell that one person what to do instead.

dev/storage.h
```diff
@@ -852,11 +852,11 @@ namespace sqlite_orm {
...
 
             template<class O, class... Cols>
-            int insert(const O& o, columns_t<Cols...> cols) {
-                static_assert(cols.count > 0, "Use insert or replace with 1 argument instead");
+            int64 insert(const O& o, columns_t<Cols...> cols) {
+                static_assert(cols.count > 0, "Specify at least one column");
                 this->assert_mapped_type<O>();
                 auto statement = this->prepare(sqlite_orm::insert(std::ref(o), std::move(cols)));
-                return int(this->execute(statement));
+                return this->execute(statement);
```
fnc12, line 859: I understand that this is legacy but it doesn't mean that all my arguments have no meaning

dev/storage.h
```diff
@@ -907,9 +907,9 @@ namespace sqlite_orm {
              *  ```
              */
             template<class... Args>
-            void insert(Args... args) {
+            int64 insert(Args... args) {
```
fnc12, line 910: there is one interesting feature in SQLite called `RETURNING` https://www.sqlite.org/lang_returning.html. @trueqbit let's imagine how we are going to implement this. E.g. this query is expected to return `void`
```c++
storage.insert(into<User>(), 
    columns(&User::id, &User::name),
    values(std::make_tuple(1, std::string("Ellie"))));
```
But this query
```c++
auto insertedUser = storage.insert(into<User>(), 
    columns(&User::id, &User::name),
    values(std::make_tuple(1, std::string("Ellie"))),
    returning(asterisk());
```
is expected to return  a `User`. And this query
```c++
auto insertedId = storage.insert(into<User>(), 
    columns(&User::id, &User::name),
    values(std::make_tuple(1, std::string("Ellie"))),
    returning(&User::id);
```
is expected to return inserted id **no matter what type id has**. And `RETURNING` feature will just erase this diff. I know I did not tell you about this feature idea but know I am doing this and looks like I am doing this right in time.

dev/storage.h
```diff
@@ -947,28 +947,28 @@ namespace sqlite_orm {
             }
 
             template<class It, class Projection = polyfill::identity>
-            void insert_range(It from, It to, Projection project = {}) {
+            int64 insert_range(It from, It to, Projection project = {}) {
```
fnc12, line 950: It makes no sense why inserting a range has to return inserted id. I mean who needs to insert let's say 10 objects and then expect to get latest object's id. `RETURNING` feature is more appropriate here and once it is added your diff in this function will just removed. So let's just skip changing a lot of public APIs cause one person wants it. It is related to both `insert_range` overloads

fnc12 (changes requested):

Issue #1332, opened 2024-07-05
Why not reurn int64 for insert method?


shuai132:
I have seen at https://github.com/fnc12/sqlite_orm/issues/551, 

`storage.last_insert_rowid()` function is a workaround way, but it will call sqlite3_last_insert_rowid() once (even it can be ignored).

why not return int64 type directly?

fnc12:
> Why not reurn int64 for insert method?

cause `int` is a regular default and the most popular type. Also regular `int` covers most ids of most SQLite DBs. If `int` is not enough then use `storage.last_insert_rowid()`

shuai132:
Correctly, `int` on most 32bit **and** 64bit platform is 32bit,  it is enough usually: 2147 483 647 ≈ 2147M items ≈ 2G items.

Thanks

trueqbit:
I don't quite understand why `int` was chosen as the "most popular type".
  It doesn't cover the value range that's being returned by the underlying SQLite API, and sqlite_orm [silently casts](https://github.com/fnc12/sqlite_orm/blob/64342a09b4d94ad3c2f331d7889d09c08f37026e/dev/storage.h#L872) an `int64` to `int`. Of course things are hypothetical, and there's a workaround, but then it raises questions and assumes a certain functionality of SQLite and a certain use of SQLite by programmers.

I think sqlite_orm as a library should be accurate here and it is easy to implement.
sqlite_orm should just forward whatever is returned by `sqlite3_last_insert_rowid()`, and [`storage_t<>::execute()`](https://github.com/fnc12/sqlite_orm/blob/64342a09b4d94ad3c2f331d7889d09c08f37026e/dev/storage.h#L1455) does this already.

shuai132:
I suddenly thought of some situation:

1. When a program runs for a very long time with numerous `add and delete` operations, the `id` easily increases beyond 2G.

2. The `storage.last_insert_rowid()` workaround will be incorrect in a multi-thread environment if someone forget **locking**.

Therefore, I think it's better to use the `int64` type in new project(if not very small).
I have seen the discussion on PR: https://github.com/fnc12/sqlite_orm/pull/1334.
For avoid API break change or bring warnings to old users, we can make `int64` configurable and use `int` as default. 

``` cpp
#ifdef  SQLITE_ORM_INSERT_RETURN_TYPE_INT64
#define SQLITE_ORM_INSERT_RETURN_TYPE int64
#else
[...]

PR #1335, opened 2024-07-16
Support int64 type for insert api #1332

shuai132:


fnc12:
hey @shuai132 . Thanks for the PR. This PR is a good candidate for `sqlite_orm`'s fork but it is not going to be merged cause adding one more macro multiplies all support cases by 2 which has to be supported by us. On the other hand the user has to not forget to use exclusive transaction. So it is not a good idea to merge such a feature. But you still can make your own `sqlite_orm`'s fork with this feature and use it with no regrets

shuai132:
Ok, Thanks!

PR #1336, opened 2024-07-18
sqlite_orm as a C++ named module

trueqbit:
sqlite_orm can be imported as a C++ named module now.

* The named module is exported by the the module interface unit file _modules/sqlite_orm.ixx_ [msvc], _modules/sqlite_orm.cppm_ [clang], _modules/sqlite_orm.cxx_ [gcc].
* Currently the sqlite_orm module in turn consumes the STL as a module, too, via `import std.compat;`. This means that the C++23 standard named modules `std` and `std.compat` must be available.
* The CMake scripts install _modules/sqlite_orm.*_ into _${CMAKE_INSTALL_LIBDIR}/cmake/SqliteOrm/src_ (following Craig Scott's [writeup](https://crascit.com/2024/04/04/cxx-modules-cmake-shared-libraries/#h-installing-shared-libraries-with-c-20-modules)).
* There's a new test target named "modules_test". It builds the sqlite_orm module and simply imports it in a bare .cpp file, in order to test whether those steps work. Otherwise no specific tests are executed.
* C++ named modules require support from build systems. At the time of writing, CMake only provides experimental support for the dependency on the standard library modules [see [CMAKE_EXPERIMENTAL_CXX_IMPORT_STD](https://www.kitware.com/import-std-in-cmake-3-30/)].
  -> Therefore, the test target “modules_test” is only defined when testing with Visual C++.
  -> Therefore, the named module still has to be consumed manually in a CMake script.

To discuss:
* Because C++ named modules don't export macros, I think we need to provide sqlite_orm feature test macros in a traditional dedicated header file. These macros are e.g. used in examples to decide whether an example should be built, which is a public facing part of sqlite_orm [e.g. SQLITE_ORM_WITH_CTE, SQLITE_ORM_OPTIONAL_SUPPORTED].


fnc12 (approved):

kelteseth:
Nice! Please create a PR when this is done at https://arewemodulesyet.org/ :D

dev/cte_storage.h
```diff
@@ -61,15 +63,15 @@ namespace sqlite_orm {
 
         // aliased column expressions, explicit or implicitly numbered
         template<typename F, typename ColRef, satisfies_is_specialization_of<ColRef, alias_holder> = true>
-        static auto make_cte_column(std::string name, const ColRef& /*finalColRef*/) {
```
fnc12, line 64: why not static anymore?

dev/functional/cstring_literal.h
```diff
@@ -1,15 +1,17 @@
...
 #include <algorithm>  //  std::copy_n
 #endif
+#endif
 
 #ifdef SQLITE_ORM_WITH_CPP20_ALIASES
 namespace sqlite_orm::internal {
-    /*
-     *  Wraps a C string of fixed size.
-     *  Its main purpose is to enable the user-defined string literal operator template.
+    /*
```
fnc12: what is this?

dev/functional/cxx_functional_polyfill.h
```diff
@@ -1,10 +1,14 @@
 #pragma once
 
+#ifdef SQLITE_ORM_IMPORT_STD_MODULE
+#include <version>
```
fnc12, line 4: why do we need version here?

dev/cte_storage.h
trueqbit: I don't remember the specifics anymore.
However, there's no need for a free-standing function to be static.

dev/functional/cstring_literal.h
trueqbit: That's a '\r' character - mac-style newline ending. No idea where that comes from!

dev/functional/cxx_functional_polyfill.h
trueqbit: C++ named modules do not export macros, including standardized feature-test macros.

PR #1367, opened 2024-12-17
first implementation of sql_auditing

juandent:


juandent:
I don't know what went wrong with formatting

fnc12:
@juandent probably clang-format version is incorrect (use 17) but maybe something else. I can't say cause I it happened on your machine and you have to fix it no matter the reason

fnc12:
@juandent you need to fix formatting

juandent:
> @juandent you need to fix formatting

the error seems to be in .clang-format file
can you send me this file?

fnc12:
@juandent the error is not in `clang-format` file cause all other PR creator have it working. The file is located in repo root

PR #1370, opened 2024-12-20
Feature/sql auditing

juandent:
First implementation of sql_auditing -- testing use of it

### DIFF
--- a/dev/sql_auditing.h
+++ b/dev/sql_auditing.h
@@ -1,0 +1,86 @@
+#pragma once
+#include <fstream>
+#include "error_code.h"
+#include <chrono>
+#include <ctime>  // For std::localtime and std::tm
+#include <iomanip>  // For std::put_time
+#include <string>
+
+enum class auditing_behavior : signed char { OFF = 0, ON = 1 };
+
+class sql_auditor_settings;
+
+class sql_auditor {
+
+    std::ofstream log_file;
+    friend class sql_auditor_settings;
+    sql_auditor();
+    void open();
+    inline static sql_auditor& auditor() {
+        static sql_auditor auditor{};
+        return auditor;
+    }
+
+  public:
+    static void log(const std::string& message);
+};
+
+class sql_auditor_settings {
+    std::string destination_file = "sql_auditor.txt";
+    friend class sql_auditor;
+    sql_auditor_settings() {}
+    inline static sql_auditor_settings& settings() {
+        static sql_auditor_settings sql_settings;
+        return sql_settings;
+    }
+    auditing_behavior behavior = auditing_behavior::ON;
+    std::string format_str = "%Y-%m-%d %H:%M:%S";
+
+  public:
+    static void set_destination_file(const std::string& filename);
+    static void set_behavior(auditing_behavior behavior) {
+        settings().behavior = behavior;
+    }
+    static void set_format(const std::string& format_str) {
+        settings().format_str = format_str;
+    }
+};
+
+inline sql_auditor::sql_auditor() {
+    open();
+}
+
+inline void sql_auditor_settings::set_destination_file(const std::string& filename) {
+    settings().destination_file = filename;
+    sql_auditor::auditor().open();
+}
+
+inline void sql_auditor::open() {
+    using namespace std;
+    if(!log_file.is_open()) {
+        log_file.open(sql_auditor_settings::settings().destination_file, ios::trunc | ios::out);
+        if(!log_file.good()) {
+            throw std::system_error{sqlite_orm::orm_error_code::failure_to_init_logfile};
+        }
+    }
+}
+
+inline void sql_auditor::log(const std::string& message) {
+    // guard and exit if off
+    if(sql_auditor_settings::settings().behavior == auditing_behavior::OFF)
+        return;
+
+    // would use format if C++ 20
+    auto now = std::chrono::system_clock::now();
+
+    std::time_t now_time = std::chrono::system_clock::to_time_t(now);
+
+    // Convert to local time (std::tm structure)
+    // WARNING: localtime is not thread safe!
+    std::tm local_time = *std::localtime(&now_time);
+
+    // Print the local time in a human-readable format
+    auditor().log_file << "@: " << std::put_time(&local_time, sql_auditor_settings::settings().format_str.c_str())
+                       << " = ";
+    auditor().log_file << message << std::endl;
+}
--- a/tests/prepared_statement_tests/select.cpp
+++ b/tests/prepared_statement_tests/select.cpp
@@ -14,6 +14,14 @@ TEST_CASE("Prepared select") {
 
     auto filename = "prepared.sqlite";
     remove(filename);
+
+#define JD_AUDITING_SETTINGS
+#ifdef JD_AUDITING_SETTINGS
+    sql_auditor_settings::set_destination_file("log.txt");
+    sql_auditor_settings::set_behavior(auditing_behavior::ON);
+    sql_auditor_settings::set_format("%H:%M:%S");
+#endif
+
     auto storage = make_storage(filename,
                                 make_index("user_id_index", &User::id),
                                 make_table("users",
--- a/dev/error_code.h
+++ b/dev/error_code.h
@@ -39,6 +39,7 @@ namespace sqlite_orm {
         index_is_out_of_bounds,
         value_is_null,
         no_tables_specified,
+        failure_to_init_logfile,
     };
 }
 
--- a/dev/util.h
+++ b/dev/util.h
@@ -5,6 +5,7 @@
 #include <utility>  //  std::move
 
 #include "error_code.h"
+// Before clang-format 17
 
 namespace sqlite_orm {
 
--- a/dev/util.h
+++ b/dev/util.h
@@ -5,7 +5,8 @@
 #include <utility>  //  std::move
 
 #include "error_code.h"
-// Before clang-format 17
+
+#include "sql_auditing.h"
 
 namespace sqlite_orm {
 
@@ -62,6 +63,7 @@ namespace sqlite_orm {
             if(sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr) != SQLITE_OK) {
                 throw_translated_sqlite_error(db);
             }
+            sql_auditor::log(query);
             return stmt;
         }
 
@@ -70,6 +72,7 @@ namespace sqlite_orm {
             if(rc != SQLITE_OK) {
                 throw_translated_sqlite_error(db);
             }
+            sql_auditor::log(query);
         }
 
         inline void perform_exec(sqlite3* db,
@@ -80,6 +83,7 @@ namespace sqlite_orm {
             if(rc != SQLITE_OK) {
                 throw_translated_sqlite_error(db);
             }
+            sql_auditor::log(query);
         }
 
         inline void perform_exec(sqlite3* db,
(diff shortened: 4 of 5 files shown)


dev/sql_auditing.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+#include <fstream>
+#include "error_code.h"
+#include <chrono>
+#include <ctime>  // For std::localtime and std::tm
```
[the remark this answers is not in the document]
juandent: it really is std::tm but also std::time_t
```
   std::tm local_time = *std::localtime(&now_time);
```

dev/sql_auditing.h
```diff
@@ -0,0 +1,86 @@
...
+
+  public:
+    static void log(const std::string& message);
+};
+
+class sql_auditor_settings {
+    std::string destination_file = "sql_auditor.txt";
+    friend class sql_auditor;
+    sql_auditor_settings() {}
+    inline static sql_auditor_settings& settings() {
```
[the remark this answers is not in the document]
juandent: the destination file could be part of the storage object but then we could not log the initial instructions (i.e. make_storage) -- what if we want to see the results of make_storage()?

dev/sql_auditing.h
```diff
@@ -0,0 +1,86 @@
...
+    inline static sql_auditor& auditor() {
+        static sql_auditor auditor{};
+        return auditor;
+    }
+
+  public:
+    static void log(const std::string& message);
+};
+
+class sql_auditor_settings {
```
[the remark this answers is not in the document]
juandent: ok - one question about namespace: what to use here? namespace sqlite_orm or global namespace?

dev/sql_auditing.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+#include <fstream>
+#include "error_code.h"
```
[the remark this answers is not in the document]
juandent: ok

tests/prepared_statement_tests/select.cpp
```diff
@@ -14,6 +14,14 @@ TEST_CASE("Prepared select") {
 
     auto filename = "prepared.sqlite";
     remove(filename);
+
+#define JD_AUDITING_SETTINGS
```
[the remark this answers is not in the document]
juandent: its the way I mark tentative code...
experimental code...

dev/sql_auditing.h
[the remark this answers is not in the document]
juandent: since the call to sql_auditor::log() occurs in util.h, how do we know which storage it belongs to? I am talking about e.g. void perform_exec()??? we only know the sqlite3* but that does not map to storage so its no use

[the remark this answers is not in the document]
juandent: we could intercept on_open to register every connection used by a storage_base?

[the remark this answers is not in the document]
juandent: and create a map of sqlite* pointers and storage_base* pointers!!

dev/sql_auditing.h
```diff
@@ -0,0 +1,146 @@
...
+
+enum class auditing_behavior : signed char { OFF = 0, ON = 1 };
+
+namespace sqlite_orm {
+    namespace internal {
+        class storage_base;
+    }
+}
+using sqlite_orm::internal::storage_base;
```
fnc12, line 19: why this `using` is here?

dev/sql_auditing.h
```diff
@@ -0,0 +1,146 @@
...
+        behavior = behave;
+        log_file = file;
+        bool ok = log_file->good();
+        bool open = log_file->is_open();
+        log_file_name = file_name;
+        time_format = format;
+        status = stats;
+        global_settings = glob_settings;
+    }
+    log_settings() {}
```
fnc12, line 45: why do we need default constructor here?

dev/sql_auditing.h
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+using sqlite_orm::internal::storage_base;
+
+struct log_settings {
+    auditing_behavior behavior;
+    std::shared_ptr<std::ofstream> log_file;
+    std::string log_file_name;
+    std::string time_format;
+    enum status_code { FAIL, GOOD } status = status_code::FAIL;
```
fnc12, line 26: avoid using non-class enums cause old enums add their constants into global scope

dev/sql_auditing.h
```diff
@@ -0,0 +1,146 @@
...
+
+            // Print the local time in a human-readable format
+            *log_file << "@: " << std::put_time(&local_time, time_format.c_str()) << " = ";
+            *log_file << message << std::endl;
+        }
+    }
+};
+
+inline std::shared_ptr<std::ofstream> get_file(std::string log_file_name) {
+    static std::map<std::string, std::shared_ptr<std::ofstream>> files;
```
fnc12, line 77: why do we need to keep file streams in a static map? better avoid static vars at all cause it is not thread safe

dev/sql_auditing.h
```diff
@@ -0,0 +1,146 @@
...
+    auto it = files.find(log_file_name);
+    if(it == files.end()) {
+        std::shared_ptr<std::ofstream> pointer_to_shared{new std::ofstream()};
```
fnc12, line 80: use `std::make_shared` instead of `new`. Avoid using `new` and `delete` if you can

dev/storage_base.h
```diff
@@ -766,6 +766,9 @@ namespace sqlite_orm {
                 if(this->on_open) {
                     this->on_open(db);
                 }
+                // JD added
```
fnc12, line 769: this comment is useless cause git blame has info about authors of every line

dev/storage_base.h
```diff
@@ -766,6 +766,9 @@ namespace sqlite_orm {
...
+                log_admin::admin().register_db(this, db);
+                // JD end
```
fnc12, line 771: same here: this comment is useless cause git blame has info about authors of every line

fnc12 (changes requested):
architecture of current implementation of sql auditing can't be combined with the project's architecture which (projects) have logs system like boost::log or custom written ones. How this approach is going to wok in projects which use boost::log for example?

dev/sql_auditing.h
juandent: using is used here because log_settings and log_admin are declared in the glocal namespace

dev/sql_auditing.h
juandent: ok

dev/sql_auditing.h
juandent: __the other alternative would be a global object;....

dev/storage_base.h
juandent: ok

dev/storage_base.h
juandent: go

dev/sql_auditing.h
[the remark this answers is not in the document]
fnc12: > What do you think? is storage_base the place to put the sql_auditor?

I don't think we need `sql_auditor` at all but it is up to you

dev/sql_auditing.h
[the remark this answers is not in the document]
fnc12: don't use global namespace in the lib cause it may lead to collisions in the user code

dev/sql_auditing.h
fnc12: don't use global namespace in the lib code

dev/sql_auditing.h
fnc12: don't use global/static objects cause it it not thread safe and break the architecture. Global object is anti-pattern

juandent:
Why do they break the architecture?Sent from my iPhoneOn 5 Jan 2025, at 4:07 AM, Yevgeniy Zakharov ***@***.***> wrote:﻿
@fnc12 commented on this pull request.



In dev/sql_auditing.h:
> +
+            std::time_t now_time = std::chrono::system_clock::to_time_t(now);
+
+            // Convert to local time (std::tm structure)
+            // WARNING: localtime is not thread safe!
+            std::tm local_time = *std::localtime(&now_time);
+
+            // Print the local time in a human-readable format
+            *log_file << "@: " << std::put_time(&local_time, time_format.c_str()) << " = ";
+            *log_file << message << std::endl;
+        }
+    }
+};
+
+inline std::shared_ptr<std::ofstream> get_file(std::string log_file_name) {
+    static std::map<std::string, std::shared_ptr<std::ofstream>> files;

don't use global/static objects cause it it not thread safe and break the architecture. Global object is anti-pattern

fnc12:
> Why do they break the architecture?

Cause global object (singleton) is anti-pattern. If we can avoid it we have to avoid it

juandent:
Which is the global/static object you refer to?Sent from my iPhoneOn 5 Jan 2025, at 10:28 AM, Juan Dent ***@***.***> wrote:﻿Why do they break the architecture?Sent from my iPhoneOn 5 Jan 2025, at 4:07 AM, Yevgeniy Zakharov ***@***.***> wrote:﻿
@fnc12 commented on this pull request.



In dev/sql_auditing.h:
> +
+            std::time_t now_time = std::chrono::system_clock::to_time_t(now);
+
+            // Convert to local time (std::tm structure)
+            // WARNING: localtime is not thread safe!
+            std::tm local_time = *std::localtime(&now_time);
+
+            // Print the local time in a human-readable format
+            *log_file << "@: " << std::put_time(&local_time, time_format.c_str()) << " = ";
+            *log_file << message << std::endl;
+        }
+    }
+};
+
+inline std::shared_ptr<std::ofstream> get_file(std::string log_file_name) {
+    static std::map<std::string, std::shared_ptr<std::ofstream>> files;

don't use global/static objects cause it it not thread safe and break the architecture. Global object is anti-pattern

juandent:
But I am using singleton patternSent from my iPhoneOn 5 Jan 2025, at 10:30 AM, Yevgeniy Zakharov ***@***.***> wrote:﻿

Why do they break the architecture?

Cause global object (singleton) is anti-pattern. If we can avoid it we have to avoid it

fnc12:
> Which is the global/static object you refer to?

The one defined as static in a function and returns as a ref

> But I am using singleton pattern

This comment doesn't make any sense

fnc12:
@juandent you are still ignoring this comment

> architecture of current implementation of sql auditing can't be combined with the project's architecture which (projects) have logs system like boost::log or custom written ones. How this approach is going to wok in projects which use boost::log for example?

It is the main issue with this PR. No reason to spend time on talking about singleton and other stuff until this one is solved

juandent:
“But I am using singleton pattern”

I am following Dimitri Nesteruk’  “Design patterns in Modern C++” and Fedor Pikus’ “Hands-on design patterns with C++” both of which implement a safe way for 

Singletons.

 

So why so negative about the correct use of singletons?

 

Regards,

Juan

fnc12:
> I am following Dimitri Nesteruk’  “Design patterns in Modern C++” and Fedor Pikus’ “Hands-on design patterns with C++” both of which implement a safe way for singletons.

Good for you. Still it is not acceptable for `sqlite_orm` cause singleton is anti-pattern

juandent:
This is no way to react!

“Good for you” you say! Can you be more sarcastic?

fnc12:
> This is no way to react!
> “Good for you” you say! Can you be more sarcastic?

I am not sarcastic at all. Please don't ignore my other comments

fnc12:
Discussed in DM. Closing

PR #1380, opened 2025-01-25
Corrected rowset deduplicators DISTINCT and ALL

trueqbit:
* Fixed the missing ability to write `distinct(struct_<>())`.
* Corrected serialization of DISTINCT/ALL: They don't introduce a parenthesized context.
* Added Unit Tests.
* Moved rowset deduplicators to the syntax tree. Previously, the DISTINCT/ALL result set deduplicators were stored as a member flag (i.e. runtime value) in `columns_t` and `struct_t`. I moved it to the AST because it is part of the logical structure of an SQL select expression.
* The previous point also fixes collecting column names.


dev/statement_serializer.h
```diff
@@ -1788,10 +1782,14 @@ namespace sqlite_orm {
                         ss << "(";
                     }
                     ss << "SELECT ";
+                    call_if_constexpr<is_rowset_deduplicator_v<T>>(
```
fnc12, line 1785: nit: probably it would be more readable if we replace `T` here with `statement_type::return_type` (e.g. we can declare `using return_type = typename statement_type::return_type` just below `using statement_type`

tests/operators/glob.cpp
```diff
@@ -78,19 +78,19 @@ TEST_CASE("Glob") {
...
         REQUIRE(rows.size() == 9);
-        auto trueValuesCount = std::count(rows.begin(), rows.end(), true);
+        auto trueValuesCount = count(rows.begin(), rows.end(), true);
         REQUIRE(trueValuesCount == 3);
     }
     {
-        auto rows = storage.select(glob(distinct(&Employee::firstName), "S*"));
+        auto rows = storage.select(distinct(glob(&Employee::firstName, "S*")));
         REQUIRE(rows.size() == 2);
-        auto trueValuesCount = std::count(rows.begin(), rows.end(), true);
```
fnc12, line 87: why did you remove `std::`?

tests/statement_serializer_tests/statements/select.cpp
```diff
@@ -235,6 +235,50 @@ TEST_CASE("statement_serializer select_t") {
...
+            }
+            SECTION("all struct") {
+                auto expression = select(all(struct_<User>(&User::name)));
+                expression.highest_level = true;
+                stringValue = serialize(expression, context);
+                expected = R"(SELECT ALL "users"."name" FROM "users")";
+            }
+            SECTION("distinct aggregate function") {
+                auto expression = select(count(distinct(&User::name)));
+                expression.highest_level = true;
```
fnc12, line 277: probably we also need to add the same tests with `highest_level = false`

fnc12 (approved):

tests/operators/glob.cpp
trueqbit: ADL. It's the same with `find_if` somes lines above that you used.

PR #1387, opened 2025-01-29
Using range-based `for` with init-statement where possible

trueqbit:
C++20 features a range-based `for` loop with a so-called init-statement. It helps sqlite_orm to dispense with index-based loops in certain contexts.

Additionally:
PR #1311 introduced an internal, universally available, distilled feature test macro `SQLITE_ORM_CPP20_RANGES_SUPPORTED` to replace `__cpp_lib_ranges >= 201911L`.


dev/pragma.h
```diff
@@ -33,8 +33,7 @@ namespace sqlite_orm {
             res.reserve(argc);
             const auto rowExtractor = column_text_extractor<std::string>();
             for (int i = 0; i < argc; ++i) {
-                auto rowString = rowExtractor.extract(argv[i]);
-                res.push_back(std::move(rowString));
+                res.push_back(rowExtractor.extract(argv[i]));
```
fnc12, line 36: you make it harder to debug =)

examples/any.cpp
```diff
@@ -3,7 +3,7 @@
  *  It is a bit more complex than the other examples, because it
  *  uses custom row_extractor and statement_binder.
  *  Please note that this implementation is not the one and only
- *  option of implementation of std:any bindong to sqlite_orm.
+ *  option of implementation of `std:any` binding to sqlite_orm.
```
fnc12, line 6: thank you!

fnc12 (approved):

dev/pragma.h
trueqbit: Well, nice and terse I thought. Shall I revert it?

fnc12: No it is not so critical

PR #1388, opened 2025-01-29
 C++20 defaulted comparisons

trueqbit:
Use C++20's feature of defaulted comparisons where possible.

Additionally:
* Runner-up: Qualified all calls to `std::remove()` with `std`.
* Included the C header `assert.h` instead of the C++ header `cassert`. This will be important when importing the Standard Library as a named module.

### DIFF
--- a/dev/udf_proxy.h
+++ b/dev/udf_proxy.h
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <sqlite3.h>
-#include <cassert>  //  assert macro
+#include <assert.h>  //  assert macro
 #include <type_traits>  //  std::true_type, std::false_type
 #include <new>  //  std::bad_alloc
 #include <memory>  //  std::allocator, std::allocator_traits, std::unique_ptr
--- a/dev/arg_values.h
+++ b/dev/arg_values.h
@@ -98,9 +98,11 @@ namespace sqlite_orm {
                 return &other.container == &this->container && other.index == this->index;
             }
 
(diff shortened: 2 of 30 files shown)


dev/udf_proxy.h
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <sqlite3.h>
-#include <cassert>  //  assert macro
+#include <assert.h>  //  assert macro
```
fnc12, line 4: why? I was sure that C++ style for C includes is preferred in C++

dev/udf_proxy.h
```diff
@@ -13,9 +13,9 @@
 
 namespace sqlite_orm {
     namespace internal {
-        /*
-         *  Returns properly allocated memory space for the specified application-defined function object
-         *  paired with an accompanying deallocation function.
+        /*
+         *  Returns properly allocated memory space for the specified application-defined function object
+         *  paired with an accompanying deallocation function.
```
fnc12, line 18: I'd like to know what changed here

fnc12 (approved):

dev/udf_proxy.h
trueqbit: New-line endings. You might want to enable the option to hide whitespace when comparing the diff.

dev/udf_proxy.h
trueqbit: Not with macros. If one is importing the Standard Library as a module you need to include the C header to get macros.

PR #1393, opened 2025-02-01
#1392 : Add support for SQLite VFS and open mode flags

### LINKED ISSUE #1392: Add support for SQLite Virtual File System (VFS) Options
My team and I develop audio plugins, which are linked dynamically into plugin hosts. However, because there are multiple plugin formats which necessitates creating different dynamically linked libraries, our code immediately becomes vulnerable to [this known issue with the default locking mechanism in SQLite](https://www.sqlite.org/draft/howtocorrupt.html#multiple_copies_of_sqlite_linked_into_the_same_application). This linking approach is unavoidable as it is the industry standard, and we've determined the only way to address the issue is to use the [`unix-dotfile`](https://www.sqlite.org/vfs.html#standard_unix_vfses) SQLite VFS. However, `sqlite_orm` does not provide any methodology for using the necessary `sqlite_open_v2` API to allow for VFS selection.

I've already confirmed this fix works for us by modifying a fork of `sqlite_orm` and adding the changes myself. I am opening a PR shortly, and will un-draft it once I feel that the code is cleaned up and in line with the style of this library.

ancientjpeg:
Addresses #1392 

## Changelog
- Add `vfs_mode` enum class to simplify the `sqlite_orm` API for using different VFS systems, as compared to the raw string passed in `sqlite3.h`
- Add `open_mode` enum class to support the open flags passed by `sqlite3_open_v2`. Only two values have been added so far, but this addition makes it very easy to add more in the future.
- Add platform definitions to detect operating system
   - Necessary for exposing platform-specific VFS types
- Add tests

## More Info
https://www.sqlite.org/vfs.html - VFS options and description
https://www.sqlite.org/c3ref/open.html - open flag options


fnc12:
hey @ancientjpeg . Thank for this PR. Please work in `dev` branch not `master`

ancientjpeg:
Just rebased. Thanks and apologies again.

dev/open_mode.h
```diff
@@ -0,0 +1,25 @@
...
+
+#pragma once
+
+namespace sqlite_orm {
+
+    enum class open_mode {
+        default_mode = 0,
+        create_readwrite = 0,
+        readonly = 1,
+    };
```
ancientjpeg: I've added very limited options for now—this can be modified very easily in the future.

dev/storage_base.h
```diff
@@ -639,6 +641,29 @@ namespace sqlite_orm {
...
+                return this->connection->open_mode;
+            }
+
+            /**
+             * Return true if this database object is opened in a readonly state. 
+             */
+            bool readonly() const {
+                sqlite3* db = this->connection->get();
+                return sqlite3_db_readonly(db, "main");
+            }
```
ancientjpeg: At first I was a little nervous about using the `"main"` DB name here but then I realized that's used in other places in the lib. I figure this indicates there's no near-future plans for attachment support.

ancientjpeg (commented):
@fnc12 I've added enums to allow for full control over what open mode/VFS options users actually have access to. I'm happy with the state this PR is in, but I have some preliminary evidence that using some VFS modes like `unix-dotfile` may interfere with the ability to set certain `journal_mode`s. This is detectable by just checking `journal_mode` after you've set it, but if you'd like me to brainstorm a layer of protection for that I'd be happy to.

dev/open_mode.h
```diff
@@ -0,0 +1,27 @@
+#include <sqlite3.h>
+
+#pragma once
```
trueqbit: Can you move up the `#pragma once` directives to the first line? Just because that's how it is in the rest of the code base.

dev/open_mode.h
```diff
@@ -0,0 +1,27 @@
...
+namespace sqlite_orm {
+
+    enum class open_mode {
+        default_mode = 0,
+        create_readwrite = 0,
+        readonly = 1,
+    };
+
+    namespace internal {
+        inline constexpr int to_int_flags(sqlite_orm::open_mode open) {
```
trueqbit: `constexpr` qualifier is implicitly `inline`.

dev/open_mode.h
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+namespace sqlite_orm {
+
+    enum class open_mode {
+        default_mode = 0,
+        create_readwrite = 0,
+        readonly = 1,
+    };
+
```
trueqbit: Can you please close the `sqlite_orm` namespace here. This will be important for the upcoming 'modules' PR.

dev/storage.h
```diff
@@ -1705,9 +1710,18 @@ namespace sqlite_orm {
     /*
      *  Factory function for a storage, from a database file and a bunch of database object definitions.
      */
+    template<class... DBO>
+    internal::storage_t<DBO...>
+    make_storage(std::string filename, sqlite_orm::vfs_mode vfs, sqlite_orm::open_mode open, DBO... dbObjects) {
```
trueqbit: @fnc12 @ancientjpeg I suggest assembling possible configuration/option parameters into their own struct. With C++20 designated initializers this becomes very nice using and reading it, plus makes such an option struct extensible. I have very good experiences with this design pattern.

E.g.
```c++
namespace sqlite_orm {
struct storage_options {
    vfs_mode vfs = vfs_mode::default_vfs;
    open_mode open = open_mode::default_mode;
};

template<class... DBO>
internal::storage_t<DBO...>
    make_storage(std::string filename, storage_options opts, DBO... dbObjects) {
}
}
```
Later:
```c++
make_storage("", {.vfs = default_vfs}, make_table(/*...*/));
```

@ancientjpeg No need to repeat the sqlite_orm namespace.

dev/storage_base.h
trueqbit: From where exactly do you draw the conclusion about `"main"`?
Currently it's only the backup functionality that literally uses `"main"`.

At least dealing with the `"temp"` DB is in the works, upon which the attachment functionality could be built.

dev/vfs_mode.h
```diff
@@ -0,0 +1,48 @@
+#include <sqlite3.h>
+#include "functional/config.h"
+
+#pragma once
```
trueqbit: Same comment as above. And the `#pragma once` should certainly not come after including internal headers.

dev/vfs_mode.h
```diff
@@ -0,0 +1,48 @@
...
+#endif
+#endif
+
+#ifdef SQLITE_ORM_WIN
+        win32 = 0,
+        win32_longpath = 1,
+#endif
+
+    };
+
```
trueqbit: Please close the sqlite_orm namespace, same comment as above.

dev/vfs_mode.h
```diff
@@ -0,0 +1,48 @@
...
+    namespace internal {
+        inline const std::string& to_string(vfs_mode v) {
```
trueqbit: Please use a `serialize_result_type`. You see the pattern in _locking_mode.h_ and _journal_mode.h_.
Also I used a dedicated function name there, because it can then be easily used as a projection function; here it could be `vfs_mode_to_string()`.

dev/vfs_mode.h
```diff
@@ -0,0 +1,48 @@
...
+            static std::string res[] = {
```
trueqbit: ... continuing from review comment above:
```c++
#ifdef SQLITE_ORM_STRING_VIEW_SUPPORTED
            static constexpr serialize_result_type idx2str[] = {
#else
            static const serialize_result_type idx2str[] = {
#endif
```

tests/open_v2_tests.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+using namespace sqlite_orm;
+
+static const auto default_table = make_table("users", make_column("id", &User::id, primary_key()));
+
+/******************************************************************************/
+/******************************  STATIC CHECKS  *******************************/
+/******************************************************************************/
+
+TEST_CASE("vfs string conversion returns expected strings") {
```
trueqbit: We have 'static' assertion tests in the _tests/static_tests_ subfolder. Then you can also omit the `STATIC CHECKS` comment.

dev/connection_holder.h
```diff
@@ -41,6 +49,8 @@ namespace sqlite_orm {
             }
 
             const std::string filename;
+            const vfs_mode vfs;
+            const sqlite_orm::open_mode open;
```
trueqbit: I prefer to call the member variable `openFlags` or `openMode`.
Also there's no need to qualify it with the namespace.

dev/storage.h
```diff
@@ -24,6 +24,7 @@
 #include "tuple_helper/tuple_transformer.h"
 #include "tuple_helper/tuple_iteration.h"
 #include "type_traits.h"
+#include "vfs_mode.h"
```
trueqbit: I think you need `#include open_mode.h` as well.

dev/connection_holder.h
```diff
@@ -12,11 +14,17 @@ namespace sqlite_orm {
...
 
-            connection_holder(std::string filename_) : filename(std::move(filename_)) {}
+            connection_holder(std::string filename_, vfs_mode vfs_, sqlite_orm::open_mode open_) :
+                filename(std::move(filename_)), vfs(vfs_), open(open_) {}
 
             void retain() {
                 if (1 == ++this->_retain_count) {
-                    auto rc = sqlite3_open(this->filename.c_str(), &this->db);
+
+                    const std::string vfs_name = internal::to_string(vfs);
```
trueqbit: Should be `const serialize_result_type& vfs_name`, also see comment below for `to_string(vfs_mode v)`.

trueqbit (changes requested):
Thanks @ancientjpeg for your contribution! Here come my 2 cents...

ancientjpeg:
@trueqbit Still working through your changes, but something occurred to me while refactoring. I had some of those unnecessary `sqlite_orm::` qualifiers on the mode types because I actually got a couple compile errors complaining about ambiguous symbols. Would it be alright if I appended `_t` to these flag types? I.e.,

```cpp
enum class vfs_mode_t;
```

Then it'd be a lot clearer to re-use the same verbiage for defining variables, e.g.

```cpp
vfs_mode_t vfs_mode;
open_mode_t open_mode;

struct storage_options {
  vfs_mode_t vfs_mode;
  open_mode_t open_mode;
};
```

Let me know what you think of that.

**Edit:** I ended up implementing this because GCC complains of ambiguity in cases such as `const open_mode open_mode`. I prefer the look of `const open_mode_t open_mode` to `const open_mode open_flags` as I feel the latter might sow confusion, but I'm happy to change it if it feels more in line with the rest of the lib.

dev/storage_base.h
ancientjpeg: I drew the conclusion from the backup code, as it was the only other `sqlite3` method I could find in `sqlite_orm` that was being called with a database name parameter. How would you like me structure this in order to make it more compatible with future attempts to implement `ATTACH`? My first assumption would be to pass in the database name as one of the `storage_options`, but let me know what approach you think is best.

dev/storage.h
ancientjpeg: Agreed, I like that this will the options much more extensible.

dev/vfs_mode.h
ancientjpeg: By 'close' the namespace, I assumed you mean the following:

```cpp
namespace sqlite_orm {
    /* ... */
}

namespace sqlite_orm::internal {
    /* ... */
}
```

I implemented this, but MSVC prevents this from compiling pre-C++17 even though gcc/clang permit it as a language extension. Maybe I misunderstood?

dev/connection_holder.h
ancientjpeg: I opted to just move the entire `options` structure in here instead—this avoid some errors that GCC throws. See [this comment](https://github.com/fnc12/sqlite_orm/pull/1393#issuecomment-2637957485) for more details.

trueqbit:
> @trueqbit Still working through your changes, but something occurred to me while refactoring. I had some of those unnecessary `sqlite_orm::` qualifiers on the mode types because I actually got a couple compile errors complaining about ambiguous symbols. Would it be alright if I appended `_t` to these flag types? I.e.,
> 
> ```c++
> enum class vfs_mode_t;
> ```
> 
> Then it'd be a lot clearer to re-use the same verbiage for defining variables, e.g.
> 
> ```c++
> vfs_mode_t vfs_mode;
> open_mode_t open_mode;
> 
> struct storage_options {
>   vfs_mode_t vfs_mode;
>   open_mode_t open_mode;
> };
> ```
> 
> Let me know what you think of that.

Hmm, that's a funny problem as both are public :)
However, I do not recommend changing the enum type's name. Firstly, it is not recommended in the coding guidelines, and we do not use it in sqlite_orm with one exception.

I personally like camel case for (member) variables, but I agree that's debatable.
A good choice in any case for the variable name is `vfsOption`, `modeOption` or `vfs_option`, `mode_option` as "option" conveys that fact that a specific thing is selected .

While we are at it, I also suggest renaming `vfs_mode` -> `vfs_object`, as SQLite calls the feature Virtual File System Objects.

dev/vfs_mode.h
trueqbit: Simply open the sqlite_orm namespace in a separate line again.

dev/storage_base.h
trueqbit: You can just leave it as is.

ancientjpeg:
@trueqbit Great point about using `option` in the variable name, I think I'll go with that. I'll also change the members and parameters to `camelCase` if that's the preferred styling for this library—I had just assumed `snake_case` was the go-to given the external API is style that way, but now that I look closer I see a lot of parameter names are `camelCase`.

I'll be away from my development system until next week, at which time I'll wrap up these changes. Thanks for the feedback!

dev/storage_base.h
```diff
@@ -639,6 +639,29 @@ namespace sqlite_orm {
                 return this->connection->retain_count() > 0;
             }
 
+            /**
+             * Public method for checking the VFS implementation being used by
+             * this storage object. Mostly useful for debug.
+             */
+            sqlite_orm::vfs_object vfs_object() const {
+                return this->connection->options.vfs_option;
+            }
```
ancientjpeg: Namespacing here seems to be required to prevent errors in GCC. Mirrors existing code:
```cpp
sqlite_orm::journal_mode journal_mode()
```

ancientjpeg:
Re-drafting to address merge conflicts.

trueqbit:
> Re-drafting to address merge conflicts.

Hi @ancientjpeg,

I must apologize for the inconvenience. Independently of you, I have developed a way to pass various storage options to `make_storage()`, in particular connection control options. The important change for you is that there is now a `connection_control` aggregate that you extend with your new options.

Also, I would like to point out to you in advance that an enum to represent a VFS is most likely too restrictive as it is not a fixed set of values. But we can discuss that later.

ancientjpeg:
@trueqbit I've refactored to best match my understanding of your new `connection_control` object. Let me know if this is how you'd expect it to be used.

Also, what exactly were you thinking when you said that an enum is too restrictive for the VFS object? If you meant that you'd like to open up support to the `sqlite3_vfs_register` api, I can adjust the code to account for that. In my mind the best way would be to still use the enum, but add a `vfs_object::custom` enum value, and `connection_control` could be extended to include a `std::string custom_vfs_object` that's only used if the aforementioned enum is passed. I believe this would be a good balance of type safety and extensibility.

dev/vfs_object.h
trueqbit, line 1: I have discussed this with our core team, and our consent is to replace the enum with `inline constexpr` constants, for the following reasons:
* names are completely unrelated, and not enumerable. BTW this is also the STL's modern approach to define constants, e.g. for days of the week of type `chrono::weekday`.
* there's no closed set of constants. Yes, there are stock VFS objects available from SQLite, but other than this we can never encapsulate all possible VFS objects.
* no translation of enum value to string is necessary.

Regarding the type I think it can really be as simple as a string. This is what gets passed to sqlite3_open_v2().

So, examples of these VFS object name constants would be:
```c++
#include "functional/cxx_string_view.h"

namespace sqlite_orm {
    namespace internal {
#ifdef SQLITE_ORM_STRING_VIEW_SUPPORTED
        using string_constant_type = std::string_view;
#else
        using string_constant_type = const char*;
#endif
    }
}

SQLITE_ORM_EXPORT namespace sqlite_orm {
    SQLITE_ORM_INLINE_VAR constexpr string_constant_type default_vfs_name{};
    // SQLITE_ORM_UNIX
    SQLITE_ORM_INLINE_VAR constexpr string_constant_type unix_dotfile_vfs_name = "unix-dotfile";
    // SQLITE_ORM_WIN
    SQLITE_ORM_INLINE_VAR constexpr string_constant_type win32_longpath_vfs_name = "win32-longpath";
}
```

For other VFS objects unknown to us, programmers can easily define corresponding name constants themselves.


Addition:
I suggest placing the definition of the alias template `string_constant_type` in _serialize_result_type.h_.

dev/storage_options.h
```diff
@@ -30,6 +30,8 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
         /// Whether to open the database once and for all.
         /// Required if using a 'storage' instance from multiple threads.
         bool open_forever = false;
+        vfs_object vfs_mode = vfs_object::default_vfs;
```
trueqbit: Please rename to
```c++
std::string vfs_name{default_vfs_name};
```

dev/storage_base.h
```diff
@@ -640,6 +640,29 @@ namespace sqlite_orm {
                 return this->connection->retain_count() > 0;
             }
 
+            /**
+             * Public method for checking the VFS implementation being used by
+             * this storage object. Mostly useful for debug.
+             */
+            vfs_object vfs() const {
```
trueqbit: Please rename to
```c++
const std::string& vfs_name() const
```

dev/storage_base.h
```diff
@@ -640,6 +640,29 @@ namespace sqlite_orm {
                 return this->connection->retain_count() > 0;
             }
 
+            /**
+             * Public method for checking the VFS implementation being used by
```
trueqbit: Simplify comment:
```c++
// Return the name of the VFS object used by the database connection
```

dev/open_mode.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <sqlite3.h>
+
+namespace sqlite_orm {
+
+    enum class open_mode {
+        default_mode = 0,
```
trueqbit: Please rename to `default_`

dev/connection_holder.h
```diff
@@ -73,6 +90,8 @@ namespace sqlite_orm {
 
           private:
             const std::function<void(sqlite3* db)> _didOpenDb;
+            const vfs_object _vfsObject;
```
trueqbit: Please rename to
```c++
const std::string vfs_name;
```
... and make it a public member (like `filename`).

dev/connection_holder.h
```diff
@@ -73,6 +90,8 @@ namespace sqlite_orm {
...
+            const open_mode _openMode;
```
trueqbit: Please rename to
```c++
const db_open_mode open_mode;
```
... and make it a public member (like `filename`).

dev/open_mode.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <sqlite3.h>
+
+namespace sqlite_orm {
+
+    enum class open_mode {
```
trueqbit: I was thinking more about the problem of name disambiguation.
A great solution is to be a bit more descriptive, and allows for naming member variables of that type `open_mode`.

```c++
enum class db_open_mode;
```

dev/connection_holder.h
```diff
@@ -65,6 +74,14 @@ namespace sqlite_orm {
                 return _retainCount.load(std::memory_order_relaxed);
             }
 
+            vfs_object vfs() {
```
trueqbit: Remove as per comment on renaming the `_vfsObject` member variable -> `vfs_name` below.

dev/connection_holder.h
```diff
@@ -65,6 +74,14 @@ namespace sqlite_orm {
...
+                return _vfsObject;
+            }
+
+            sqlite_orm::open_mode open_flags() {
```
trueqbit: Remove as per comment on renaming the `_openMode` member variable -> `open_mode` below.

dev/storage_base.h
```diff
@@ -640,6 +640,29 @@ namespace sqlite_orm {
...
+            open_mode open_flags() const {
+                return this->connection->open_flags();
+            }
+
+            /**
+             * Return true if this database object is opened in a readonly state. 
+             */
+            bool readonly() const {
+                sqlite3* db = this->connection->get();
+                return sqlite3_db_readonly(db, "main");
```
trueqbit: Please cast to `bool` explicitly.

dev/storage_options.h
```diff
@@ -30,6 +30,8 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+        open_mode open_flags = open_mode::default_mode;
```
trueqbit: Please rename to
```c++
db_open_mode open_mode = db_open_mode::default_;
```

dev/open_mode.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <sqlite3.h>
+
+namespace sqlite_orm {
```
trueqbit: Meanwhile sqlite_orm can be used as a C++ named module, and public symbols must be exported:

```c++
SQLITE_ORM_EXPORT namespace sqlite_orm {
```

dev/storage_base.h
```diff
@@ -640,6 +640,29 @@ namespace sqlite_orm {
...
+             * Return the current open_mode for this storage object. 
+             */
+            open_mode open_flags() const {
+                return this->connection->open_flags();
+            }
+
+            /**
+             * Return true if this database object is opened in a readonly state. 
+             */
+            bool readonly() const {
```
trueqbit: If your intent is to access SQLite's database (handle) then please rename and remove the `const`-qualifier:
```c++
bool db_readonly() {
```

dev/storage_base.h
```diff
@@ -640,6 +640,29 @@ namespace sqlite_orm {
...
+                sqlite3* db = this->connection->get();
```
trueqbit: Better increase the connection reference count like e.g. `db_release_memory()`:
```c++
auto con = this->get_connection();
```

tests/static_tests/vfs_open_mode_static_tests.cpp
```diff
@@ -0,0 +1,29 @@
+#include <sqlite_orm/sqlite_orm.h>
+#include <catch2/catch_all.hpp>
+
+using namespace sqlite_orm;
+
+TEST_CASE("vfs string conversion returns expected strings") {
```
trueqbit: These tests aren't required anymore with constants...

tests/vfs_open_mode_tests.cpp
```diff
@@ -0,0 +1,104 @@
+#include <sqlite_orm/sqlite_orm.h>
+#include <catch2/catch_all.hpp>
+
```
trueqbit, line 8: Need to
```c++
#include <cstdio>  //  std::remove
```

tests/vfs_open_mode_tests.cpp
trueqbit, line 1: Initially I left a dozen review comments, mainly about when to use REQUIRE vs. CHECK, matching specific exceptions, or not asserting on things because they are tested elsewhere already.

I figured it's better if I simply post my version of the test cases and assertions.
I didn't change any of your actual logic!

```c++
#include <catch2/catch_all.hpp>
#include <cstdio>  //  std::remove
#if SQLITE_ORM_HAS_INCLUDE(<filesystem>)
#include <filesystem>
#endif
#include "catch_matchers.h"

#ifdef SQLITE_ORM_CTAD_SUPPORTED

struct User {
    std::string id;
};

using namespace sqlite_orm;

static const auto default_table = make_table("users", make_column("id", &User::id, primary_key()));

TEST_CASE("vfs modes open successfully") {

#if defined(SQLITE_ORM_APPLE)
    vfs_object vfs = GENERATE(vfs_object::unix, vfs_object::unix_posix, vfs_object::unix_dotfile, vfs_object::unix_afp);
#elif defined(SQLITE_ORM_UNIX)
    vfs_object vfs = GENERATE(vfs_object::unix, vfs_object::unix_posix, vfs_object::unix_dotfile);
#elif defined(SQLITE_ORM_WIN)
    vfs_object vfs = GENERATE(vfs_object::win32, vfs_object::win32_longpath);
#endif
    internal::serialize_result_type vfs_string = internal::vfs_object_to_string(vfs);

    connection_control options{true, vfs};
    auto storage = make_storage(":memory:", options, default_table);
    UNSCOPED_INFO("FAILED VFS: " << vfs_string);
    REQUIRE_NOTHROW(storage.open_forever());

    REQUIRE(storage.vfs() == vfs);
    REQUIRE(storage.open_flags() == open_mode::default_mode);

    SECTION("Storage copy operator carries over vfs option") {
        auto storage_copy = storage;
        CHECK(storage_copy.is_opened());
        CHECK(storage_copy.vfs() == vfs);
        CHECK(storage_copy.open_flags() == open_mode::default_mode);
    }
}

TEST_CASE("readwrite/readonly open modes behaves as expected") {
    const bool in_memory = GENERATE(true, false);
    const char* tmp_filename = in_memory ? ":memory:" : "open_mode.sqlite";

    connection_control options{true}, readonly_options{true};
    options.open_flags = open_mode::create_readwrite;
    readonly_options.open_flags = open_mode::readonly;

    if (!in_memory) {
        std::remove(tmp_filename);
    }

    SECTION("rw+ro") {
        auto storage = make_storage(tmp_filename, default_table, options);
        CHECK(storage.is_opened());
        REQUIRE(storage.open_flags() == open_mode::create_readwrite);
        REQUIRE_FALSE(storage.readonly());

        storage.sync_schema();
        const User dummy{"dummy"};
        storage.replace(dummy);

        SECTION("readonly open mode behaves as expected") {
            auto readonly_storage = make_storage(tmp_filename, readonly_options, default_table);
            CHECK(readonly_storage.is_opened());
            REQUIRE(readonly_storage.open_flags() == open_mode::readonly);
            REQUIRE(readonly_storage.readonly());

            if (in_memory) {
                SKIP("skipped for in-memory");
            }
            REQUIRE_NOTHROW(readonly_storage.sync_schema());
            CHECK_NOTHROW(readonly_storage.get<User>(dummy.id));
            const ErrorCodeExceptionMatcher readOnlyExceptionMatcher(sqlite_errc{SQLITE_READONLY});
            REQUIRE_THROWS_MATCHES(readonly_storage.remove<User>(dummy.id),
                                   std::system_error,
                                   readOnlyExceptionMatcher);
        }
    }
    SECTION("readonly fails with non-existing files") {
        if (in_memory) {
            SKIP("skipped for in-memory");
        }

#if __cpp_lib_filesystem >= 201703L
        CHECK_FALSE(std::filesystem::exists(tmp_filename));
#endif
        const ErrorCodeExceptionMatcher cantOpenExceptionMatcher(sqlite_errc{SQLITE_CANTOPEN});
        REQUIRE_THROWS_MATCHES(make_storage(tmp_filename, readonly_options, default_table),
                               std::system_error,
                               cantOpenExceptionMatcher);
    }
}
#endif
```

trueqbit (changes requested):

dev/connection_holder.h
```diff
@@ -1,35 +1,42 @@
...
                 // therefore we can just use an atomic increment but don't need sequencing due to `prevCount > 0`.
                 if (_retainCount.fetch_add(1, std::memory_order_relaxed) == 0) {
-                    int rc = sqlite3_open_v2(this->filename.c_str(),
-                                             &this->db,
-                                             SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE,
-                                             nullptr);
+
+                    const int open_flags = internal::db_open_mode_to_int_flags(open_mode);
+
+                    int rc = sqlite3_open_v2(this->filename.c_str(), &this->db, open_flags, vfs_name.c_str());
```
trueqbit: Better qualify member variables that are not prefixed with an underscore, like `this->vfs_name` - an unspoken policy of sqlite_orm.
```.

File: dev/storage_base.h [hunk 33]
```diff
@@ -640,6 +640,28 @@ namespace sqlite_orm {
...
+             * Return the current open_mode for this storage object. 
+             */
+            db_open_mode open_mode() const {
+                return this->connection->open_mode;
+            }
+
+            /**
+             * Return true if this database object is opened in a readonly state. 
+             */
+            bool readonly() {
```
trueqbit: Please rename to `db_readonly()` as it queries the database handle - just like `db_release_memory()`.

dev/storage_base.h
```diff
@@ -640,6 +640,28 @@ namespace sqlite_orm {
...
+                sqlite3* db = this->get_connection().get();
```
trueqbit: This doesn't work because the connection reference returned by `get_connection()` goes out of of scope after that line.
```c++
auto con = this->get_connection();
return static_cast<bool>(sqlite3_db_readonly(con.get(), "main"));
```

dev/vfs_name.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "functional/config.h"
```
trueqbit: Please remove, the configuration is universally available.

dev/db_open_mode.h
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "functional/config.h"
```
trueqbit: Please remove, the configuration is universally available.

dev/vfs_name.h
```diff
@@ -0,0 +1,24 @@
...
+SQLITE_ORM_EXPORT namespace sqlite_orm {
+
+#ifdef SQLITE_ORM_UNIX
+    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_vfs_name = "unix";
+    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_posix_vfs_name = unix_vfs_name;
+    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_dotfile_vfs_name = "unix-dotfile";
+#ifdef SQLITE_ORM_APPLE
+    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_afp_vfs_name = "unix-afp";
+#endif
+    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type default_vfs_name = "unix";
```
trueqbit: Should be ` = unix_vfs_name`?

dev/connection_holder.h
```diff
@@ -1,35 +1,42 @@
 #pragma once
 
-#include <sqlite3.h>
```
trueqbit: Why did you remove this include?

trueqbit:
@ancientjpeg Why do you always force-push changes onto your feature branch?

dev/connection_holder.h
ancientjpeg: that was an accident, I thought my LSP had accidentally added `#include <sqlite_orm.h>`.

dev/vfs_name.h
ancientjpeg: Yup, fixing now!

ancientjpeg:
> @ancientjpeg Why do you always force-push changes onto your feature branch?

@trueqbit I force-push because I generally rebase onto `dev` any time there's updates just in case there's conflicts to resolve. I generally prefer a rebase workflow but if you'd prefer a merge workflow for PRs against this repo I'll gladly do that instead!

dev/storage_base.h
ancientjpeg: Gotcha, thank you for the clarification!

trueqbit (approved):

ancientjpeg:
Corrected conflict over extended response code usage. I opted to keep the flag addition in `connection_holder.h` instead of inserting it into `db_open_mode_to_int_flags()`, as I felt it was clearer for the latter to return exactly the flags its input argument corresponds to and nothing more. The flag manipulation feels like the responsibility of the connection holder, not the flag translator IMO.

dev/functional/config.h
```diff
@@ -86,6 +86,15 @@
 #define SQLITE_ORM_WITH_CTE
 #endif
 
+#if defined(_WIN32)
+#define SQLITE_ORM_WIN
+#elif defined(__APPLE__)
```
fnc12, line 77: probably we will need it for android as well `__APPLE__` is defined for iOS

dev/serialize_result_type.h
```diff
@@ -12,9 +12,11 @@ namespace sqlite_orm {
 #ifdef SQLITE_ORM_STRING_VIEW_SUPPORTED
         using serialize_result_type = std::string_view;
         using serialize_arg_type = std::string_view;
+        using string_constant_type = std::string_view;
 #else
         using serialize_result_type = std::string;
         using serialize_arg_type = const std::string&;
+        using string_constant_type = const char*;
```
fnc12, line 19: where do we use this?

tests/vfs_open_mode_tests.cpp
```diff
@@ -0,0 +1,99 @@
...
+};
+
+using namespace sqlite_orm;
+
+static const auto default_table = make_table("users", make_column("id", &User::id, primary_key()));
+
+TEST_CASE("vfs modes open successfully") {
+
+#if defined(SQLITE_ORM_APPLE)
+    internal::string_constant_type vfs =
```
fnc12, line 22: can we use `std::string_view` instead to avoid `const char*` usage? Just want to keep OOP style and to skip C strings where possible

tests/vfs_open_mode_tests.cpp
```diff
@@ -0,0 +1,99 @@
...
+    REQUIRE_NOTHROW(storage.open_forever());
+
+    REQUIRE(storage.vfs_name() == vfs);
+    REQUIRE(storage.open_mode() == db_open_mode::default_);
+
+    SECTION("Storage copy operator carries over vfs option") {
+        auto storage_copy = storage;
+        CHECK(storage_copy.is_opened());
+        CHECK(storage_copy.vfs_name() == vfs);
+        CHECK(storage_copy.open_mode() == db_open_mode::default_);
```
fnc12: I'd put `REQUIRE` here cause `REQUIRE` fails test and we really expect these values to be true

tests/vfs_open_mode_tests.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+TEST_CASE("readwrite/readonly open modes behaves as expected") {
+    const bool in_memory = GENERATE(true, false);
+    const char* tmp_filename = in_memory ? ":memory:" : "open_mode.sqlite";
+
+    connection_control options{true}, readonly_options{true};
```
fnc12, line 56: please skip declaring >1 vars in one statement - it is hard to read. And as far as I remember the code stile in wiki requires to avoid such statements

fnc12 (approved):
the code is good, the tests are good. The feature is very good. Thank you for submitting this. I got several nits which are not so urgent. You  can fix it now within this PR or we can change it later in `dev` branch in further PRs - up to you

tests/vfs_open_mode_tests.cpp
trueqbit: no, not yet. not all STL libraries are C++17

fnc12:
Снимок экрана 2025-03-29 в 14 13 28
how to fix this?

@trueqbit @ancientjpeg

trueqbit:
> https://github.com/fnc12/sqlite_orm/pull/1393#issuecomment-2763248562
> how to fix this?
> 
> @trueqbit @ancientjpeg

Hmm, which SQLite library version are you using? It should be there since SQLite 3.3.8, and is defined in sqlite3.h: 
image

BTW, your screenshot seems not to be from the Minimal-Audio:feature/1392-support-sqlite-vfs branch...
BTW @trueqbit The version number should be `>= 3003008` instead of `>= 3008008`.

If everything fails we could of course check directly with `SQLITE_OPEN_EXRESCODE`...

fnc12:
> Hmm, which SQLite library version are you using? It should be there since SQLite 3.3.8, and is defined in sqlite3.h:

I use `3.26.0` and my `sqlite3.h` doesn't have `SQLITE_OPEN_EXRESCODE`

> BTW, your screenshot seems not to be from the Minimal-Audio:feature/1392-support-sqlite-vfs branch...

My screenshot os from `dev` merged into `feature/logger`

> BTW @trueqbit The version number should be >= 3003008 instead of >= 3008008.

I recon you put a wrong at here =)

fnc12:
Снимок экрана 2025-03-29 в 16 04 19

trueqbit:
> > Hmm, which SQLite library version are you using? It should be there since SQLite 3.3.8, and is defined in sqlite3.h:
> 
> I use `3.26.0` and my `sqlite3.h` doesn't have `SQLITE_OPEN_EXRESCODE`

Alright, I assumed this flag was introduced with the "extended result code" feature....
@ancientjpeg Maybe you want to replace the check `#if SQLITE_VERSION_NUMBER >= 3008008` with `#if SQLITE_VERSION_NUMBER >= 3037002`, then it is already done?

> > BTW @trueqbit The version number should be >= 3003008 instead of >= 3008008.
> 
> I recon you put a wrong at here =)

No, that's right to blame myself, because I was the one who introduced the handling of extended result codes with PR #1412.

fnc12:
@trueqbit oh so I posted my issue at wrong place. Sorry @ancientjpeg

dev/functional/config.h
ancientjpeg: If you don't mind I'd prefer this be done in a separate PR, but I'd be happy to implement it.

dev/serialize_result_type.h
ancientjpeg: This type was added at @trueqbit's suggestion to define string constants, such as the `vfs_name` constants:

```cpp
#pragma once

#include "serialize_result_type.h"

SQLITE_ORM_EXPORT namespace sqlite_orm {

#ifdef SQLITE_ORM_UNIX
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_vfs_name = "unix";
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_posix_vfs_name = unix_vfs_name;
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_dotfile_vfs_name = "unix-dotfile";
#ifdef SQLITE_ORM_APPLE
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type unix_afp_vfs_name = "unix-afp";
#endif
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type default_vfs_name = unix_vfs_name;
#endif

#ifdef SQLITE_ORM_WIN
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type win32_vfs_name = "win32";
    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type win32_longpath_vfs_name = "win32-longpath";

    SQLITE_ORM_INLINE_VAR constexpr internal::string_constant_type default_vfs_name = win32_vfs_name;
#endif
}
```

dev/functional/config.h
fnc12: ok no rush

ancientjpeg:
@fnc12 @trueqbit I believe I've addressed all comments! As I said before, I'm happy to expand the platform definition macros but I believe it would be wise to do that in a separate PR to keep the scope limited here. Let me know if there's any other changes you'd like made before merging.

tests/vfs_open_mode_tests.cpp
trueqbit: The assignment from the ternary expression was OK, it should have been `readonly_options` that should be defined on a separate line 😉 

But that's not why the unit tests fail - as far as I see this is why:
* We enabled a compiler error for the wrong order of member initializers:
  > C:\projects\sqlite-orm\include\sqlite_orm\sqlite_orm.h(13941,51): error C5038: data member 'sqlite_orm::internal::connection_holder::filename' will be initialized after data member 'sqlite_orm::internal::connection_holder::vfs_name'
* it's good if you merge again dev -> 1392-support-sqlite-vfs

trueqbit (approved):

trueqbit:
>@ancientjpeg
[fix visual studio error](https://github.com/fnc12/sqlite_orm/pull/1393/commits/53bad11f98beb2d045c9f0d431647ebf2c13b9b5)

IMHO we should prefer good commit messages, or at least correct ones - it's not a Visual Studio bug, it's the wrong order of initialization of members.
We just recently changed the compiler warning to an [error](https://github.com/fnc12/sqlite_orm/blob/8e8ef203bdff48f904b6fb5d74a6404384c6759c/CMakeLists.txt#L60) so that we would notice it in the future.

Otherwise thank you for your good worK and outstanding patience! 🥇

ancientjpeg:
Apologies about the commit message—I had interpreted this error as being MSVC-specific (I see those a lot), as I didn't realize that `Werror=reorder` had been added in the recent merge. Thank you for all your time spent reviewing and refining this code!

PR #1395, opened 2025-02-04
Threadsafe connection

trueqbit:
With C++20 a more light-weight binary semaphore can be conditionally used to synchronize opening and closing a database connection.

Therefore, the connection holder should be performant in all variants:
1. single-threaded use
2. opened once (open forever)
3. concurrent open/close

### DIFF
--- a/dev/connection_holder.h
+++ b/dev/connection_holder.h
@@ -2,20 +2,114 @@
 
 #include <sqlite3.h>
 #include <atomic>
+#ifdef SQLITE_ORM_CPP20_SEMAPHORE_SUPPORTED
+#include <semaphore>
+#endif
 #include <string>  //  std::string
 
 #include "error_code.h"
 
 namespace sqlite_orm {
-
     namespace internal {
 
-        struct connection_holder {
+#ifdef SQLITE_ORM_CPP20_SEMAPHORE_SUPPORTED
+        /** 
+         *  The connection holder should be performant in all variants:
+            1. single-threaded use
+            2. opened once (open forever)
+            3. concurrent open/close
 
-            connection_holder(std::string filename_) : filename(std::move(filename_)) {}
+            Hence, a light-weight binary semaphore is used to synchronize opening and closing a database connection.
+        */
+        struct connection_holder {
+            struct maybe_lock {
+                maybe_lock(std::binary_semaphore& sync, bool shouldLock) noexcept(noexcept(sync.acquire())) :
+                    isSynced{shouldLock}, sync{sync} {
+                    if (shouldLock) {
+                        sync.acquire();
+                    }
+                }
+
+                ~maybe_lock() {
+                    if (isSynced) {
+                        sync.release();
+                    }
+                }
+
+                const bool isSynced;
+                std::binary_semaphore& sync;
+            };
+
+            explicit connection_holder(std::string filename) : filename(std::move(filename)) {}
 
             void retain() {
-                if (1 == ++this->_retain_count) {
+                const maybe_lock maybeLock{this->_sync, !this->_openedForeverHint};
+
+                // `maybeLock.isSynced`: the lock above already synchronized everything, so we can just atomically increment the counter
+                // `!maybeLock.isSynced`: we presume that this the connection is opened once in a single-threaded context [open forever].
+                //                        so we can just use an atomic operation but don't need sequencing due to `prevCount > 0`.
+                const int prevCount = this->_retain_count.fetch_add(1, std::memory_order_relaxed);
+
+                if (prevCount == 0 || (maybeLock.isSynced && !this->db)) {
+                    if (int rc = sqlite3_open(this->filename.c_str(), &this->db); rc != SQLITE_OK)
+                        [[unlikely]] /*unexpected*/ {
+                        throw_translated_sqlite_error(this->db);
+                    }
+                }
+            }
+
+            void release() {
+                const maybe_lock maybeLock{this->_sync, !this->_openedForeverHint};
+
+                if (int prevCount = this->_retain_count.fetch_sub(
+                        1,
+                        maybeLock.isSynced
+                            // the lock above already synchronized everything, so we can just atomically decrement the counter
+                            ? std::memory_order_relaxed
+                            // the counter must serve as a synchronization point
+                            : std::memory_order_acq_rel);
+                    prevCount > 1) {
+                    return;
+                }
+
+                if (int rc = sqlite3_close(this->db); rc != SQLITE_OK) [[unlikely]] {
+                    throw_translated_sqlite_error(this->db);
+                } else {
+                    this->db = nullptr;
+                }
+            }
+
+            sqlite3* get() const {
+                // note: ensuring a valid DB handle was already memory ordered with `retain()`
+                return this->db;
+            }
+
+            /** 
+             *  @attention While retrieving the reference count value is well-defined it makes only sense at single-threaded points in code
+             */
+            int retain_count() const {
+                return this->_retain_count.load(std::memory_order_relaxed);
+            }
+
+            const std::string filename;
+
+          protected:
+            sqlite3* db = nullptr;
+
+          private:
+            std::binary_semaphore _sync{1};
+            const bool _openedForeverHint = false;
+            std::atomic_int _retain_count{};
+        };
+#else
+        struct connection_holder {
+            explicit connection_holder(std::string filename) : filename(std::move(filename)) {}
+
+            void retain() {
+                // first one opens the connection.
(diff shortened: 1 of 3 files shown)


dev/connection_holder.h
```diff
@@ -3,33 +3,167 @@
...
+                if (_onAfterOpen) {
+                    _onAfterOpen(this->db);
+                }
             }
 
             void release() {
                 // last one closes the connection.
                 // we assume that this might happen by any thread, therefore the counter must serve as a synchronization point.
-                if (this->_retain_count.fetch_sub(1, std::memory_order_acq_rel) == 1) {
+                if (_retainCount.fetch_sub(1, std::memory_order_acq_rel) == 1) {
```
fnc12: why did you remove `this`?

dev/tuple_helper/tuple_transformer.h
```diff
@@ -112,5 +112,23 @@ namespace sqlite_orm {
                 std::make_index_sequence<std::tuple_size<std::remove_reference_t<Tpl>>::value>{},
                 std::forward<Projection>(project));
         }
+
+#ifdef SQLITE_ORM_CTAD_SUPPORTED
+        template<template<typename...> class R, class Tpl, size_t... Idx, class Projection = polyfill::identity>
+        constexpr auto create_from_tuple(Tpl&& tpl, std::index_sequence<Idx...>, Projection project = {}) {
```
fnc12, line 126: would be nice to have it covered with unit tests

tests/storage_tests.cpp
```diff
@@ -4,6 +4,26 @@
 
 using namespace sqlite_orm;
 
+#ifdef SQLITE_ORM_CTAD_SUPPORTED
+TEST_CASE("connection control") {
+    const auto openForever = GENERATE(false, true);
```
fnc12, line 120: what is `GENERATE`?

tests/storage_tests.cpp
```diff
@@ -4,6 +4,26 @@
...
+    SECTION("") {
+        SECTION("empty") {
+            auto storage = make_storage("", connection_control{openForever}, on_open([](sqlite3*) {}));
```
fnc12: probably we may add also `on_open` call counter increment inside `on_open` and checking that is was or wasn't called

tests/storage_tests.cpp
```diff
@@ -4,6 +4,26 @@
 
 using namespace sqlite_orm;
 
+#ifdef SQLITE_ORM_CTAD_SUPPORTED
+TEST_CASE("connection control") {
+    const auto openForever = GENERATE(false, true);
+    SECTION("") {
```
fnc12: what is the purpose of empty single session?

dev/connection_holder.h
trueqbit: We have code that omits `this` if the member variable starts with an underscore, so I removed it here as well.
What do you think? I find it easier to read and it's clear enough.

tests/storage_tests.cpp
trueqbit: `GENERATE` repeats the `SECTION` following it.

tests/storage_tests.cpp
trueqbit: A Catch2 [data generator](https://github.com/catchorg/Catch2/blob/devel/docs/generators.md).

dev/tuple_helper/tuple_transformer.h
trueqbit: It's in _tests/static_tests/functional/tuple_transform.cpp_ now.

dev/functional/cxx_scope_guard.h
```diff
@@ -0,0 +1,16 @@
...
+        Poor-man's scope (exit) guard until C++29 finally comes with proper standard facilities [Draft D3610].
+     */
+    template<class F>
+    struct scope_guard {
+        explicit scope_guard(F f) : f{std::move(f)} {}
+        ~scope_guard() {
+            f();
+        }
+
+        F f;
```
fnc12: nit: I'd skip single character names. I'd name it `function` to make it more clear for future readers. Short variable names used to be popular when we had only 80 character limited lines on our screens. Now we don't need to the code to be encoded, it can be written as a regular text to make it easier to read it

dev/storage_base.h
```diff
@@ -40,38 +39,65 @@ namespace sqlite_orm {
...
+            const sqlite_executor executor;
+
+          public:
+            /** 
+             *  Attention: You must ensure that to set this function only from a single-threaded context.
+             */
             std::function<void(sqlite3*)> on_open;
             pragma_t pragma;
+            /** 
+             *  Attention: You must ensure that to set database limite only from a single-threaded context.
```
fnc12: typo: limite[s]

dev/storage_base.h
```diff
@@ -503,6 +545,8 @@ namespace sqlite_orm {
             /**
              *  Delete aggregate function you created before.
              *  Can be called at any time (in a single-threaded context) no matter whether the database connection is open or not.
+             * 
+             * Attention: You must ensure that to call this method only in a single-threaded context.
```
fnc12: nit: probably you skipped one space before `Attention`

dev/storage_base.h
```diff
@@ -512,12 +556,21 @@ namespace sqlite_orm {
             }
 
 #ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+            /**
+             *  Delete aggregate function you created before.
+             *  Can be called at any time (in a single-threaded context) no matter whether the database connection is open or not.
+             * 
+             * Attention: You must ensure that to call this method only in a single-threaded context.
```
fnc12: nit: probably you skipped one space before `Attention`

dev/storage_base.h
```diff
@@ -563,35 +621,41 @@ namespace sqlite_orm {
...
             void begin_immediate_transaction() {
-                this->begin_transaction_internal("BEGIN IMMEDIATE TRANSACTION");
+                sqlite3* db = this->connection->retain();
+                this->executor.perform_void_exec(db, "BEGIN IMMEDIATE TRANSACTION");
             }
 
             void begin_exclusive_transaction() {
-                this->begin_transaction_internal("BEGIN EXCLUSIVE TRANSACTION");
+                sqlite3* db = this->connection->retain();
+                this->executor.perform_void_exec(db, "BEGIN EXCLUSIVE TRANSACTION");
```
fnc12, line 640: now we have `this->executor.perform_void_exec(db, "BEGIN TRANSACTION");` calls in two places which looks like code duplication. I mean I understand that one like can barely be called as code duplication but before this PR we did not have it duplicated. It relates to all `void begin_*transaction()` functions in this file. Probably we can avoid duplicating this just to make it easier in future updating transaction calls.

tests/storage_tests.cpp
```diff
@@ -1,10 +1,122 @@
...
+        operator bool() const {
+            return locked;
+        }
+
+        std::mutex& mtx;
+        const bool locked;
+    };
+
+    const bool openForever = GENERATE(false, true);
+    SECTION("") {
```
fnc12: what's the point of empty named section?

fnc12 (approved):

dev/functional/cxx_scope_guard.h
trueqbit: It lives up to its name - a poor man's scope guard 🤣

tests/storage_tests.cpp
trueqbit: Because of `GENERATE` above...

fnc12: I was sure `GENERATE` works even without sections just cause of `TEST_CASE` macro

dev/storage_base.h
trueqbit: I know what you mean, and I thought about it, too.

But I also feel that duplicating functions just for the purpose of forwarding a string isn't the best design choice either, and makes things more complicated in my mind - honestly it was hard to understand before I changed it, so this is the smallest common denominator.

What about an (internally used) string constant `constexpr orm_gsl::czstring begin_<bla>_transaction_str = "BEGIN <BLAH> TRANSACTION"`?

tests/storage_tests.cpp
trueqbit: You are right, `GENERATE` introduces a virtual section! However I would leave the braces for the scope, just to make it visually clear.

dev/storage_base.h
fnc12: I'd leave it like that instead of `constexpr orm_gsl::czstring begin_<bla>_transaction_str = "BEGIN <BLAH> TRANSACTION"`. I am totally ok with your opinion as well

PR #1399, opened 2025-02-16
Database connection control options

### LINKED ISSUE #1000: on_open is never called for in memory storage
Hi,

assuming the standard way to obtain a sqlite3 handle in the FAQ (https://github.com/fnc12/sqlite_orm/wiki/FAQ)

```auto storage = make_storage(...);
storage.on_open = [](sqlite3 *db) {
    // do whatever you want except database closing
};
```

in combination with in-memory storage, the on_open lambda is never executed.

The reason is the following special treatment of in-memory in storage_base:

``` storage_base(const std::string& filename_, int foreignKeysCount) :
                pragma(std::bind(&storage_base::get_connection, this)),
                limit(std::bind(&storage_base::get_connection, this)),
                inMemory(filename_.empty() || filename_ == ":memory:"),
                connection(std::make_unique<connection_holder>(filename_)), cachedForeignKeysCount(foreignKeysCount) {
                if(this->inMemory) {
                    this->connection->retain();
                    this->on_open_internal(this->connection->get());
                }
            }
```

Basically, I see no way to have anything registered for on_open at this point already.
Attaching it after the storage has been created has no effect, because, e.g.:

```
 void begin_transaction() {
                this->connection->retain();
                if(1 == this->connection->retain_count()) {
                    this->on_open_internal(this->connection->get());
                }
                auto db = this->connection->get();
                perform_void_exec(db, "BEGIN TRANSACTION");
            }
```

So, retain_count is already 2 (for the in-memory case) and on_open_internal is never called (which makes sense).

Is this on purpose / by design? 
One option would be to provide the on_open handler directly for make_storage, but in general, I am not sure if the on_open semantics make sense for the in-memory model (which by definition is just ... open). Thoughts?

trueqbit:
* Introduced a connection control struct as an aggregate of options. This opens the way for various options and modes to establish a database connection, and can be used to make certain assumption in a multi-threading environment. E.g. `make_storage("", connection_control{.open_forever = true});`. Also see PR #1393.
* An 'on open' callback can now be specified in a declarative way to the `make_storage()` call. This finally solves the issue that in-memory databases didn't call the 'on open' callback. E.g. `make_storage("", on_open([](sqlite3* db) {}));`.
* Reordered members of the connection holder. This will be important for upcoming multi-threading improvements (see PR #1395), avoiding false L1 cache sharing.

### DIFF
--- a/dev/connection_holder.h
+++ b/dev/connection_holder.h
@@ -33,10 +33,10 @@ namespace sqlite_orm {
                     if (rc != SQLITE_OK) SQLITE_ORM_CPP_UNLIKELY /*possible, but unexpected*/ {
                         throw_translated_sqlite_error(this->db);
                     }
-                }
 
-                if (_onAfterOpen) {
-                    _onAfterOpen(this->db);
+                    if (_onAfterOpen) {
+                        _onAfterOpen(this->db);
+                    }
                 }
             }
 
--- a/dev/connection_holder.h
+++ b/dev/connection_holder.h
@@ -13,13 +13,13 @@ namespace sqlite_orm {
     namespace internal {
 
         struct connection_holder {
-            connection_holder(std::string filename, std::function<void(sqlite3*)> onAfterOpen) :
-                _onAfterOpen{std::move(onAfterOpen)}, filename(std::move(filename)) {}
+            connection_holder(std::string filename, std::function<void(sqlite3*)> didOpenDb) :
+                _didOpenDb{std::move(didOpenDb)}, filename(std::move(filename)) {}
 
             connection_holder(const connection_holder&) = delete;
 
-            connection_holder(const connection_holder& other, std::function<void(sqlite3*)> onAfterOpen) :
-                _onAfterOpen{std::move(onAfterOpen)}, filename{other.filename} {}
+            connection_holder(const connection_holder& other, std::function<void(sqlite3*)> didOpenDb) :
+                _didOpenDb{std::move(didOpenDb)}, filename{other.filename} {}
 
             void retain() {
                 // first one opens the connection.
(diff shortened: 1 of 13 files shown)


dev/connection_holder.h
```diff
@@ -12,25 +13,38 @@ namespace sqlite_orm {
     namespace internal {
 
         struct connection_holder {
-            connection_holder(std::string filename) : filename(std::move(filename)) {}
+            connection_holder(std::string filename, std::function<void(sqlite3*)> onAfterOpen) :
```
fnc12: nit: if we want to specify whether callback is going to be called before or after some event then we'd better avoid 'on*' naming and use 'willOpen' and 'didOpen' instead just like Apple does cause it is more concise than 'onAfterOpen' or 'onBeforeOpen'.

tests/storage_tests.cpp
```diff
@@ -4,6 +4,49 @@
 
 using namespace sqlite_orm;
 
+#ifdef SQLITE_ORM_CTAD_SUPPORTED
+TEST_CASE("connection control") {
+    const auto openForever = GENERATE(false, true);
+    SECTION("") {
+        bool onOpenCalled = false;
```
fnc12, line 11: I'd change `bool` to `int` to cover amount of `onOpen` calls count just in case. I know your code doesn't produce such a bug but if we cover it with small amount of energy then it will be a win

tests/storage_tests.cpp
```diff
@@ -4,6 +4,49 @@
 
 using namespace sqlite_orm;
 
+#ifdef SQLITE_ORM_CTAD_SUPPORTED
+TEST_CASE("connection control") {
+    const auto openForever = GENERATE(false, true);
```
fnc12, line 9: `GENERATE` solves the issue with parallels sections tree. I have to use it as well

fnc12 (approved):

dev/connection_holder.h
trueqbit: Interesting 👍

tests/storage_tests.cpp
trueqbit: thx! You just saved my butt :)

fnc12 (approved):

PR #1404, opened 2025-03-06
Enhanced support for primary and foreign key table constraints with inheritance

### LINKED ISSUE #1403: Foreign key and inheritance (self-referencing table)
Hi! I am trying to store tree-like structure in db. Please clarify is that possible to create a self-referencing tables like that?

```
#include <optional>

struct Base {
    int id;
    std::optional<int> parentId;

    Base() = default;
    Base(int id_, std::optional<int> parentId_ = std::nullopt)
        : id(id_), parentId(parentId_) {}
};

struct Container : public Base {
    std::string name;

    Container() = default;
    Container(int id_, std::optional<int> parentId_, std::string name_)
        : Base(id_, parentId_), name(name_) {}
};

int main(int, char**) {
    using namespace sqlite_orm;
    auto storage = make_storage("tree.sqlite",
                                make_table<Container>("Containers",
                                                  make_column("id", &Container::id, primary_key()),
                                                  make_column("name", &Container::name),
                                                  make_column("parent_id", &Container::parentId),
                                                  foreign_key(&Container::parentId).references(column<Container>(&Container::id)))
                                                  // foreign_key(&Container::parentId).references<Container>(column<Container>(&Container::id)))
                                );
    storage.sync_schema();
    return 0;
}
```

But i still face with error about it tries to find `<Base>` table (`pick_table<Base>`).
I have read related issues about explicit type specifiers etc (https://github.com/fnc12/sqlite_orm/issues/1175, https://github.com/fnc12/sqlite_orm/issues/1207), but in my case i need to use that 'parent_id' as a foreign key. Without foreign_key it works.

I understand it will be solved with another table with `id` to `parent_id` mapping, but i don't actually need that table in raw SQL. It is possible to create such table with e.g. `CREATE TABLE "containers" ("id" INTEGER PRIMARY KEY NOT NULL, "name" TEXT NOT NULL, "parent_id" INTEGER NULL, FOREIGN KEY("parent_id") REFERENCES "containers"("id"))` and it will work.

Thank you for the library (it is great) and hope to find out the answer.

trueqbit:
Primary keys and foreign keys now accept an explicit template parameter for a derived class.

Additionally:
* Introduced unit tests for primary and foreign key table constraints with inheritance
* Introduced unit tests for comparing (finding) fields
* Split primary key table and column constraint tests
* Making primary key expressions can now be done at compile-time [constexpr]
* Foreign keys are now counted at compile-time


dev/constraints.h
```diff
@@ -333,28 +331,23 @@ namespace sqlite_orm {
...
                 on_delete(*this, false, other.on_delete._action) {}
 
-            self& operator=(const self& other) {
-                this->columns = other.columns;
-                this->references = other.references;
-                this->on_update = {*this, true, other.on_update._action};
-                this->on_delete = {*this, false, other.on_delete._action};
-                return *this;
-            }
+            foreign_key_t& operator=(const foreign_key_t&) = delete;
```
fnc12, line 350: why do we remove `operator=`?

fnc12 (approved):

dev/constraints.h
trueqbit: > why do we remove `operator=`?

Counter question: Why do we need one? It's unused, not defaulted, and not trivial either, so I think we should just remove dead code.

fnc12: fair enough

PR #1407, opened 2025-03-20
Fixed binding values in several 'order by' clauses

### LINKED ISSUE #1405: `multi_order_by` causes string serialization to fail
## Issue Description

`multi_order_by` appears to be unable to properly serialize runtime strings into prepared statements. The result is that the ordering fails silently, and results are returned in an unexpected order. I'm including code that demonstrates the issue and a corresponding SQLite query that demonstrates the intended result. Given that `order_by` on its own is capable of executing these statements, I believe this is a bug.

As for the cause, all I can determine right now is that after the statement is prepared, `sqlite3_bind_text` is never being called, hence my suspicion this is a failure in the serializer's type detection.

### C++ test case
```cpp
TEST_CASE("Multi order string detection") {

    struct Datum {
        int id;
        std ::string text;
    };

    auto storage = make_storage("multistring.sqlite",
                                make_table("datum", make_column("id", &Datum::id), make_column("text", &Datum::text)));
    storage.sync_schema();

    constexpr int count = 1000;
    Datum match;
    for (int i = 0; i < count; ++i) {
        Datum d;
        d.id = i;
        d.text = std::to_string(d.id);

        if (i == count / 2) {
            match = d;
        }

        storage.insert(std::move(d));
    }

    SECTION("Test with order_by") {
        auto results = storage.get_all<Datum>(order_by(c(&Datum::text) == match.text).desc());
        CHECK(results[0].id == match.id);  // succeeds
    }

    SECTION("Test with multi_order_by with single clause") {
        auto results = storage.get_all<Datum>(multi_order_by(order_by(c(&Datum::text) == match.text).desc()));
        CHECK(results[0].id == match.id);  // fails
    }

    SECTION("Test with multi_order_by with multiple clauses") {
        auto results = storage.get_all<Datum>(
            multi_order_by(order_by(c(&Datum::text) == match.text).desc(), order_by(&Datum::id).desc()));
        CHECK(results[0].id == match.id);  // fails
        CHECK(results[1].id == (count - 1));  // fails
    }
}
```

### Functional SQL query

```sql
SELECT * FROM datum 
  ORDER BY
  text == '500' DESC,
  id ASC

```

The above query generates the below result, which is what is expected:

```sh
500|500
999|999
998|998
997|997
996|996
995|995
994|994
993|993
# ...
```

trueqbit:
Bindable values contained in several 'order by' clauses (with 'multi_order_by' expressions) were left unbound, which resulted in wrong query results or non-executable SQL queries.

### DIFF
--- a/dev/ast_iterator.h
+++ b/dev/ast_iterator.h
@@ -707,6 +707,16 @@ namespace sqlite_orm {
             }
         };
 
+        template<class... Args>
+        struct ast_iterator<multi_order_by_t<Args...>, void> {
+            using node_type = multi_order_by_t<Args...>;
+
+            template<class L>
+            void operator()(const node_type& node, L& lambda) const {
+                iterate_ast(node.args, lambda);
+            }
+        };
+
         template<class T>
         struct ast_iterator<collate_t<T>, void> {
             using node_type = collate_t<T>;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -113,7 +113,7 @@ for:
   install:
     - |-
       cd C:\Tools\vcpkg
-      git fetch --tags && git checkout 2025.01.13
+      git fetch --tags && git checkout 2025.02.14
       cd %APPVEYOR_BUILD_FOLDER%
       C:\Tools\vcpkg\bootstrap-vcpkg.bat -disableMetrics
       C:\Tools\vcpkg\vcpkg integrate install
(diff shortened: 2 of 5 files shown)


dev/ast_iterator.h
```diff
@@ -707,6 +707,16 @@ namespace sqlite_orm {
             }
         };
 
+        template<class... Args>
+        struct ast_iterator<multi_order_by_t<Args...>, void> {
```
fnc12, line 711: this is such an oversight!

fnc12 (approved):

dev/ast_iterator.h
trueqbit: Yep, it's actually a severe bug...

PR #1411, opened 2025-03-23
Logger

fnc12:
Added logger which runs a specified callable before running a query and after. For more info please check `logger_tests.cpp`. `sync_schema` is avoided cause we don't guarantee a set of calls in `sync_schema` and t is going to be changed at any time. But one can check logs during `sync_schema` call as well. So this PR has all crud, non-crud, pragma and misc calls covered with logs and according unit tests. One thing left to be done: iteration (both object and raw select iteration). I'd like to discuss this and update this PR with the result of discussion or merge it later. The question is when do we have to call loggers during iteration call? When we create an iterable object? When we start iteration? When we finish iteration? @trueqbit I'd like to know your opinion on it. Thanks


dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+            }
+
+            template<class L>
+            void perform_steps(sqlite3_stmt* stmt, L&& lambda) {
+                const auto sql = sqlite3_sql(stmt);
+                if (this->will_run_query) {
+                    this->will_run_query(sql);
+                }
+                int rc;
+                do {
```
trueqbit: Just reuse `internal::perform_step()` below! It returns early on `SQLITE_DONE`.

Alternatively a label and goto works, too, which avoids another function.

dev/connection_holder.h
```diff
@@ -29,7 +29,7 @@ namespace sqlite_orm {
                     int rc = sqlite3_open_v2(this->filename.c_str(),
                                              &this->db,
                                              SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE
-#if SQLITE_VERSION_NUMBER >= 3008008
+#if SQLITE_VERSION_NUMBER >= 3037002
```
fnc12, line 37: version fix is inlined. I can put it into a dedicated PR if it is important

dev/implementations/storage_definitions.h
```diff
@@ -149,12 +149,16 @@ namespace sqlite_orm {
...
+            std::string sql;
+            {
+                std::stringstream ss;
+                ss << "INSERT INTO " << streaming_identifier(destinationTableName) << " ("
+                   << streaming_identifiers(columnNames) << ") "
+                   << "SELECT " << streaming_identifiers(columnNames) << " FROM "
+                   << streaming_identifier(sourceTableName) << std::flush;
+                sql = ss.str();
+            }
+            this->executor.perform_void_exec(db, sql);
```
fnc12: now string stream is destroyed earlier reducing simultaneous memory consumption

dev/mapped_view.h
```diff
@@ -51,12 +51,14 @@ namespace sqlite_orm {
...
                 using context_t = serializer_context<db_objects_type>;
+
                 auto& dbObjects = obtain_db_objects(this->storage);
                 context_t context{dbObjects};
                 context.skip_table_name = false;
                 context.replace_bindable_with_question = true;
 
-                statement_finalizer stmt{prepare_stmt(this->connection.get(), serialize(this->expression, context))};
+                const std::string sql = serialize(this->expression, context);
+                statement_finalizer stmt{prepare_stmt(this->connection.get(), sql)};
```
fnc12, line 61: this is to think about: when do we have to call logger callbacks during iteration

dev/storage.h
```diff
@@ -1257,15 +1273,17 @@ namespace sqlite_orm {
...
+
                 context_t context{exprDBOs};
                 context.skip_table_name = false;
                 context.replace_bindable_with_question = true;
 
-                auto con = this->get_connection();
-                std::string sql = serialize(statement, context);
-                sqlite3_stmt* stmt = prepare_stmt(con.get(), std::move(sql));
-                return prepared_statement_t<S>{std::forward<S>(statement), stmt, std::move(con)};
+                auto conection = this->get_connection();
```
fnc12, line 1283: I am fighting with shortened versions of words

dev/storage.h
```diff
@@ -1460,7 +1478,17 @@ namespace sqlite_orm {
             void execute(const prepared_statement_t<replace_raw_t<Args...>>& statement) {
                 sqlite3_stmt* stmt = reset_stmt(statement.stmt);
                 iterate_ast(statement.expression, conditional_binder{stmt});
+                std::string sql;
+                if (this->executor.will_run_query || this->executor.did_run_query) {
+                    sql = statement.sql();
```
fnc12, line 1483: don't allocate a string if we don't have any callback set

dev/storage_base.h
```diff
@@ -224,24 +236,27 @@ namespace sqlite_orm {
                 return sqlite3_libversion();
             }
 
-            bool transaction(const std::function<bool()>& f) {
+            bool transaction(const std::function<bool()>& function) {
+                if (!function) {
```
fnc12, line 240: to remove UB

dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
 
 namespace sqlite_orm {
     namespace internal {
+
+        struct sqlite_executor {
+            std::function<void(serialize_arg_type sql)> will_run_query;
+            std::function<void(serialize_arg_type sql)> did_run_query;
+
+            inline void perform_void_exec(sqlite3* db, serialize_arg_type sql) const {
```
trueqbit: `serialize_arg_type` is wrong here, because `sqlite3_exec()` doesn't accept the length of the passed SQL query string. If `serialize_arg_type` is a`string_view` then we cannot assume that the string is null-terminated.

dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
             sqlite3_reset(stmt);
             return stmt;
         }
 
         // note: query is deliberately taken by value, such that it is thrown away early
-        inline sqlite3_stmt* prepare_stmt(sqlite3* db, std::string query) {
+        inline sqlite3_stmt* prepare_stmt(sqlite3* db, serialize_arg_type query) {
             sqlite3_stmt* stmt;
-            int rc = sqlite3_prepare_v2(db, query.c_str(), -1, &stmt, nullptr);
+            const int rc = sqlite3_prepare_v2(db, query.data(), -1, &stmt, nullptr);
```
trueqbit: Since `query` is of type `serialize_arg_type`, you must also pass the length of the SQL query string. If `serialize_arg_type` is a`string_view` then we cannot assume that the string is null-terminated.

dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+
+            inline void perform_exec(sqlite3* db,
+                                     const serialize_result_type& query,
+                                     int (*callback)(void* data, int argc, char** argv, char**),
+                                     void* user_data) const {
+                return perform_exec(db, query.data(), callback, user_data);
+            }
+
+            template<class L>
+            void perform_steps(sqlite3_stmt* stmt, L&& lambda) {
```
trueqbit: `sqlite_executor::perform_steps()` needs to be const-qualified, too.

dev/storage_base.h
```diff
@@ -1031,6 +1069,7 @@ namespace sqlite_orm {
             std::function<int(int)> _busy_handler;
             std::list<udf_proxy> scalarFunctions;
             std::list<udf_proxy> aggregateFunctions;
+            sqlite_executor executor;
```
trueqbit: Can we const-qualify the member? We will not allow changing the executor after creating the storage object, and the compiler can possibly improve code branches if it knows that the 'run query' callbacks are either set or null.
Also it allows us to optimize at compile-time at a later stage.

dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+                const auto sql = sqlite3_sql(stmt);
```
trueqbit: Better: `const char* sql` - I like explicit types unless undoubtedly clear or irrelevant.

dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+            }
+        };
+
         // Wrapper to reduce boiler-plate code
         inline sqlite3_stmt* reset_stmt(sqlite3_stmt* stmt) {
             sqlite3_reset(stmt);
             return stmt;
         }
 
         // note: query is deliberately taken by value, such that it is thrown away early
```
trueqbit: Remove the comment if you stay with `serialize_arg_type`.

dev/util.h
```diff
@@ -53,49 +58,91 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+                if (rc != SQLITE_OK) {
+                    throw_translated_sqlite_error(rc);
+                }
+                if (this->did_run_query) {
+                    this->did_run_query(sql);
+                }
+            }
+
+            inline void perform_exec(sqlite3* db,
+                                     const serialize_result_type& query,
```
trueqbit: `serialize_arg_type` is wrong here, because `sqlite3_exec()` doesn't accept the length of the passed SQL query string. If `serialize_arg_type` is a`string_view` then we cannot assume that the string is null-terminated.

trueqbit (commented):
> The question is when do we have to call loggers during iteration call? When we create an iterable object? When we start iteration? When we finish iteration? @trueqbit I'd like to know your opinion on it.

IMO this is the single best answer:
Log when the iterator is created [in `abc_view::begin()`] and when the last iterator is destroyed, using a special sqlite3_stmt deleter for the `abc_iterator::stmt` member.

dev/util.h
fnc12: done

dev/util.h
fnc12: fixed

dev/util.h
fnc12: fixed

dev/util.h
fnc12: done

dev/storage_base.h
fnc12: sure. Done

dev/util.h
fnc12: done

dev/util.h
fnc12: removed

dev/util.h
fnc12: fixed

fnc12:
> > The question is when do we have to call loggers during iteration call? When we create an iterable object? When we start > > iteration? When we finish iteration? @trueqbit I'd like to know your opinion on it.
>
> IMO this is the single best answer:
> Log when the iterator is created [in abc_view::begin()] and when the last iterator is destroyed, using a special sqlite3_stmt > deleter for the abc_iterator::stmt member.

I think the better option is to add will callback in `mapped_view` ctor and to add did callback in `mapped_view` dtor cause iterators can be copied.

trueqbit:
> > > The question is when do we have to call loggers during iteration call? When we create an iterable object? When we start > > iteration? When we finish iteration? @trueqbit I'd like to know your opinion on it.
> > 
> > 
> > IMO this is the single best answer:
> > Log when the iterator is created [in abc_view::begin()] and when the last iterator is destroyed, using a special sqlite3_stmt > deleter for the abc_iterator::stmt member.
> 
> I think the better option is to add will callback in `mapped_view` ctor and to add did callback in `mapped_view` dtor cause iterators can be copied.

I don't think this will work.
1. The view is designed
  a. to be lightweight and copyable.
  b. create the real sqlite3 statement in `begin()`.
  c. the iterators can outlive the view (borrowed range). Even if we juggle around things the view cannot just log the SQL statement assuming that iteration is completed.
2. The result set iterator cannot be copied, it's a move-only type.
3. Yes, the iterator of a mapped view can be copied. But it's an input iterator and it's not wise to keep iterators around.

fnc12:
> > > > The question is when do we have to call loggers during iteration call? When we create an iterable object? When we start iteration? When we finish iteration? @trueqbit I'd like to know your opinion on it.
> > > 
> > > IMO this is the single best answer:
> > > Log when the iterator is created [in abc_view::begin()] and when the last iterator is destroyed, using a special sqlite3_stmt > > > > deleter for the abc_iterator::stmt member.
> > 
> > I think the better option is to add will callback in mapped_view ctor and to add did callback in mapped_view dtor cause 
> > iterators can be copied.
>
> I don't think this will work.
> 
> The view is designed
> a. to be lightweight and copyable.
> b. create the real sqlite3 statement in begin().
> c. the iterators can outlive the view (borrowed range). Even if we juggle around things the view cannot just log the SQL statement assuming that iteration is completed.
> The result set iterator cannot be copied, it's a move-only type.
> Yes, the iterator of a mapped view can be copied. But it's an input iterator and it's not wise to keep iterators around.

Wait the view is definitely not intended to be copied or stored. Yes it is a lightweight entity which can be technically copied but I never saw any reason to do it. View is an intermediate object between storage and iterators, it is created before iteration and destroyed when iteration is over. That's why I think the best way is to react on view lifecycle events.

Anyway if this is such a non-clear topic let's merge logger as is. We can add logging for iteration later with no side effects. What do you think?

dev/util.h
```diff
@@ -53,82 +58,103 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
 
 namespace sqlite_orm {
     namespace internal {
+
+        template<class L>
```
trueqbit, line 62: In commit [Maximum efficient result row stepping loop](https://github.com/fnc12/sqlite_orm/pull/1412/commits/aaa11e5446ec84571abafd056ab36a246df41c2f) I updated the loop over a rowset.

Sorry that I am being a bit of a PITA here, but: I looked at the compiler output of several compilers and they all produced less code. So why not take advantage of the benefits, which in this case are free and add up?

If we could get back to the implementation of this commit for `perform_step()` and `perform_steps()`, then simply use `internal::perform_steps()` in `sqlite_executor::perform_steps()`:
```c++
        template<class L>
        void perform_step(sqlite3_stmt* stmt, L&& lambda) {
            switch (int rc = sqlite3_step(stmt)) {
                case SQLITE_ROW: {
                    lambda(stmt);
                } break;
                case SQLITE_DONE:
                    return;
                default: {
                    throw_translated_sqlite_error(rc);
                }
            }
        }

        template<class L>
        void perform_steps(sqlite3_stmt* stmt, L&& lambda) {
            for (;;) {
                switch (int rc = sqlite3_step(stmt)) {
                    case SQLITE_ROW: {
                        lambda(stmt);
                    } break;
                    case SQLITE_DONE:
                        return;
                    default: {
                        throw_translated_sqlite_error(stmt);
                    }
                }
            }
        }
```

Alternatively, I can also create another PR after this one if it seems too complicated.

dev/util.h
trueqbit: Sorry for being unclear and a bit wrong of what I suggested.

Simply use `internal::perform_steps()` as per review comment above.

trueqbit:
> > [...]
> > The view is designed
> > a. to be lightweight and copyable.
> > b. create the real sqlite3 statement in begin().
> > c. the iterators can outlive the view (borrowed range). Even if we juggle around things the view cannot just log the SQL statement assuming that iteration is completed.
> > The result set iterator cannot be copied, it's a move-only type.
> > Yes, the iterator of a mapped view can be copied. But it's an input iterator and it's not wise to keep iterators around.
> 
> Wait the view is definitely not intended to be copied or stored. Yes it is a lightweight entity which can be technically copied but I never saw any reason to do it. View is an intermediate object between storage and iterators, it is created before iteration and destroyed when iteration is over. That's why I think the best way is to react on view lifecycle events.
> 
> Anyway if this is such a non-clear topic let's merge logger as is. We can add logging for iteration later with no side effects. What do you think?

A view itself may be copyable, and sqlite_orm's result views definitely are at the moment. If it is copyable then the copy operation must have constant time complexity. See [C++ view concept](https://en.cppreference.com/w/cpp/ranges/view).
In general, it is a valid use case to capture a view expression in a variable and reuse it.

In addition, the iterators are not necessarily bound to the lifetime of the view.

So, I think sqlite_orm's result row views and iterators are well-designed., and for now we should leave that open for discussion.

dev/util.h
fnc12: reverted as far as I can (I need `perform_step` to return `int` now)

dev/util.h
fnc12: done. Please check

dev/util.h
trueqbit: Hmm, I don't see where you need the return value of `perform_step()`?

trueqbit (approved):

dev/util.h
fnc12: oh you're right. What I was saying is already removed. Fixed

Issue #707, opened 2021-04-23
How to use in multi-threaded mode?


NewYaroslav:
Hey! I am unable to use the library in multi-threaded mode

After creating the database, I do the following methods
```
storage.open_forever();
storage.pragma.journal_mode(journal_mode::WAL);
storage.busy_timeout(10);
storage.sync_schema();
```
A separate stream periodically records:
```
storage.transaction([&]() mutable {
    for (const Tick& t : transaction_buffer) {
        storage.replace(t);
    }
    return true;
});
```

The main thread is trying to read data
```
first_tick = storage.get_all<Tick>(
                            where(c(&Tick::timestamp) <= timestamp_ms),
                            order_by(&Tick::timestamp).desc(),
                            limit(1));
```

This causes the program to crash.

If you add a mutex to the code to eliminate the possibility of simultaneous reading and writing, then it will be impossible to read after writing. Reading will hang.

Plus, if a library can't read while writing, what's the point of WAL?

NewYaroslav:
I tried with all the options SQLITE_THREADSAFE (0,1,2)

fnc12:
`sqlite_orm` does nothing about multithreading. If you have multithreaded issues using `sqlite_orm` then you'd have the same issues if you use raw libsqlite3 library. To eliminate this issue just google how to use libsqlite3 in multithreaded environment and use `sqlite_orm` as is. `sqlite_orm` is just a frontend for SQLite3: it does not manage threads or data for you. It just takes AST you provided, serializes it to a raw string SQL and sends it to SQLite3 engine.

NewYaroslav:
I don't know how your library works internally. Here is what I found while looking for why reads are blocked while writing in WAL mode:
https://web-answers.ru/c/bd-sqlite-s-wal-zablokirovana-pri-podgotovke-quot.html
https://stackoverflow.com/questions/1680249/how-to-use-sqlite-in-a-multi-threaded-application

> check if you have sqlite3_reset after every sqlite3_step because this is one of the cases that leads to a database lock error.
> after preparing a statement with sqlite3_prepare and executing it with sqlite3_step, you must always reset it with sqlite3_reset.
> 
> Sqlite3_reset (S) interface resets prepared statement S back to
> the beginning of his program.
> 
> hope this solves your problem ... !!!

[...]

PR #1412, opened 2025-03-23
Fix for race condition in relation to error handling for query errors

### LINKED ISSUE #1416: Race condition between the execution of a statement and the retrieval of the last error from SQLite
> [...] sqlite_orm is doing some thread-unsafe calls that it doesn't have to do, and could definitely use some improvement in this area.  For example, there is *lots* of code that looks like this:
> 
> ```C++
> if(sqlite3_step(stmt) == SQLITE_DONE) {
>     //  done..
> } else {
>     throw std::system_error(std::error_code(sqlite3_errcode(db), get_sqlite_error_category()),
>                             sqlite3_errmsg(db));
> }
> ```
> 
> which isn't thread safe because there's a race condition between the `sqlite3_step(stmt)` and the `sqlite3_errcode(db)/sqlite3_errmsg(db)` calls, which means the code and message in the exception aren't reliable when used multithreaded.
> 
> 
> If I was writing it using raw libsqlite3, then I would write it to be thread-safe by writing it as:
> ```C++
> auto rc = sqlite3_step(stmt);
> if (rc == SQLITE_DONE) {
>     // done..
> } else {
>     throw std::system_error(std::error_code(rc, get_sqlite_error_category()), sqlite3_errstr(rc));
> }
> ```
> 
> but I *can't* currently get that thread safety with sqlite_orm.

 _Originally posted by @jagerman in [#707](https://github.com/fnc12/sqlite_orm/issues/707#issuecomment-859649406)_

trueqbit:
This PR addresses a [comment by @jagerman](https://github.com/fnc12/sqlite_orm/issues/707#issuecomment-859649406).

The problem solved is the race condition between executing a statement and retrieving the last error from SQLite.
There are two ways to fix this: Either handle errors solely based on the return code of an API call. Or serialize access to the DB handle.

This PR implements handling errors based on the return code.

When handling errors solely based on the return code, SQLite obviously cannot provide as much information as to why (or where) a statement failed. I consider this as a flaw in SQLite as it would be easily solvable with storing the last error per thread in the SQLite3 library.

Important things to know:
* The library can handle empty ranges of insert/replace statements and unregistered function errors up-front. SQLite would only return a generic `SQLITE_ERROR` error value.
* Checking missing application-defined collation sequences wasn't pursued because SQLite will at least return a dedicated `SQLITE_ERROR_MISSING_COLLSEQ` extended error value. It would be easily implementable.
* Detailed information of constraint errors [`SQLITE_CONSTRAINT_*`], e.g. about the specific column violating a constraint (e.g. primary key, foreign key, not null) is lost; only the type of constraint violation is known.

Additional improvements:
* Enabled SQLite extended error codes
* Maximum efficient result row stepping loop. Returning instead of looping on a condition saves a couple of instructions.

### DIFF
--- a/dev/ast_iterator.h
+++ b/dev/ast_iterator.h
@@ -6,7 +6,9 @@
 #include <functional>  //  std::reference_wrapper
 #endif
 
+#include "functional/cxx_functional_polyfill.h"  //  std::is_invocable
 #include "tuple_helper/tuple_iteration.h"
+#include "functional/static_magic.h"
 #include "type_traits.h"
 #include "conditions.h"
 #include "alias.h"
@@ -68,12 +70,11 @@ namespace sqlite_orm {
         void iterate_ast(const T& t, L&& lambda) {
             ast_iterator<T> iterator;
 
-#if defined(SQLITE_ORM_IF_CONSTEXPR_SUPPORTED) && __cpp_lib_is_invocable >= 201703L
             // possibly invoke lambda with node itself
-            if constexpr (std::is_invocable<L, polyfill::bool_constant<true>, const T&>::value) {
-                lambda(polyfill::bool_constant<true>{}, t);
-            }
-#endif
+            call_if_constexpr<polyfill::is_invocable<L, polyfill::bool_constant<true>, const T&>::value>(
+                lambda,
+                polyfill::bool_constant<true>{},
+                t);
 
             iterator(t, lambda);
         }
@@ -126,7 +127,6 @@ namespace sqlite_orm {
 
             template<class L>
             void operator()(const node_type& expression, L& lambda) const {
-                lambda(expression);
                 iterate_ast(expression.argument0, lambda);
                 iterate_ast(expression.argument1, lambda);
                 iterate_ast(expression.argument2, lambda);
--- a/dev/ast_iterator.h
+++ b/dev/ast_iterator.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #ifndef SQLITE_ORM_IMPORT_STD_MODULE
+#include <type_traits>  //  std::is_invocable
 #include <vector>  //  std::vector
 #include <functional>  //  std::reference_wrapper
 #endif
@@ -36,7 +37,7 @@ namespace sqlite_orm {
          *  ast_iterator is used in finding literals to be bound to
          *  a statement, and to collect table names.
          *  
-         *  Note that not all leaves of the expression tree are always visited:
+         *  Note that not all leaves of the expression tree are visited:
          *  Column expressions can be more complex, but are passed as a whole to the callable.
          *  Examples are `column_pointer<>` and `alias_column_t<>`.
          *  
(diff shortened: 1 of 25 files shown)


dev/storage.h
```diff
@@ -1202,6 +1238,10 @@ namespace sqlite_orm {
 
             template<typename S>
             prepared_statement_t<S> prepare_impl(S statement) {
+#if defined(SQLITE_ORM_IF_CONSTEXPR_SUPPORTED) && __cpp_lib_is_invocable >= 201703L
+                iterate_ast(statement, udf_presence_checker{this->scalarFunctions, this->aggregateFunctions});
```
fnc12: we don't need to iterate AST one more time if `this->scalarFunctions.empty() && this->aggregateFunctions.empty()`. We can add `if` for that to skip increasing `prepare_impl` work time for people who never binds application defined funcs

dev/ast_iterator.h
```diff
@@ -52,17 +53,28 @@ namespace sqlite_orm {
...
+         *  The callable can opt in to be invoked for a node expression.
          */
         template<class T, class L>
         void iterate_ast(const T& t, L&& lambda) {
             ast_iterator<T> iterator;
+
+#if defined(SQLITE_ORM_IF_CONSTEXPR_SUPPORTED) && __cpp_lib_is_invocable >= 201703L
+            // possibly invoke lambda with node itself
+            if constexpr (std::is_invocable<L, polyfill::bool_constant<true>, const T&>::value) {
+                lambda(polyfill::bool_constant<true>{}, t);
```
fnc12: where do we use it?

trueqbit: The `udf_presence_checker` uses it.

dev/storage.h
trueqbit: Unfortunately skipping the AST iteration is impossible. We need to examine the AST for function calls, and throw an error if the function wasn't registered earlier.

dev/ast_iterator.h
trueqbit: ... and now also the table name collector.

dev/storage.h
fnc12: wait but if we have `this->scalarFunctions.empty() && this->aggregateFunctions.empty()` it means we find nothing

tests/ast_iterator_tests.cpp
```diff
@@ -10,16 +10,41 @@ using internal::column_alias;
 using internal::column_pointer;
 using internal::iterate_ast;
 
+#if __cpp_generic_lambdas >= 201707L
+template<class... L>
+struct overload : L... {
```
fnc12: this is the best part of C++ =)

fnc12 (approved):

dev/storage.h
trueqbit: You can write a statement that uses a function that you didn't register (create) before... So if both containers are empty it is an error - in order to find out whether a function call is in a statement we need to examine the statement.

fnc12: ah I see. Ok

PR #1414, opened 2025-03-24
added find_column_name to virtual table

### LINKED ISSUE #1410: nested query using virtual using_fts5 table
Hello

(full minimal example repo here : https://github.com/conradjones/sqlite_orm_fts_test/tree/main )

I'm having difficulty getting a query working on a normal table using a sub query (match) to select the IDs from the virtual fts5 table , to use to select the records in the primary normal table.

Its a contrived minimal example to demonstrate the point, 

Standard SQLite query would look like this
`SELECT * FROM normal_table where id in (SELECT normal_table_id FROM search_table WHERE text MATCH ?)`

Two tables / storage look like this 
```
struct NormalTable {
    int id;
    std::string text;
    int otherValue;
};

struct SearchTable {
    int normal_table_id;
    std::string text;
};

inline auto initStorage(const std::string &databaseFilePath) {
    return sqlite_orm::make_storage(
        databaseFilePath,
        sqlite_orm::make_table("normal_table",
                               sqlite_orm::make_column("id", &NormalTable::id,
                                                       sqlite_orm::primary_key().autoincrement()),
                               sqlite_orm::make_column("path", &NormalTable::text),
                               sqlite_orm::make_column("other_value", &NormalTable::otherValue)),

        sqlite_orm::make_virtual_table(
            "search_table",
            sqlite_orm::using_fts5(
                sqlite_orm::make_column("text", &SearchTable::text),
                sqlite_orm::make_column("normal_table_id", &SearchTable::normal_table_id)))
    );
}
````

any attempt to use them say like this 

```
auto rows = storage.iterate<NormalTable>(
        sqlite_orm::where(sqlite_orm::eq(&NormalTable::id,
                                         sqlite_orm::select(&SearchTable::normal_table_id,
                                                            sqlite_orm::where(
                                                                sqlite_orm::match<SearchTable>("Some Text"))))

        )


    );

 for (const auto &row: rows) {
    }
```

results in an error like 

```
error: no member named 'find_column_name' in 'sqlite_orm::internal::virtual_table_t<sqlite_orm::internal::using_fts5_t<db::SearchFile, sqlite_orm::internal::column_t<std::string db::SearchFile::*, sqlite_orm::internal::empty_setter>, sqlite_orm::internal::column_t<int db::SearchFile::*, sqlite_orm::internal::empty_setter>>>'
 12528 |             return pick_table<O>(dbObjects).find_column_name(field);
```

### PR DESCRIPTION
**Pull request by fnc12**
Fixing #1410

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,12 @@ target_sources(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/incl
 
 target_include_directories(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
 
+# Enable reorder-ctor warning and make it an error
+target_compile_options(sqlite_orm INTERFACE
+    -Wreorder-ctor
+    -Werror=reorder-ctor
+)
+
 include(ucm)
 
 if (MSVC)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,10 +55,10 @@ target_sources(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/incl
 
 target_include_directories(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
 
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
     target_compile_options(sqlite_orm INTERFACE
-        -Wreorder-ctor
-        -Werror=reorder-ctor
+        -Wreorder
+        -Werror=reorder
     )
 endif()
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: tests/schema/virtual_table.cpp [hunk 1]
```diff
@@ -20,11 +20,22 @@ TEST_CASE("virtual table") {
 #endif
     };
 
+    auto virtualTable =
+        make_virtual_table("posts", using_fts5(make_column("title", &Post::title), make_column("body", &Post::body)));
+    {
+        auto compareColumnName = [](const std::string* foundValue, std::string expectedValue) {
```
trueqbit: Nit: I usually `constexpr` or at least `const` qualify a lambda variable.

trueqbit (approved):

tests/schema/virtual_table.cpp
fnc12: good point. Let me add it next time

fnc12: added

CMakeLists.txt
```diff
@@ -55,6 +55,13 @@ target_sources(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/incl
 
 target_include_directories(sqlite_orm INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
```
trueqbit, line 58: The equivalent for msvc is `/w15038` and `/we5038`.

BTW I somehow question whether it's a good idea to introduce this in the main CMake target (i.e. library). This way you modify someone's build settings.

It would be enough to only enable this warning/error for unit tests and examples.

fnc12: I was sure it works only within `sqlite_orm` and is not spreading all over the project which uses `sqlite_orm`. I it works differently I can remove it

PR #1417, opened 2025-03-26
C++17 round 2: constexpr if statements

trueqbit:
The following C++ language features a now a requirement:

* constexpr if statements. See corresponding entry in the [compiler support matrix](https://en.cppreference.com/w/cpp/compiler_support/17#cpp_if_constexpr_201606L).
* `[[maybe_unused]]` attribute


dev/ast_iterator.h
```diff
@@ -71,10 +70,9 @@ namespace sqlite_orm {
             ast_iterator<T> iterator;
 
             // possibly invoke lambda with node itself
-            call_if_constexpr<polyfill::is_invocable<L, polyfill::bool_constant<true>, const T&>::value>(
-                lambda,
-                polyfill::bool_constant<true>{},
-                t);
+            if constexpr (polyfill::is_invocable<L, polyfill::bool_constant<true>, const T&>::value) {
```
fnc12, line 73: this PR breaks C++14 compatibility, right?

dev/cte_column_names_collector.h
```diff
@@ -159,8 +159,9 @@ namespace sqlite_orm {
             std::vector<std::string> columnNames = get_cte_column_names(sel.col, context);
 
             // 2. override column names from cte expression
-            if (size_t n = std::tuple_size_v<ExplicitColRefs>) {
-                if (n != columnNames.size()) {
+            constexpr size_t nExplicitColumns = std::tuple_size_v<ExplicitColRefs>;
+            if constexpr (nExplicitColumns > 0) {
```
fnc12, line 163: this improves the lib a lot

dev/functional/static_magic.h
```diff
@@ -24,58 +17,6 @@ namespace sqlite_orm {
         };
         template<class R = void>
         constexpr empty_callable_t<R> empty_callable{};
```
fnc12: I think we don't this as well just like the whole file

dev/implementations/storage_definitions.h
```diff
@@ -22,18 +22,25 @@
...
                 return sync_schema_result::already_in_sync;
+            } else {
+                return this->_sync_normal_table(table, db, preserve);
             }
-#endif  //  SQLITE_ENABLE_DBSTAT_VTAB
+        }
+
+        template<class... DBO>
+        template<class Table, satisfies<is_table, Table>>
+        sync_schema_result storage_t<DBO...>::_sync_normal_table(const Table& table, sqlite3* db, bool preserve) {
```
fnc12: maybe `sync_generic_table`

dev/statement_serializer.h
```diff
@@ -1787,23 +1788,22 @@ namespace sqlite_orm {
...
                     }
                     ss << "SELECT ";
-                    call_if_constexpr<is_rowset_deduplicator_v<return_type>>(
+                    if constexpr (is_rowset_deduplicator_v<return_type>) {
                         // note: make use of implicit to-string conversion
-                        [&ss](std::string keyword) {
-                            ss << keyword << ' ';
-                        },
-                        sel.col);
+                        std::string keyword = sel.col;
```
fnc12, line 1798: do we have `std::string` copying here?

dev/storage.h
```diff
@@ -1211,6 +1208,9 @@ namespace sqlite_orm {
             template<class Table, satisfies<is_table, Table> = true>
             sync_schema_result sync_table(const Table& table, sqlite3* db, bool preserve);
 
+            template<class Table, satisfies<is_table, Table> = true>
+            sync_schema_result _sync_normal_table(const Table& table, sqlite3* db, bool preserve);
```
fnc12: func name starting with underscore looks very odd. I think regular `sync_normal_table` is well enough

fnc12 (approved):

dev/ast_iterator.h
trueqbit: Correct. C++17 is required with regard to the language standard. Further updates for the rest of C++17 that sqlite_orm uses is already in the pipeline.

dev/functional/static_magic.h
trueqbit: It's still in use for `statement_serializer<replace_t<T>>`.
It actually has nothing to do with "static magic", but is a useful variadic function that always returns a specific value. I can rename it? We already have a metafunction called `always_false` and a quoted metafunction `mpl::always`. I would choose `always_default`, which probably fits best..

dev/statement_serializer.h
trueqbit: Yes, see comment above, which didn't change.

dev/functional/static_magic.h
fnc12: > I can rename it?

up to you

dev/implementations/storage_definitions.h
trueqbit: You probably mean "general". I chose "regular".

fnc12: ok

fnc12 (approved):

dev/functional/static_magic.h
trueqbit: Renamed and moved to `functional/always_default.h`

dev/select_constraints.h
```diff
@@ -270,30 +270,30 @@ namespace sqlite_orm {
 
         template<typename Moniker, class ExplicitCols>
         struct cte_builder {
-            ExplicitCols explicitColumns;
+            ExplicitCols _explicitColumns;
```
fnc12, line 273: do you want to replace all `this->` places with `_` prefix?

fnc12 (approved):

dev/select_constraints.h
trueqbit: Do you mean across the entire library? That would be especially good for the internally created classes with public members. I would save that for another day, is that okay?

fnc12: Yes I mean across the entire lib. I am asking in terms of your global plans actually cause I am a fan if `this->` and removing it triggers me a bit. On another hand I am ok with the leading underscore. But combining both looks ugly so I am just asking.

> I would save that for another day, is that okay?

Not a problem at all.

trueqbit: I am not a fan of `this->` at all, it makes things much harder to read and is usually not necessary from a syntax perspective :)
But I am OK with either style.

So we should get together one day and find a guideline for the coding style that everyone follows.
By the way, I was not the one who introduced the leading underscore. For structs as public interfaces (even if they are from the internal namespace) the leading underscore is important, as you already mentioned. Firstly because of sorting in an IDE, and secondly to easily distinguish between public and private members.

PR #1422, opened 2025-04-08
#1420 : add platform definition macros

### LINKED ISSUE #1420: Add platform definition macros
As discussed in #1393, it would be nice to expand the platform definition macros to handle the mobile platforms as well. I intend to handle the following:

- Android
- iOS
- Linux
- BSD

I will begin a new branch and open a PR with these changes shortly.

ancientjpeg:
Resolves #1420. 

## Notes

- I added the VfxWorks platform detection as SQLite supports it and has distinct VFS objects specifically for that operating system. I copied their platform detection logic for it
- `SQLITE_ORM_IOS` will be defined for all of `iOS|watchOS|tvOS` which I believe is suitable as they share very similar OS APIs. There should be virtually no difference from the perspective of this library, so it'll be up to users to make those differentiations if they need to.


dev/functional/config.h
```diff
@@ -72,13 +72,39 @@
...
 #define SQLITE_ORM_APPLE
 #define SQLITE_ORM_UNIX
-#elif defined(__unix__) || defined(__unix) || defined(__linux__) || defined(__FreeBSD__)
+
+#elif defined(__linux__)
+#if defined(__ANDROID__)
+#define SQLITE_ORM_ANDROID
+#endif
+#define SQLITE_ORM_LINUX
+#define SQLITE_ORM_UNIX
```
ancientjpeg: `__linux__` should always be defined on android systems so I figured this was safe.

ancientjpeg:
Sorry about the multiple commits in rapid succession there—just a heads up, `clang-format` version 20 appears to be ignoring a rule about spacing the `&` symbol after a declaration.

dev/functional/config.h
```diff
@@ -72,13 +72,39 @@
 
 #define SQLITE_ORM_WITH_CTE
 
-#if defined(_WIN32)
+#if defined(_WIN32) || defined(_WIN64)
```
trueqbit: Checking for `_WIN64` is unnecessary as `_WIN32` is [always defined](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros). Only `_WIN32` should be used for detecting the operating system.

dev/functional/config.h
trueqbit, line 1: It's getting a bit lengthy... :)

I suggest splitting architecture/platform detection into a separate file just like we have for testing C++ features, e.g. _platform_config.h_.

ancientjpeg:
Hey @trueqbit, I've implemented those changes you asked for.

fnc12 (approved):

dev/vfs_name.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "serialize_result_type.h"
+#include "functional/platform_definitions.h"
```
trueqbit: Oh, I meant that this header file should be #included in _config.h_ so that it is universally available.

trueqbit (changes requested):

trueqbit (approved):

trueqbit:
Thx @ancientjpeg !

PR #1426, opened 2025-05-14
moved cast_t to ast folder, added unit tests artifacts to gitignore a…

fnc12:
1) moved `cast_t` to `ast` folder to a dedicated file 
2) added unit test launch artifacts to gitignore which happens when you run unit tests from repo root not from `build` folder
3) added several `std::remove` to reduce creating unit tests artifacts. It may not run always cause in case of exception file will be preserved but anyway it reduces amount of artifacts

### DIFF
--- a/dev/ast/cast.h
+++ b/dev/ast/cast.h
@@ -5,6 +5,7 @@
 #include <utility>  // std::move
 #endif  //  SQLITE_ORM_IMPORT_STD_MODULE
 namespace sqlite_orm {
+
     namespace internal {
 
         /**
@@ -19,8 +20,6 @@ namespace sqlite_orm {
             using expression_type = E;
 
             expression_type expression;
-
-            cast_t(expression_type expression_) : expression(std::move(expression_)) {}
         };
     }
 }
--- a/dev/ast/cast.h
+++ b/dev/ast/cast.h
@@ -2,7 +2,6 @@
 
 #ifndef SQLITE_ORM_IMPORT_STD_MODULE
 #include <string>  // std::string
-#include <utility>  // std::move
 #endif  //  SQLITE_ORM_IMPORT_STD_MODULE
 namespace sqlite_orm {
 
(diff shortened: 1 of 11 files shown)


dev/ast/cast.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#ifndef SQLITE_ORM_IMPORT_STD_MODULE
+#include <string>  // std::string
+#include <utility>  // std::move
```
trueqbit, line 4: This header doesn't need `std::string` anymore. (Comment applies to line above of course)

dev/ast/cast.h
```diff
@@ -0,0 +1,37 @@
...
+         *  Example: cast<std::string>(&User::id)
+         */
+        template<class T, class E>
+        struct cast_t {
+            using to_type = T;
+            using expression_type = E;
+
+            expression_type expression;
+
+            cast_t(expression_type expression_) : expression(std::move(expression_)) {}
```
trueqbit: The constructor could be removed.

dev/ast/cast.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#ifndef SQLITE_ORM_IMPORT_STD_MODULE
+#include <string>  // std::string
+#include <utility>  // std::move
+#endif  //  SQLITE_ORM_IMPORT_STD_MODULE
+namespace sqlite_orm {
```
trueqbit, line 6: Nit: An empty line before is missing.

trueqbit (approved):

dev/ast/cast.h
fnc12: it differs from std implementation so I'd not rely on sub-includes. Especially if you create a C++ program on macOS, them compiled it for Linux and get includes missing errors

dev/ast/cast.h
fnc12: will be fixed in the next commit

dev/ast/cast.h
fnc12: will be removed in the next commit

dev/ast/cast.h
trueqbit: I don't quite understand. I meant including <string> isn't necessary and should be removed.

PR #1428, opened 2025-05-15
Static call operator

trueqbit:
This PR makes use of new C++23 and C++26 features:

1. Static call operator of lambdas and function objects.
  This might save code size - for the unit test executable of this PR the saving for "Debug" builds is 558.5 KB, which I find quite a lot for such an unspectacular feature!
  Though it doesn't have this great effect in optimized "Release" builds, I like to be explicit about the logical intention.
2. Tuples can be unpacked into a structured binding pack instead of using index sequences, which simplifies operations on tuple elements.

Additionally:
* Corrected exporting symbols (from the named C++ module) that are forward-declared in _dev/function.h_.
* Corrected binding BLOB data in ` statement_binder<std::vector<char>>::bind()`.
* Corrected throwing SQL statement errors in `perform_step()`.
* Corrected datatype conversion warning in `sqlite_executor::perform_exec()`


dev/statement_binder.h
```diff
@@ -329,33 +329,43 @@ namespace sqlite_orm {
 
             explicit tuple_value_binder(sqlite3_stmt* stmt) : stmt{stmt} {}
 
+#ifdef SQLITE_ORM_STRUCTURED_BINDING_PACK_SUPPORTED
+            void operator()(const auto& tpl, auto project) const {
+                int nthSqlParameter = 0;
+                auto& [... elements] = tpl;
+                (this->bind(std::invoke(project, elements), ++nthSqlParameter), ...);
```
fnc12, line 336: don't we need to move `project` here or pass by ref?

dev/statement_serializer.h
```diff
@@ -93,38 +94,38 @@ namespace sqlite_orm {
 #endif
                                       ,
                                       bool> = true>
-            std::string do_serialize(const X& c) const {
+            static std::string do_serialize(const X& c) {
```
fnc12, line 97: what's the point here? why do we refuse using objects and use static calls just like in pure C?

dev/storage.h
```diff
@@ -128,7 +128,7 @@ namespace sqlite_orm {
 
             // swallow leaf expressions
             template<class T>
-            void operator()(const T&) const {}
+            SQLITE_ORM_STATIC_CALLOP void operator()(const T&) SQLITE_ORM_OR_CONST_CALLOP {}
```
fnc12: nit: probably `udf_presence_checker` has to be moved to a dedicated file in future to avoid confusion that it is located in `storage.h` and that `operator()` is located in `storage_t`

dev/util.h
```diff
@@ -110,7 +110,7 @@ namespace sqlite_orm {
                                      const std::string& query,
                                      int (*callback)(void* data, int argc, char** argv, char**),
                                      void* user_data) const {
-                return perform_exec(db, query.data(), callback, user_data);
+                return perform_exec(db, query.c_str(), callback, user_data);
```
fnc12, line 113: wha's the difference?

fnc12 (approved):

dev/statement_binder.h
trueqbit: No. `std::invoke()` takes the callable `project` by reference.

dev/statement_serializer.h
trueqbit: I don't exactly understand the comparison to "pure C".

The whole point of this PR is to make use of C++23's "static call operators" of lambdas and function objects. This means that no `this` needs to be passed around, which would be unused anyway. Good for compiler and optimizer.

So, as an implementation detail, the private `do_serialize()` methods need to be static methods as well.

dev/util.h
trueqbit: TBH nothing but expressing intent, find it a bit clearer. But I am not objecting to revert it if you feel that I am stepping on your code :)

dev/storage.h
trueqbit: There's no `operator()` located in `storage_t`? It's defined right at the line that you commented on. Otherwise can you point me to where this would be and what you mean exactly?
The `udf_presence_checker` is currently only used by `storage_t::prepare_impl()`, so it makes sense to leave it where it is.

dev/util.h
fnc12: just curious. Thanks

dev/storage.h
fnc12: I am describing the order of thoughts reading this PR. I see `dev/storage.h` has changes in `operator()` and trying to remember whether `operator()` exists in `storage_t`. Imagine if the file is named `dev/udf_presence_checker.h` - it would be easier to read and understand it with the lesser amount of thoughts. Anyway it is just a nit

trueqbit: You're right, let's keep up the good work and put things in their proper place.

fnc12 (approved):

PR #1431, opened 2025-05-19
Started using C++ Core Guidelines Support facilities

trueqbit:
The goal is to use C++ type aliases for raw pointers and (null-terminated) C strings where appropriate according to the C++ Core Guidelines.
While the semantics cannot be changed - which is good - this improves the expression of intent and potentially increases the coverage of code analysis in the future and the understanding of security relevant aspects.

### DIFF
--- a/dev/row_extractor.h
+++ b/dev/row_extractor.h
@@ -122,7 +122,7 @@ namespace sqlite_orm {
 
 SQLITE_ORM_EXPORT namespace sqlite_orm {
     template<class R>
-    int extract_single_value(void* data, int argc, char** argv, char**) {
+    int extract_single_value(void* data, int argc, orm_gsl::zstring* argv, orm_gsl::zstring*) {
         auto& res = *(R*)data;
         if (argc) {
             const auto rowExtractor = internal::column_text_extractor<R>();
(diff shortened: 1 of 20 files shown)


dev/row_extractor.h
```diff
@@ -232,16 +231,16 @@ namespace sqlite_orm {
         }
 
         T extract(sqlite3_stmt* stmt, int columnIndex) const {
-            if (auto cStr = (const char*)sqlite3_column_text(stmt, columnIndex)) {
-                return cStr;
+            if (orm_gsl::czstring text = (const char*)sqlite3_column_text(stmt, columnIndex)) {
```
fnc12, line 234: why don't we put `orm_gsl::czstring` inside cast operator before `sqlite3_column_text`?

fnc12 (approved):

dev/row_extractor.h
trueqbit: Very good question.
`sqlite3_column_text()` returns a utf-8 byte sequence as `const unsigned char*`, which can be cast to `const char*`. I just thought it is clearer to separate it into 2 steps. That was just my train of thought at that time, let me know what you think.

fnc12: ok

PR #1434, opened 2025-05-23
Allow retrieval of all existing triggers (issue #1429)

HansH:
This pull requests adds a `trigger_names` method, that can be used to list all existing triggers in the database.

### DIFF
--- a/dev/storage_base.h
+++ b/dev/storage_base.h
@@ -755,23 +755,19 @@ namespace sqlite_orm {
                 return res;
             }
 
-            std::vector<std::string> object_names(const std::string& type) {
+            std::vector<std::string> object_names(string_constant_type type) {
                 using data_t = std::vector<std::string>;
 
                 auto connection = this->get_connection();
                 data_t objectNames;
                 std::stringstream ss;
-                ss << "SELECT name FROM sqlite_master WHERE type = " << quote_string_literal(type);
+                ss << "SELECT name FROM sqlite_master WHERE type=" << quote_string_literal(type);
                 this->executor.perform_exec(
                     connection.get(),
                     ss.str(),
                     [](void* data, int argc, orm_gsl::zstring* argv, orm_gsl::zstring* /*columnName*/) -> int {
                         auto& objectNames_ = *(data_t*)data;
-                        for (int i = 0; i < argc; ++i) {
-                            if (argv[i]) {
-                                objectNames_.emplace_back(argv[i]);
-                            }
-                        }
+                        objectNames_.emplace_back(argv[0]);
                         return 0;
                     },
                     &objectNames);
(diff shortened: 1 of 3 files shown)


dev/storage_base.h
```diff
@@ -764,6 +755,30 @@ namespace sqlite_orm {
                 return res;
             }
 
+            std::vector<std::string> object_names(const std::string& type) {
```
fnc12: nit: one can accept `std::string_view` instead of `std::string` to reduce amount of dynamic allocations

fnc12 (approved):

fnc12:
@HansH thanks for this PR. Please fix formatting using `clang-format`. More details can be found here https://github.com/fnc12/sqlite_orm/wiki/Contribution-guide

dev/storage_base.h
```diff
@@ -764,6 +755,30 @@ namespace sqlite_orm {
...
+                auto connection = this->get_connection();
+                data_t objectNames;
+                std::stringstream ss;
+                ss << "SELECT name FROM sqlite_master WHERE type = " << quote_string_literal(type);
+                this->executor.perform_exec(
+                    connection.get(),
+                    ss.str(),
+                    [](void* data, int argc, orm_gsl::zstring* argv, orm_gsl::zstring* /*columnName*/) -> int {
+                        auto& objectNames_ = *(data_t*)data;
+                        for (int i = 0; i < argc; ++i) {
```
trueqbit: While you are at it: Can we get rid of that loop, please. There's only one column in the result row. There's no need for a generic callback implementation.

dev/storage_base.h
```diff
@@ -764,6 +755,30 @@ namespace sqlite_orm {
...
+                            if (argv[i]) {
```
trueqbit: Can we also get rid of that check, please? SQLite will ever only return names of existing table. There's no need for a generic callback implementation.

dev/storage_base.h
trueqbit: Let's choose `string_constant_type`: `std::vector<std::string> object_names(string_constant_type type)`

dev/storage_base.h
```diff
@@ -764,6 +755,26 @@ namespace sqlite_orm {
                 return res;
             }
 
+            std::vector<std::string> object_names(string_constant_type type) {
+                using data_t = std::vector<std::string>;
+
+                auto connection = this->get_connection();
+                data_t objectNames;
+                std::stringstream ss;
+                ss << "SELECT name FROM sqlite_master WHERE type=" << quote_string_literal(type);
```
trueqbit: You need to explicitly construct a `std::string`:
```c++
quote_string_literal(std::string(type))
```

trueqbit (approved):

fnc12 (approved):

fnc12:
@HansH thanks!

PR #1440, opened 2025-06-07
Improving quoted functions et al

trueqbit:
This PR improves the handling of scalar application-defined functions from quoted functions and those with a static call operator.

* Support for static call operators (of quoted lambdas and function objects, and of traditional scalar function classes).
* Overloaded call operators are now picked by the explicitly specified signature. [Note: This was partially done before at the point when creating the quoted function. It is now picked as well when invoking the actual call operator. This also restores a [deleted type alias](https://github.com/fnc12/sqlite_orm/pull/1439/commits/4233cf64179db008c72eac0b9641e03e590d9229#diff-bec2e0d185434e23554ae853bf5b507efc69da2d88ddcfe8b5ab378d6b08201f) as a proper static assertion.]
* Added unit tests for scalar functions with static call operators.

Additionally:
* Added an example of a quoted application-defined function.
* Simplified tuple from boxed SQLite values.
* CMake: Fallback to C++17 if no explicit C++ standard is requested.
* CMake: Added build variable to request C++26


dev/values_to_tuple.h
```diff
@@ -2,43 +2,41 @@
...
-            }
-        };
-    }
+    template<class Tpl>
+    struct tuple_from_values {
+        SQLITE_ORM_STATIC_CALLOP Tpl operator()(sqlite3_value** values, int nValues) SQLITE_ORM_OR_CONST_CALLOP {
+#ifdef SQLITE_ORM_CONTRACTS_SUPPORTED
+            contract_assert(nValues == std::tuple_size<Tpl>::value);
+#else
+            (void)nValues;
```
fnc12: nit: I'd use `std::ignore` to make it more modern

examples/user_defined_functions.cpp
```diff
@@ -148,13 +177,12 @@ int main() {
...
     //  SELECT ASUM(a), ASUM(b), ASUM(c)
     //  FROM t
-    auto aSumRows =
-        storage.select(columns(accelerated_sum(&Table::a), accelerated_sum(&Table::b), accelerated_sum(&Table::c)));
     cout << "SELECT ASUM(a), ASUM(b), ASUM(c) FROM t:" << endl;
-    for (auto& row: aSumRows) {
-        cout << '\t' << get<0>(row) << endl;
-        cout << '\t' << get<1>(row) << endl;
-        cout << '\t' << get<2>(row) << endl;
+    for (auto [a, b, c]: storage.iterate(
```
fnc12, line 181: this looks a lot better and attractive

fnc12 (approved):

dev/values_to_tuple.h
trueqbit: I don't know. I am not convinced this is the right usage of `ignore`. It should only be used for ignoring results (of functions).

fnc12: well nobody prohibits doing it =)

fnc12 (approved):

PR #1452, opened 2025-09-17
Type-erased iteration of result sets

trueqbit:
Type-erased iteration of result sets

Like originally asked in an [inline question](https://github.com/fnc12/sqlite_orm/pull/1311#issuecomment-2181411844) [[answer](https://github.com/fnc12/sqlite_orm/issues/1311#issuecomment-2182267282)] it can be desirable to pass a lazily generated query result across library boundaries. This library boundary should be independent of template arguments only used for traditional C++ views and iterators.

With C++23 we can make use of a coroutine-based generator for ranges.

Additionally to the `iterate()` method sqlite_orm's "storage" class now comes with a `yield()` method that returns a `std::generator<result>`.

I have chosen `yield` for the method name to create the range generator, as I didn't like `generate`, and `yield` says what it does.
Another good name for the new method could be `query`.

We have to agree on a name for the new "storage" method:
`yield`
`query`
`generate`
`produce`


fnc12 (approved):

fnc12:
`yield` is a really good name @trueqbit . Thanks

PR #1455, opened 2025-09-26
Generalize and streamline code base

trueqbit:
While working on [sql views](../tree/experimental/sql-view), I came across some parts that I thought would be worth generalizing and streamlining the code base before making other, more profound changes.

In this PR:
* Removed superfluous `serializer_context_builder`.
* Renamed serializer context flag variables - e.g. omitting column types with the `fts5_column` flag will be needed for views as well.
* Encapsulated logging executes SQL statements in all locations, while step functions are kept in their respective functions for greater clarity.
* Select statements and "raw" DML (Data Modification Language) statements are now treated uniformly, regardless of whether they are used directly or embedded as the main statement in a With clause.
* Used `c_str()` instead of `data()` to more clearly express the intent where it makes sense.p
* Simplified code involving column filters with a new alias template.
* Renamed `storage_t<>::sync_table()` -> `storage_t<>::sync_dbo()` as this method is overloaded for all possible database object types.

### DIFF
--- a/dev/select_constraints.h
+++ b/dev/select_constraints.h
@@ -421,10 +421,8 @@ namespace sqlite_orm {
         constexpr decltype(auto) access_main_select(const Select& select) {
             if constexpr (is_with_clause_v<Select>) {
                 return (select.expression);
-            } else if constexpr (is_select_v<Select>) {
-                return select;
             } else {
-                static_assert(polyfill::always_false_v<Select>);
+                return select;
             }
         }
 
--- a/dev/select_constraints.h
+++ b/dev/select_constraints.h
@@ -318,7 +318,16 @@ namespace sqlite_orm {
                 }
             }
         };
+
+        template<class T>
+        inline constexpr bool is_with_clause_v = polyfill::is_specialization_of<T, with_t>::value;
+#else
+        template<class T>
+        inline constexpr bool is_with_clause_v = false;
 #endif
+
+        template<class T>
+        using is_with_clause = polyfill::bool_constant<is_with_clause_v<T>>;
 
         template<class T>
         struct asterisk_t {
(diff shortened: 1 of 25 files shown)


dev/select_constraints.h
```diff
@@ -388,6 +397,38 @@ namespace sqlite_orm {
...
+            With,
+            std::enable_if_t<polyfill::conjunction_v<is_with_clause<With>, is_select<expression_type_t<With>>>>> = true;
+
+        /*  
+         *  Access the main select expression of a with clause or the passed in select expression.
+         */
+        template<class Select, satisfies<is_select_expression, Select> = true>
+        constexpr decltype(auto) access_main_select(const Select& select) {
+            if constexpr (is_with_clause_v<Select>) {
+                return (select.expression);
```
fnc12, line 423: this returns a ref cause of parenthesis and `decltype(auto)`, lower `return select;` returns value's copy. Is it what you intended to do here?

dev/serializer_context.h
```diff
@@ -19,20 +19,6 @@ namespace sqlite_orm {
 
             serializer_context(const db_objects_type& dbObjects) : db_objects{dbObjects} {}
         };
-
-        template<class S>
-        struct serializer_context_builder {
```
fnc12, line 24: it is very good that we've removed this extra class. Less abstractions - easier code

dev/util.h
```diff
@@ -78,90 +115,96 @@ namespace sqlite_orm {
             std::function<void(serialize_arg_type sql)> will_run_query;
             std::function<void(serialize_arg_type sql)> did_run_query;
 
-            inline void perform_void_exec(sqlite3* db, orm_gsl::czstring sql) const {
+            void perform_void_exec(sqlite3* db, orm_gsl::czstring sql) const {
```
fnc12, line 118: removing `inline` may cause symbol duplicate errors during compilation cause non-template non-inline funcs better be located at `.cpp` files not in headers. Same with other inlines within this file

fnc12 (approved):

dev/util.h
trueqbit: This rule doesn't apply to member functions defined in the class body as they are inlined by default.

dev/select_constraints.h
trueqbit: Both should return a reference. The lower `return select` also returns a reference - after all the function argument is a reference (and that's how `decltype(auto)` works.

dev/util.h
fnc12: I skipped that it is located inside class. Ok then

dev/select_constraints.h
fnc12: I see, ok. I skipped that argument is a ref too

PR #1456, opened 2025-09-26
Restructure classes for table database objects

trueqbit:
Sharing is caring!

With the potential SQL views as yet another schema object type, code needs to be duplicated - or better factored into shared class templates.

#### In this PR:
* Restructured classes representing table database objects:
  * Factored out common functionality into base classes and mixins.
  * Split CTE table class into its own structure, effectively decoupling it from the base table DBO.
  * Moved the facilities for virtual tables into a dedicated header file.
  * Recomposition of the facilities for virtual tables:
    * Split virtual table facilities into: a module tag, virtual table traits that can be specialized, an intermediary description resulting from a `using module` expression and a concrete virtual table class.
* Partially renamed structures by removing the `_t` suffix. Since the advent of alias template for type aliases IMO it isn't the best choice to use it for the library's regular types.
  * `virtual_table_t` -> `virtual_table`.
  * Renamed `table_t` to `base_table` denoting that it is a table that actually stores data. "base table" is the best word in SQL land for this.
  * The renaming should be done for all other DBOs (`index_t`, `trigger_t`) if we agree on this different naming convention. Let me know what you think [yes/no/revert].

#### Additionally:
* Ability to create the FTS5 virtual table using table references.

#### Further changes since the last review on Sep 30 2025:
* Separated generic facilities for virtual tables from facilities for concrete virtual table provided by modules/extensions.
* Explicit object mapping for a virtual table must now be specified when calling `make_virtual_table<O>()` instead of when calling `using_module<O>()`.
* Converted the `dbstat` table into an eponymous virtual table, additionally with the option of creating further instances of it.


dev/schema/table_base.h
```diff
@@ -0,0 +1,246 @@
...
+         */
+#ifdef SQLITE_ORM_DEDUCING_THIS_SUPPORTED
+    template<class O>
+#else
+    template<class O, class Definition>
+#endif
+    struct mapped_object_mixin {
+        using object_type = O;
+
+        /**
```
fnc12, line 211: this formatting is sad =(

fnc12 (approved):
The naming is ok. The PR is good. Thank you @trueqbit

dev/schema/table_base.h
trueqbit: Hmm, what exactly do you mean? Do you want me to change something?

fnc12: I am talking about different indentations on the first two lines. I don't want you to change something until formatter CI tells everything is ok. It is not the first time formatter produces weird output. I think it is something we can't avoid. But of you know how to fix it fast it would be great. If you merge it right now without changing - ok as well.

dev/schema/table_base.h
```diff
@@ -0,0 +1,246 @@
...
+            using excluded_col_index_sequence = col_index_sequence_excluding<elements_type, Trait>;
+            return int(excluded_col_index_sequence::size());
+        }
+
+        /**
+         *  Call passed lambda with all defined columns.
+         *  @param lambda Lambda called for each column. Function signature: `void(auto& column)`
+         */
+        template<class L>
+        void for_each_column(L&& lambda) const {
```
fnc12, line 70: why do we pass an r-value here? if we don't call `std::move(lambda)` on the next line then it goes further as an l-value. In that case the smarter way is just passing an l-value, probably const l-value

dev/schema/virtual_table.h
```diff
@@ -0,0 +1,151 @@
...
+    using is_virtual_table = polyfill::bool_constant<is_virtual_table_v<T>>;
+}
+
+SQLITE_ORM_EXPORT namespace sqlite_orm {
+    /**
+     *  Factory function for a virtual table.
+     *  
+     *  [Deprecation notice] This factory function is deprecated and will be removed in v1.11.
+     */
+    template<class O, class M, class... Cs>
```
fnc12, line 112: can we add deprecate C++ tag here to produce compiler's warnings?

dev/storage.h
```diff
@@ -1188,44 +1187,45 @@ namespace sqlite_orm {
...
+                // eponymous virtual table instances with the same name as their module exist already
+                if constexpr (Table::module_traits_type::is_eponymous::value) {
+                    if (virtualTable.name == Table::module_type::name()) {
+                        return sync_schema_result::already_in_sync;
+                    }
+                }
 
                 const auto res = sync_schema_result::already_in_sync;
-                context_t context{this->db_objects};
+                const serializer_context<db_objects_type> context{this->db_objects};
```
fnc12, line 1200: like for `const`ness here

fnc12 (approved):

dev/schema/virtual_table.h
trueqbit: I already added the deprecation warning to the `using_fts5<O>()` factory function. This should be enough, otherwise you get two warnings...

dev/schema/table_base.h
trueqbit: Well, it's just taking anything, which is usually a lambda in its real sense, i.e. created ad-hoc. In the library's logic there's nNo need to pass it on as an rvalue, so we just pass it on as an lvalue. const-qualifying it is not good as the lambda might be modifyable.

PR #1457, opened 2025-09-29
Support for the R*Tree virtual table

### LINKED ISSUE #279: create virtual table using rtree module
hi,
   i want create a virtual table by using rtree moudle.what should i do?
```
CREATE VIRTUAL TABLE table_demo USING rtree(id, minX, maxX, minY, maxY, +rowid INTEGER);

```

trueqbit:
This PR introduces support for R*Tree virtual tables used in (geo)spatial searching.

Supported features:
* 32-bit floating point RTREE virtual table.
* 32-bit signed integer RTREE_32 virtual table.
* Auxiliary columns

Preconditions:
* The feature is available if `SQLITE_ENABLE_RTREE` is defined. Package and build managers like vcpkg automatically define it.


examples/rtree.cpp
```diff
@@ -0,0 +1,219 @@
...
+        storage.select(&DemoIndex::id,
+                       where(demo_index->*&DemoIndex::maxY >= 35.1 and demo_index->*&DemoIndex::minY <= 35.1));
+    cout << "Zipcodes that overlap with the 35.1th parallel:" << endl;
+    for (int64 zipCode: zipCodesOverlappingWith35thParallel) {
+        cout << '\t' << zipCode << endl;
+    }
+}
+
+// Degrees to radians
+static float d2r(float d) {
```
fnc12: nit: I'd rename this function to `degrees2radians` and remove this comment above. It will reduce amount of lines and make the code more readable in the places of this function calls

tests/schema/virtual_table.cpp
```diff
@@ -22,8 +22,9 @@ TEST_CASE("fts5 virtual table schema") {
 
     auto virtualTable =
         make_virtual_table("posts", using_fts5(make_column("title", &Post::title), make_column("body", &Post::body)));
+
     {
-        const auto compareColumnName = [](const std::string* foundValue, std::string expectedValue) {
+        constexpr auto compareColumnName = [](const std::string* foundValue, std::string expectedValue) {
```
fnc12, line 27: how `constexpr` affects lambda expression here?

fnc12 (approved):
this is amazing. Now `sqlite_orm` can be used in C++ servers which use RTree functionality

tests/schema/virtual_table.cpp
trueqbit: Not at all, only the variable is `constexpr`, just like one of any other type like `constexpr int x`.

examples/rtree.cpp
trueqbit: Haha, yes, forgot about that after a bit of copy-catting :)

PR #1458, opened 2025-10-02
Corrected expressions with table references

trueqbit:
* Added missing expressions with table references: `cross_join<>()`, `natural_join<>()`, `into<>()`.
* Fixed aliasing a table reference.
  Let `user_table` be a table reference of `User`, then `"u"_alias.for_<user_table>()` must be the same as `"u"_alias.for_<User>()`.

### DIFF
--- a/dev/conditions.h
+++ b/dev/conditions.h
@@ -912,10 +912,24 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
         return {};
     }
 
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+    template<orm_refers_to_recordset auto alias>
+    auto cross_join() {
+        return cross_join<internal::auto_decay_table_ref_t<alias>>();
+    }
+#endif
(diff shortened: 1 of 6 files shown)


dev/conditions.h
```diff
@@ -912,11 +912,25 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
         return {};
     }
 
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+    template<orm_refers_to_recordset auto alias>
+    auto cross_join() {
```
fnc12, line 917: no tests or example with it?

fnc12 (approved):

dev/conditions.h
trueqbit: There are tests in _tests/static_tests/column_pointer.cpp_ specifically for expressions that involve table references. Otherwise, there are already the usual tests and examples for cross and natural joins.

PR #1461, opened 2025-10-12
Hidden virtual table columns and table-valued functions

trueqbit:
SQLite features [Hidden columns in virtual tables](https://www.sqlite.org/vtab.html#hiddencol), which then can be used as [Table-valued functions](https://www.sqlite.org/vtab.html#table_valued_functions).

This PR brings:
* Table references for C++17 in order to facilitate table-valued expressions.
* Support for eponymous virtual tables with hidden columns.
  * Replaced `rank()`, `match<Table>()`, `is_equal<Table>()`, `highlight<Table>()` with proper column-oriented logic.
  * `match` can match specific columns as well, not only any column.
* Support for eponymous virtual tables with hidden columns as table-valued functions.
* `get_all<>()` supports an explicit `from` condition, which is useful when selection from table-valued functions.
* Support for the ["generate_series" extension](https://www.sqlite.org/series.html).

Examples in sqlite_orm speak are:
#### dbstat
Aggregated database statistics:
```c++
storage.get_all<dbstat_table>(from(dbstat_table("main", true)));
```
#### generate series
Generate six random numbers:
```c++
storage.select(random(), from(generate_series_table(1, 6)))
```
Where the equivalent by explicitly using hidden columns is:
```c++
storage.select(random(), where(generate_series_table->*&generate_series::hidden::start == 1 and
                               generate_series_table->*&generate_series::hidden::stop == 6);
```
#### fts5
This is of course available for the fts5 module and its hidden `rank` column respectively its hidden eponymous column (named after the fts5 table). In C++ it is represented by a placeholder field that I named `any`:

```c++
struct Post {};
constexpr orm_table_reference auto post_table = c<Post>();

storage.select(highlight(post_table->*&fts5::hidden::any, 0, "<b>", "</b>"),
               where(match(post_table->*&fts5::hidden::any, "SQLite")),
               order_by(post_table->*&fts5::hidden::rank)));
```

A slightly alternative syntax to using column pointers is by explicitly mapping hidden fields into the FTS5 table with an alias template:
```c++
struct Post {
    using hidden = fts5::hidden_fields_of<Post>;
};

storage.select(highlight(Post::hidden::any_field, 0, "<b>", "</b>"),
               where(match(Post::hidden::any_field, "SQLite")),
               order_by(Post::hidden::rank_field)));
```

Note that this approach is simply "syntactic sugar" and up to the programmer to use an alias template. The alias template can have a different name, it can be defined as a nested type name, or it can be defined outside just like a table reference.

For instance (additionally showing that `match()` accepts a regular column):
```c++
struct Post {
    std::string title;
};
using post_hidden = fts5::hidden_fields_of<Post>;

storage.select(highlight(post_hidden::any_field, 0, "<b>", "</b>"),
               where(match(&Post::title, "SQLite")),
               order_by(post_hidden::rank_field)));
```

### DIFF
--- a/dev/serializing_util.h
+++ b/dev/serializing_util.h
@@ -107,6 +107,7 @@ namespace sqlite_orm {
             conditions_tuple,
             actions_tuple,
             expressions_tuple,
+            filtered_expressions_tuple,
             dynamic_expressions,
             compound_expressions,
             serialized,
@@ -137,6 +138,7 @@ namespace sqlite_orm {
         constexpr streaming<stream_as::conditions_tuple> streaming_conditions_tuple{};
         constexpr streaming<stream_as::actions_tuple> streaming_actions_tuple{};
         constexpr streaming<stream_as::expressions_tuple> streaming_expressions_tuple{};
+        constexpr streaming<stream_as::filtered_expressions_tuple> streaming_filtered_expressions_tuple{};
         constexpr streaming<stream_as::dynamic_expressions> streaming_dynamic_expressions{};
         constexpr streaming<stream_as::compound_expressions> streaming_compound_expressions{};
         constexpr streaming<stream_as::serialized> streaming_serialized{};
@@ -187,6 +189,22 @@ namespace sqlite_orm {
             auto& context = std::get<2>(tpl);
 
             iterate_tuple(args, [&ss, &context, first = true](auto& arg) mutable {
+                static constexpr std::array<orm_gsl::czstring, 2> sep = {", ", ""};
+                ss << sep[std::exchange(first, false)] << serialize(arg, context);
+            });
+            return ss;
+        }
+
+        // serialize and stream a tuple of expressions including only those specified by an index sequence;
+        // comma-separated
+        template<class T, class Seq, class Ctx>
+        std::ostream& operator<<(std::ostream& ss,
+                                 std::tuple<const streaming<stream_as::filtered_expressions_tuple>&, Seq, T, Ctx> tpl) {
+            const auto& args = std::get<1>(tpl);
+            const auto& included_index_sequence = std::get<2>(tpl);
+            auto& context = std::get<3>(tpl);
+
+            iterate_tuple(args, included_index_sequence, [&ss, &context, first = true](auto& arg) mutable {
                 static constexpr std::array<orm_gsl::czstring, 2> sep = {", ", ""};
                 ss << sep[std::exchange(first, false)] << serialize(arg, context);
             });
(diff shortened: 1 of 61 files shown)


dev/serializing_util.h
```diff
@@ -416,7 +434,7 @@ namespace sqlite_orm {
             return ss;
         }
 
-        // serialize and stream a tuple of column constraints;
+        // serialize and stream a tuple of column constraints exluding auxiliary columns;
```
fnc12: *ex[c]luding

dev/statement_serializer.h
```diff
@@ -276,10 +288,9 @@ namespace sqlite_orm {
                                                             const Ctx& context) SQLITE_ORM_OR_CONST_CALLOP {
                 static_assert(is_bindable_v<type_t<statement_type>>, "A literal value must be also bindable");
 
-                Ctx literalCtx = context;
+                auto literalCtx = context;
                 literalCtx.replace_bindable_with_question = false;
-                statement_serializer<type_t<statement_type>> serializer{};
-                return serializer(literal.value, literalCtx);
```
fnc12, line 282: it was a typo?

dev/storage_base.h
```diff
@@ -724,7 +724,7 @@ namespace sqlite_orm {
                     *other.connection,
                     std::bind(&storage_base::on_open_internal, this, std::placeholders::_1))),
                 cachedForeignKeysCount(other.cachedForeignKeysCount),
-                executor{std::move(other.executor.will_run_query), std::move(other.executor.did_run_query)} {
+                executor{other.executor.will_run_query, other.executor.did_run_query} {
```
fnc12, line 727: looks like a bug left after my PR. Thanks for fixing it

tests/schema/virtual_table.cpp
```diff
@@ -149,34 +187,132 @@ TEST_CASE("dbstat virtual table schema") {
...
+                REQUIRE_THAT(rows, UnorderedEquals(std::vector<std::string>{"c15000", "c15100"}));
+            }
         }
     }
 }
 #endif
 
 #ifdef SQLITE_ENABLE_RTREE
 TEST_CASE("rtree virtual table schema") {
+    constexpr auto compareColumnName = [](const std::string* foundValue, std::string expectedValue) {
```
fnc12: nit: one can put this outside of test case in the file scope and reuse it in all test cases

fnc12 (approved):

dev/statement_serializer.h
trueqbit: No, in a7414bd849725789d8b844540089895c9354218b I probably thought to somehow eliminate indirect calls because I wanted to stream a literal value. But now it seems to that it should simply go through the `serialize()` function just like in all other places.

dev/serializing_util.h
trueqbit: WTF.

dev/storage_base.h
trueqbit: Sometimes my compiler spits out useful warning messages :)

PR #1468, opened 2025-11-27
Correcting ordinary insert statement for tables with composite primary keys

### LINKED ISSUE #1467: False NOT NULL constraint failure for table with true composite primary key
Background. This is a minimized version of a key logging database that collects various datasets of keystrokes from published researches on human-computer interaction.  For every i-th keystroke of a session, the key log records key code, time of press, t_i, and the k-th order inter-key interval, defined as IKI_{k}(i) = t_{i} - t_{i - k} for i >= k and zero (or any non-NULL value) otherwise.  As a dataset may be very large, the keystroke table should use (session_id,press_time) as the 'true' primary key, i.e. without the hidden ROWID column.

In this minimal test case, the keystroke record should be inserted successfully into the table. 

It isn't.

Database structure and content after the false failure
```
SQLite version 3.42.0 2023-05-16 12:36:15
Enter ".help" for usage hints.
sqlite> .tables
keystrokes  sessions
sqlite> .schema
CREATE TABLE IF NOT EXISTS "keystrokes" ("session_id" INTEGER NOT NULL, "press_t
ime" INTEGER NOT NULL, "release_time" INTEGER NOT NULL, "iki1" INTEGER NOT NULL,
 FOREIGN KEY("session_id") REFERENCES "sessions"("id"), PRIMARY KEY("session_id"
, "press_time")) WITHOUT ROWID;
CREATE TABLE IF NOT EXISTS "sessions" ("id" INTEGER PRIMARY KEY NOT NULL, "ident
ifier" TEXT NOT NULL);
sqlite> SELECT * FROM sessions;
1|test_session
sqlite> SELECT * FROM keystrokes;
sqlite> .quit
```
Source code
```C++
// SQLITE_ORM test
// issue 1: semantics of UNIQUE vs composite PRIMARY KEY without ROWID.
// I want this
//
//    PRIMARY KEY(session_id, press_time) WITHOUT ROWID
//
// SQLITE_ORM seems to fail to do this.

// minimal import

#include <iostream>

struct MinimalSession {
    int id;
    std::string identifier;
};

struct MinimalKeystroke {
    int session_id;
    int press_time;
    int iki1;
};

auto initMinimalStorage() {
    using namespace sqlite_orm;
    return make_storage("test_minimal.db",
        make_table("sessions",
            make_column("id", &MinimalSession::id, primary_key()),
            make_column("identifier", &MinimalSession::identifier)
        ),
        make_table("keystrokes",
            make_column("session_id", &MinimalKeystroke::session_id),
            make_column("press_time", &MinimalKeystroke::press_time),
            make_column("iki1", &MinimalKeystroke::iki1),
            foreign_key(&MinimalKeystroke::session_id).references(&MinimalSession::id),
            primary_key(&MinimalKeystroke::session_id, &MinimalKeystroke::press_time)
        ).without_rowid()
    );
}

void test_minimal_import() {
    auto storage = initMinimalStorage();
    storage.sync_schema();
    storage.remove_all<MinimalKeystroke>();
    storage.remove_all<MinimalSession>();

    // Create one session
    MinimalSession session{-1, "test_session"};
    int session_id = storage.insert(session);
    std::cout << "Created session with id: " << session_id << std::endl;

    // Insert one keystroke
    MinimalKeystroke keystroke{session_id, 100, 50};
    storage.insert(keystroke);

    std::cout << "Minimal test PASSED" << std::endl;
}

int main() {
    std::cout << "=== SQLite ORM Bug Investigation ===" << std::endl;
    std::cout << "Testing PRIMARY KEY + WITHOUT ROWID behavior" << std::endl;

    try {
        test_minimal_import();
        std::cout << "SUCCESS: Test completed without exceptions" << std::endl;
        return 0;
    }
    catch (const std::exception& e) {
        std::cout << "\n!!! DIAGNOSIS REPORT !!!" << std::endl;
        std::cout << "EXCEPTION: " << e.what() << std::endl;
        std::cout << "\nROOT CAUSE ANALYSIS:" << std::endl;
        std::cout << "1. Schema: PRIMARY KEY(session_id, press_time) + WITHOUT ROWID" << std::endl;
        std::cout << "2. Session was successfully inserted with valid ID" << std::endl;
        std::cout << "3. Keystroke object created with valid session_id" << std::endl;
        std::cout << "4. BUT: NOT NULL constraint failed on keystrokes.session_id" << std::endl;
        std::cout << "\nCONCLUSION: sqlite_orm bug in value binding for composite PK + WITHOUT ROWID" << std::endl;
        std::cout << "The library is not properly passing the session_id value to SQLite." << std::endl;
        return 1;
    }
    catch (...) {
        std::cout << "\n!!! UNKNOWN EXCEPTION !!!" << std::endl;
        std::cout << "Unexpected exception type caught" << std::endl;
        return 2;
    }
}
```

trueqbit:
### Improvements:
* Objects mapped to tables with composite primary keys are now insertable with an ordinary insert statement: they get inserted with the values of the specified object except for default values.
  * This improvement contains a bug fix for insert statements into tables without rowid where providing values for primary key columns is required (and expected).
  
### Other improvements:
* Validate base table definition when calling `make_table()`.
  * Specifically this ensures that only one primary key definition is specified.
  * It also means that we can get rid of a runtime error when serializing a `get` query statement.
* Validate column definition with respect to auto-increment specifier on a column primary key and the field type.
* Improved documentation as a comment on the ordinary 'insert' statement into tables, describing how the routine behaves with rowid vs. without rowid tables and with a single-column primary key vs. composite keys.
* Many more unit tests of course.

### Fixes
* Corrected some SQLite version test that enable version-dependent SQLite features.


dev/schema/column.h
```diff
@@ -84,6 +85,14 @@ namespace sqlite_orm {
                 return tuple_has<constraints_type, Trait>::value;
             }
 
+            /**
+             *  Checks whether contraints contain specified class template.
```
fnc12: typo: con[s]traints

dev/schema/column.h
```diff
@@ -165,19 +174,65 @@ namespace sqlite_orm {
             field_type_t,
             filter_tuple_sequence_t<Elements, mpl::disjunction_fn<is_column, is_hidden_column>::template fn>>;
 
-#if SQLITE_VERSION_NUMBER >= 3031000
+        // Custom type: programmer's responsibility to garantee data integrity in the value range of a 64-bit signed integer
```
fnc12: typo: g[u]arantee

dev/schema/table.h
```diff
@@ -117,6 +115,25 @@ namespace sqlite_orm {
...
+            using pk_index_sequence = filter_tuple_sequence_t<elements_type, is_primary_key>;
+            using pkcol_index_sequence = col_index_sequence_with<elements_type, is_primary_key>;
+
+            static_assert(pk_index_sequence::size() + pkcol_index_sequence::size() <= 1,
+                          "A base table can only have 1 primary key definition");
+            if constexpr (pk_index_sequence::size() > 0) {
+                constexpr size_t nTablePrimaryKeyColumns =
+                    nested_tuple_size_for_t<columns_tuple_t, elements_type, pk_index_sequence>::value;
+
+                static_assert(nTablePrimaryKeyColumns > 0, "Tabel primary key definition must contain one column");
```
fnc12: typo: `Table` not `Tabel`

dev/storage.h
```diff
@@ -928,24 +939,33 @@ namespace sqlite_orm {
             }
 
             template<class O, class... Cols>
-            int insert(const O& o, columns_t<Cols...> cols) {
+            int64 insert(const O& o, columns_t<Cols...> cols) {
```
fnc12: on man, it will produce a lot of warnings. I'd leave it like this just like it is done in Java (where almost all integer operations are handles by simple `int`s) for simplicity and if somebody wants to get more precise `int64` result one can use `last_insert_rowid()` instead or use a prepared statement of `insert`. I mean this function `int insert` is a real legacy and I'd leave its signature like this just like the below one's

examples/except_intersection.cpp
```diff
@@ -9,12 +9,12 @@ using std::cout;
 using std::endl;
 
 struct DeptMaster {
-    int deptId = 0;
+    sqlite_orm::int64 deptId = 0;
```
fnc12, line 12: what happens to this example if we skip this change?

tests/prepared_statement_tests/replace_range.cpp
```diff
@@ -75,11 +75,19 @@ TEST_CASE("Prepared replace range") {
         SECTION("pointers") {
             userPointers.push_back(std::make_unique<User>(user));
             auto statement = storage.prepare(
-                replace_range<User>(userPointers.begin(), userPointers.end(), &std::unique_ptr<User>::operator*));
+                replace_range(userPointers.begin(), userPointers.end(), &std::unique_ptr<User>::operator*));
             REQUIRE(get<0>(statement) == userPointers.begin());
             REQUIRE(get<1>(statement) == userPointers.end());
             storage.execute(statement);
         }
+        SECTION("references") {
```
fnc12, line 83: this type of sections is a really good shot

tests/tests5.cpp
```diff
@@ -315,9 +315,9 @@ TEST_CASE("issue822") {
                                            make_column("value", &A::getValue, &A::setValue),
                                            primary_key(&A::getAddress, &A::getType, &A::getIndex)));
     storage.sync_schema();
-    storage.replace(A(1, 1, 0, std::make_shared<double>(55.5)));
+    storage.insert(A(1, 1, 0, std::make_shared<double>(55.5)));
```
fnc12, line 318: this is a really good bug fix

examples/except_intersection.cpp
trueqbit: Does it hurt if we keep educating programmers to use `int64`? On my platforms int is 32-bit.
But I can revert, and would need to do so for all examples.

fnc12: no I mean I really want you to answer the question so I can ensure I understood your changes right. It is totally ok to educate programmers to do right things.

trueqbit: Oh ok. The reason I changed it is to use the datatype that SQLite uses and assumes for the [rowid](https://sqlite.org/autoinc.html).

fnc12: what happens to this example if we skip this change? will it compile? will the warning appear?

trueqbit: The example will compile with both types - `int` and `int64` - without warnings. As discussed there will be no deprecation warning for integral types other than `int64`, however `int64` is still the most accurate type for the rowid alias.

fnc12: got it, thank you. Now I understand it quite correct =)

dev/storage.h
trueqbit: OK!

dev/storage.h
```diff
@@ -244,17 +254,16 @@ namespace sqlite_orm {
...
-                              "of 'insert', or you can use 'insert' with explicit column listing.");
+                if constexpr (pkcol_index_sequence::size()) {
+                    constexpr auto pkcol_idx = index_sequence_value_at<0>(pkcol_index_sequence{});
+                    using pkcol_type = std::tuple_element_t<pkcol_idx, elements_type>;
+                    static_assert(
+                        mpl::invoke_t<check_if<is_pkcol_implicitly_insertable>, pkcol_type>::value,
+                        "While SQLite allows primary keys of any type, sqlite_orm restricts an ordinary 'insert' into "
+                        "tables with single-column primary keys to those with an implicitly insertable column because "
+                        "it is the 'rowid' alias or has a default value."
+                        "Instead, please use 'replace' or 'insert' with explicitly specified columns.");
```
fnc12: nit: I'd replace 
```
'insert' with explicitly specified columns.
```
with
```
insert(object, columns(...));
```
just to make it more recognizable cause people usually understand examples more likely than definitions

dev/storage.h
```diff
@@ -927,6 +939,14 @@ namespace sqlite_orm {
                 this->execute(statement);
             }
 
+            /**
+             *  Insert routine with explicitly specified columns.
+             *  
+             *  @return The ID of the last inserted record for a rowid table, otherwise a meaningless value.
+             *          Attention: `sqlite3_last_insert_rowid()` is used to retrieve the last inserted ID, therefore the ID is only useful in single-threaded contexts.
+             *          Attention: While SQLite returns a 64-bit integer as rowid, this function returns an `int` that most likely has less precision.
+             *                     If you need the full 64-bit rowid, use `storage_t<>::execute()` instead.
```
fnc12: "...or use `last_insert_rowid()` function call manually"
(and the same to the comment below)

dev/storage.h
```diff
@@ -1530,17 +1564,25 @@ namespace sqlite_orm {
...
-                                         mpl::disjunction_fn<is_primary_key, is_generated_always>>>(
-                        call_as_template_base<column_field>([&table, &bindValue, &object](auto& column) {
-                            if (!exists_in_composite_primary_key(table, column)) {
-                                bindValue(polyfill::invoke(column.member_pointer, object));
+                                      bindValue = field_value_binder{stmt}](const object_type& object) mutable {
+                    using table_type = polyfill::remove_cvref_t<decltype(table)>;
+                    using without_rowid = typename table_type::is_without_rowid;
+
+                    table.template for_each_column_excluding<  ///
+                        mpl::disjunction<
```
fnc12: nit: probably this could be more readable with local aliases like
```c++
using mpl_or = mpl::disjunction;
using mpl_and = mpl::conjunction;
using mpl_not = mpl::not_;
```

fnc12 (approved):

dev/storage.h
trueqbit: Well, if this is a serious wish, then we should flatten the namespaces and just define everything in terms of `mpl_` prefixes...

fnc12 (approved):

dev/storage.h
fnc12: it is not so serious wish, it is just a nit, cause you're more familiar with mpl part of the lib

PR #1469, opened 2025-12-09
Yet another round of C++17 hygiene

trueqbit:
Yet another PR with a bit of C++17 hygiene and a few improvements.

### C++17:
* Collapsed nested namespaces, meaning: `namespace sqlite_orm { namespace internal {} }` -> `namespace sqlite_orm::internal {}`. This reduces code intendation, and makes it more obvious (to collaborators) what's part of the public namespace.
**Note**: Because this changes a lot of white spacing it is best to **hide whitespace** for [diff comparison](https://github.com/fnc12/sqlite_orm/pull/1469/files?w=1#diff-6ec8fc411a3160c70a21f477044792080726f74a0bc7d62ddfa30a6388c69bf5).
* Avoided using standard type trait variable templates: As long as we don't require a C++17 Standard Library base line we cannot assume that variable templates are available.

### Improvements:
* Qualified calls to `std::get()`. This function is really widely overloaded, so I consider it good practice in template code to qualify it if called with a tuple.
* Simplified a trait testing for whether an expression is a binary operator.

### Fixes:
* Exported `limit()` that was missed in PR #1336.


dev/implementations/column_definitions.h
```diff
@@ -12,20 +12,16 @@
 #include "../default_value_extractor.h"
 #include "../schema/column.h"
 
-namespace sqlite_orm {
-    namespace internal {
+namespace sqlite_orm::internal {
+    template<class... Op>
+    std::unique_ptr<std::string> column_constraints<Op...>::default_value() const {
```
fnc12, line 17: nit: later this can be translated to `std::optional` instead of `std::unique`. I used `std::unique` initially cause C++14 doesn't have `std::optional`

dev/select_constraints.h
```diff
@@ -25,433 +25,429 @@
...
+    /**
+     *  Monikered (aliased) CTE expression.
+     */
+    template<class Moniker, class ExplicitCols, class Hints, class Select>
+    struct common_table_expression {
+        using cte_moniker_type = Moniker;
+        using expression_type = Select;
+        using explicit_colrefs_tuple = ExplicitCols;
+        using hints_tuple = Hints;
+        static constexpr size_t explicit_colref_count = std::tuple_size<ExplicitCols>::value;
```
fnc12, line 255: why do we come back from `*_v` to `*::value` approach here?

fnc12 (approved):

dev/select_constraints.h
trueqbit: Like I wrote in the PR description: "As long as we don't require a C++17 Standard Library base line we cannot assume that variable templates are available."

fnc12: ah I see. Pardon, thanks

dev/implementations/column_definitions.h
trueqbit: I absolutely agree, however as I wrote in the PR description: "As long as we don't require a C++17 Standard Library base line we cannot assume that variable templates are available.". This is also true for all other C++17 Library facilities.

PR #1470, opened 2025-12-09
Claude Code markdown

trueqbit:
I am successfully using Claude when working with the library, and Claude produced a rather good CLAUDE markdown file that provides guidance to Claude Code when working with code in this repository.

I thought I add it to the repo.

As of 2026-08-15 this PR also contains:
- Updated documentation for contributors with respect to the internal vocabulary layer.
- An `AGENTS.md` file.
- After our most recent work I already tackled the task to split and move the former `dev/storage_lookup.h` under `dev/schema/db_objects.h` respectively `dev/schema/algorithms/table_lookup.h`.


third_party/fix_comment_alignment.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
```
fnc12, line 1: why do we need this file?

fnc12 (approved):

third_party/fix_comment_alignment.py
trueqbit: We usually don't need it. I used it to reformat the comments after collapsing the namespaces, because clang-format didn't do it. I just thought I add it because it required a few iterations to instruct Claude to produce something that works. If you think it is unnecessary I can remove it, so we don't keep dead freight :)

fnc12:
@trueqbit let's fix the conflicts and merge it

third_party/fix_comment_alignment.py
trueqbit: @fnc12 yes/no?

fnc12: yes

fnc12 (approved):

PR #1471, opened 2026-01-10
changed collate a bit

fnc12:
Just going on with refactoring of the code order especially with AST nodes


dev/statement_serializer.h
```diff
@@ -925,12 +925,12 @@ namespace sqlite_orm::internal {
...
 
         template<class Ctx>
-        SQLITE_ORM_STATIC_CALLOP std::string operator()(const statement_type& c,
+        SQLITE_ORM_STATIC_CALLOP std::string operator()(const statement_type& statement,
                                                         const Ctx& context) SQLITE_ORM_OR_CONST_CALLOP {
             auto newContext = context;
             newContext.use_parentheses = false;
-            auto res = serialize(c.expr, newContext);
-            return res + " " + static_cast<std::string>(c);
+            auto res = serialize(statement.expr, newContext);
```
trueqbit: It is preferable not to use an intermediate result variable, as this prevents move-on-return optimization.

dev/statement_serializer.h
```diff
@@ -911,12 +911,12 @@ namespace sqlite_orm::internal {
...
 
         template<class Ctx>
-        SQLITE_ORM_STATIC_CALLOP std::string operator()(const statement_type& c,
+        SQLITE_ORM_STATIC_CALLOP std::string operator()(const statement_type& statement,
                                                         const Ctx& context) SQLITE_ORM_OR_CONST_CALLOP {
             auto newContext = context;
             newContext.use_parentheses = false;
-            auto res = serialize(c.expr, newContext);
-            return res + " " + static_cast<std::string>(c);
+            auto res = serialize(statement.expression, newContext);
```
trueqbit: It is preferable not to use an intermediate result variable, as this prevents move-on-return optimization.

trueqbit (approved):

dev/statement_serializer.h
fnc12: fixed, thanks

dev/statement_serializer.h
fnc12: fixed, thanks

PR #1480, opened 2026-03-19
added triggers sync

### LINKED ISSUE #1429: sync_schema does not sync triggers

fnc12:
sync_schema() now detects changed triggers and recreates them (DROP + CREATE). Previously triggers were created once with CREATE TRIGGER IF NOT EXISTS and never updated.

Fixes #1429

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,12 +53,12 @@ jobs:
             cxx_standard: "-DSQLITE_ORM_ENABLE_CXX_20=ON"
             triplet: x86-windows
 
-          - name: "VS 2019, x64, C++17"
-            os: windows-2019
-            platform: x64
-            arch: x64
-            cxx_standard: "-DSQLITE_ORM_ENABLE_CXX_17=ON"
-            triplet: x64-windows
+          # - name: "VS 2019, x64, C++17"
+          #   os: windows-2019
+          #   platform: x64
+          #   arch: x64
+          #   cxx_standard: "-DSQLITE_ORM_ENABLE_CXX_17=ON"
+          #   triplet: x64-windows
 
     name: Windows - ${{ matrix.name }}
 
--- a/.cursor/rules/project-structure.mdc
+++ b/.cursor/rules/project-structure.mdc
@@ -1,0 +1,11 @@
+---
+description: Project structure rules for sqlite_orm library development
+alwaysApply: true
+---
+
+# Project Structure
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -53,12 +53,12 @@ jobs:
             cxx_standard: "-DSQLITE_ORM_ENABLE_CXX_20=ON"
             triplet: x86-windows
 
-          - name: "VS 2019, x64, C++17"
-            os: windows-2019
-            platform: x64
-            arch: x64
-            cxx_standard: "-DSQLITE_ORM_ENABLE_CXX_17=ON"
-            triplet: x64-windows
+          # - name: "VS 2019, x64, C++17"
```
fnc12, line 56: commenting out cause rn it doesn't even start on gh action IDK why

Issue #1438, opened 2025-05-27
Is there support for sqlite *view* in sqlite_orm?


juandent:
I have been away for a while due to health issues and was wondering if support for sqlite views has been added to sqlite_orm since it was a desired feature since a long time ago...

Did you get around implementing it?

fnc12:
duplicate https://github.com/fnc12/sqlite_orm/issues/306

PR #1481, opened 2026-04-26
SQL Query Views

### LINKED ISSUE #306: Using views
Any suggestions on how to use views from within sqlite_orm?  Can the sync_schema functionality create views and can they be accessed the same way as tables?

trueqbit:
Views in SQL are virtual tables based on the result-set of a SELECT statement, also known as query views.

With sqlite_orm, query views are defined using make_view() which leverages C++ reflection to automatically map columns from the SELECT statement to struct fields and to specify the query view's name with a class-scope annotation.

SQL views created with make_view() don't require explicit column definitions - the columns are automatically determined from the view's object type through reflection.

```c++
struct[[= "high_earners"_orm_name]] HighEarner {
    int64 id;
    std::string name;
    double salary;
};

make_storage("",
             make_table("employee", /*...*/),
             make_view<HighEarner>(select(columns(&Employee::id, &Employee::name, &Employee::salary),
                                          where(c(&Employee::salary) > 0.0))));
```

Resolves the following discussions and tickets:
- https://github.com/fnc12/sqlite_orm/discussions/968
- https://github.com/fnc12/sqlite_orm/discussions/1437
- https://github.com/fnc12/sqlite_orm/issues/306
- https://github.com/fnc12/sqlite_orm/issues/1438


dev/conditions.h
```diff
@@ -1016,6 +1017,14 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
         return {std::move(o)};
     }
 
+    /** 
+     *  [Deprecation notice] This expression factory function is deprecated and will be removed in v1.11.
```
fnc12, line 1020: no need to deprecate it cause it is also used in window funcs and it is not going to be deprecated there. So I'd remove `[Deprecation notice]`

dev/storage_base.h
```diff
@@ -214,6 +234,36 @@ namespace sqlite_orm::internal {
             return result;
         }
 
+        /**
+         *  Directly checks the actual database whether the specified view exists, bypassing the library's 'storage' mapping.
+         *  @return true if view with the specified name exists in the database, false otherwise.
+         */
+        bool view_exists(const std::string& tableName) {
```
fnc12: `viewName` instead of `tableName` I guess

dev/storage_base.h
```diff
@@ -214,6 +234,36 @@ namespace sqlite_orm::internal {
...
+            auto connection = this->get_connection();
+            return this->view_exists(connection.get(), tableName);
+        }
+
+        bool view_exists(sqlite3* db, const std::string& tableName) const {
+            bool result = false;
+            std::string sql;
+            {
+                std::stringstream ss;
+                ss << "SELECT COUNT(*) FROM sqlite_master WHERE type = " << quote_string_literal("view")
```
fnc12: can't we reuse existing log of `table_exists` here instead of  constructing this SQL from scratch?

dev/storage_base.h
```diff
@@ -214,6 +234,36 @@ namespace sqlite_orm::internal {
...
+                   << " AND name = " << quote_string_literal(tableName) << std::flush;
+                sql = ss.str();
+            }
+            this->executor.perform_exec(
+                db,
+                sql,
+                [](void* userData, int /*argc*/, orm_gsl::zstring* argv, orm_gsl::zstring* /*azColName*/) -> int {
```
fnc12: this callback is also duplicated from `table_exists`. Maybe we can move it to a dedicated free func and bind it where we need it

dev/storage_base.h
```diff
@@ -806,7 +865,7 @@ namespace sqlite_orm::internal {
             auto connection = this->get_connection();
             data_t objectNames;
             std::stringstream ss;
-            ss << "SELECT name FROM sqlite_master WHERE type=" << quote_string_literal(std::string(type));
+            ss << "SELECT name FROM sqlite_master WHERE type=" << quote_string_literal(std::string(type)) << std::flush;
```
fnc12, line 836: nit: we can make it more thiner in memory consumption: declare `std::stringstream` inside sub-scope/lambda which returns string and pass string while `std::stringstream` is already destroyed

dev/storage_base.h
```diff
@@ -1048,6 +1107,20 @@ namespace sqlite_orm::internal {
             this->executor.perform_void_exec(db, sql.c_str());
         }
 
+        void drop_view_internal(sqlite3* db, const std::string& viewName, bool ifExists) {
```
fnc12: probably we can pass `std::string_view` as `viewName` instead of const ref of `std::string` cause we already dropped C++11 support

examples/view.cpp
```diff
@@ -0,0 +1,210 @@
...
+    int64 id;
+    std::string name;
+    std::string location;
+};
+
+// View objects - note how we only define the struct fields, no column mappings needed!
+// The fields are automatically mapped through C++ reflection
+
+// View 1: High earners (employees earning more than 60000)
+struct[[= dbo_name("high_earners")]] HighEarner {
```
fnc12: this looks like a dedicated feature (putting name as C++ annotation right inside type and it can work for both tables and view

examples/view.cpp
```diff
@@ -0,0 +1,210 @@
...
+                   make_column("location", &Department::location)),
+
+        // Define views - notice how we only specify the SELECT statement.
+        // The column mappings and view name are derived from the view object type
+        // (column names from non-static data members; view name from the optional
+        // `[[=dbo_name("…")]]` annotation, falling back to the type's identifier).
+
+        // View 1: Filter high earners
+        make_view<HighEarner>(
+            select(columns(&Employee::id, &Employee::name, &Employee::salary), where(c(&Employee::salary) > 60000.0))),
```
fnc12, line 91: looks like the user doesn't have an opportunity to specify explicit column name of the view right?

examples/view.cpp
```diff
@@ -0,0 +1,210 @@
...
+        cout << "=== View Metadata ===" << endl;
+        auto viewNames = storage.view_names();
+        cout << "Views in database:" << endl;
+        for (const auto& viewName: viewNames) {
+            cout << "  - " << viewName << endl;
+        }
+        cout << endl;
+
+        cout << "Example completed successfully!" << endl;
+    } catch (const std::system_error& e) {
```
fnc12, line 204: this feature is definitely GOAT

tests/schema/view_tests.cpp
```diff
@@ -0,0 +1,66 @@
...
+TEST_CASE("view::find_column_name") {
+    struct User {
+        int64 id = 0;
+        std::string name;
+    };
+
+    SECTION("direct") {
+        auto view = make_view<UserViewSchemaTests>(select(asterisk<User>()));
+
+        SECTION("fields") {
```
fnc12: looks like an excess section here and below

dev/storage_base.h
trueqbit: I don't think I'll address this in this PR ;)

dev/conditions.h
trueqbit: `order_by(rank())` can only be used with FTS not with Window functions, that's why I added it here (actually moved from the `rank()` factory).

examples/view.cpp
trueqbit: This PR is only about query views. Upcoming updates to tables are in the pipeline, see branch [`dbo-reflection`](https://github.com/fnc12/sqlite_orm/compare/experimental/sql-view...dbo-reflection).

examples/view.cpp
trueqbit: Not yet. We can add this in the future with C++ annotations.

dev/storage_base.h
trueqbit: Yes, but then the function needs an additional `std::string` variable, whose size is an additional 40 bytes (with msvc).

tests/schema/view_tests.cpp
trueqbit: I probably additional things in mind 2 years ago :)

dev/conditions.h
fnc12: ah ok got it. Sorry my oversight. Anyway I'd rename `o` var to something more readable like `rank` or `expression`

dev/storage_base.h
fnc12: why is it a problem?

dev/storage_base.h
```diff
@@ -294,6 +310,15 @@ namespace sqlite_orm::internal {
         }
 #endif
 
+        /**
+         *  Returns the names of existing permanent view in the database. Doesn't check storage itself - works only with
+         *  actual database.
+         *  @return Returns list of tables in database.
+         */
+        std::vector<std::string> view_names() {
```
fnc12, line 320: idea: we can add API for this without heap allocation which will return iteratable object over prepared statement so one can iterate over view names the same way. What do you think? No need to add it into this PR though

fnc12 (approved):

dev/storage_base.h
trueqbit: The question is what programmers use these APIs for... However it's a good idea to return a C++ view range.

dev/storage_base.h
trueqbit: I was thinking of function size (in terms of stack size the function requires).
TBH overall I am not convinced why assigning to a string variable from a scope in the same function is an improvement over simply accessing the stream's string. A better solution would be to just reuse the statement API itself - like you implicitly suggested on `view_names()` for iterating over views/tables/triggers.

examples/view.cpp
```diff
@@ -0,0 +1,210 @@
...
+    int64 id;
+    std::string name;
+    std::string location;
+};
+
+// View objects - note how we only define the struct fields, no column mappings needed!
+// The fields are automatically mapped through C++ reflection
+
+// View 1: High earners (employees earning more than 60000)
+struct[[= "high_earners"_dbo_name]] HighEarner {
```
fnc12: do we really want to expose `_dbo_name` as public API? I mean that `dbo` is pretty clear as is no doubt but the case is that it is used only as an internal word. The user who downloaded `sqlite_orm` to glue his/her code with SQLite doesn't operate with this word in mind. He/She uses 'table', 'view' etc instead, not 'dbo'. So maybe we will change `_dbo_name` to `_sqlite_orm_name` or even `_sqlite_name`? it looks like more `@Sqlite` tags in modern languages like Java and Python

fnc12 (approved):

examples/view.cpp
trueqbit: Hmm, for me "dbo" is pretty clear, however I agree that `dbo_name` looses meaning compared to the specific `table_name`.

Another advantage of this change would be that columns can also be renamed using the same API (in a later PR)!

We already have public-facing C++ concepts that are prefixed with "orm", e.g. `orm_table_reference`, etc... I would prefer `_orm_name` and `orm_name()` over `_sqlite_orm_name` or `_sqlite_name`...

fnc12: sounds good until we make the same lib for let's say `postrgres_sql` with the same API =) But I am ok with `_orm_name` and `orm_name()` cause it sounds pretty reasonable

fnc12 (approved):

fnc12:
@trueqbit thank you!

PR #1482, opened 2026-04-27
added cpp23

fnc12:


dev/statement_serializer.h
```diff
@@ -1797,7 +1797,7 @@ namespace sqlite_orm::internal {
                                                         const Ctx&) SQLITE_ORM_OR_CONST_CALLOP {
             std::stringstream ss;
             ss << "SET ";
-#ifdef SQLITE_ORM_CPP20_RANGES_SUPPORTED
+#if defined(SQLITE_ORM_CPP20_RANGES_SUPPORTED) && (!defined(__clang__) || (__clang_major__ >= 15))
```
fnc12: @trueqbit probably you'll advice to rename it to `SQLITE_ORM_*` macros instead

tests/static_tests/function_static_tests.cpp
```diff
@@ -24,6 +24,11 @@ concept storage_aggregate_callable = requires(S& storage) {
     { storage.template create_aggregate_function<f>() };
     { storage.template delete_aggregate_function<f>() };
 };
+
+constexpr const int& clamp_int_ref(const int& v, const int& lo, const int& hi) {
```
fnc12: fallback to old days compiler

trueqbit: This seems to unused in the meantime...

Do you have a reproducible example that we can try on godbolt? Because a simpler fallback could be taking the address of the instantiated template function: `&std::clamp<int>` instead of skipping the tests
Or if a global function is required just create a function pointer variable explicitly: `constexpr auto clamp_int = &std::clamp<int>`.

dev/statement_serializer.h
trueqbit: Yes, let's encapsulate this.

It seems that Clang-14/libstdc++ 11 don't fully support C++20 range view adapters yet... so let's introduce a feature test macro in *dev/functional/config.h* below `SQLITE_ORM_CPP20_RANGES_SUPPORTED`:

```c++
#if __cpp_lib_ranges >= 201911L
#define SQLITE_ORM_CPP20_RANGES_SUPPORTED
#endif

#if __cpp_lib_ranges >= 202110L
#define SQLITE_ORM_CPP20_VIEWS_SUPPORTED
#endif
```

For our own sanity, temporary containers should be usable for views (`owning_view)` and that the `range`, `viewable_range` and `view` concepts should properly interact with `views::all`.

Unfortunately Clang-15 seems to still have a bug when evaluating concepts - Clang-15/libstdc++ 12 don't work but Clang-16/libstdc++ 12 does, see this code snippet on [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIABwA7KSuADJ4DJgAcj4ARpjEEgCspAAOqAqETgwe3r4BwemZjgLhkTEs8YlcKXaYDtlCBEzEBLk%2BfkG2mPYlDE0tBGXRcQnJts2t7fldCpNDESOVYzUAlLaoXsTI7BzmAMwRyN5YANQm%2B27EhsCYChfYJhoAggdHJ5jnl2LAJIQILAeT1eZkODGOXjOFzcADd6kRiECXsCIgRTiwmBEIKtzoErC9ToTTgB6YmnAh3NEUljpa7EACepzQgkxkWI92RBKJpNOTC8RFOMK%2BABFTnN0CAQHCHCQTHiuKRTmZFfs5aK5W4xQQJVK8JgAO4KSUEa4MBT8YgsCDiyV4LCCQj0uVWQLC1YXfHPInnLmEmGeJiOei8/moU6qEWnAD6UeQqVSUdoeFiUdNt3u%2B093v9BiDnz5AsZF1FMYA4qEAJIWNwADRrMY9wOzAbzIYFAC9I1Hy1Xa/WAErYULYZ5CR6Z4FNolzQN4ZBRpgKJSta3ayVpu6SmF6/VMWL0VM3TDQm1S%2BGyy6ooH7bDuzOEnnEGFiLyfZnNRaI31a2fzxfLghVx1DcjV1A09wPDcTzXM8ZURS9BCBMwADYHjvCwSTJWhn28N9ZFZBJJ2/HlhTuVJCE%2BAgEE%2BBRWE%2BIQABVQkFBJMgEJlUBpOg7lOfV/lONwDCMABaLhkNOWJ%2BVOBhUDRPiqIEoTgFEpILgAMSnQkCzDMRaCjbcDUjU8DMNSVdKArdz3gtwrxvZ01XQzTTi8TIjF5Wg9JMqM0WLU4sGOAh6VSTAIF0/Sd0c78XIiYBTlQfUGBi8KDW8oyYJAyV4sSoxkv1aCdWlBFoVs8dxyzadmkcP8lwSQDT1otgFwzNwwq8ghFSypK2rQj1MLihKup3PzMHQLwGgESdXWBDh1loTgkl4PwOC0UhUE4NxrGsMVNm2T4Dh4UgCE0Gb1gAaxAfZ9gAOku267vu5D9E4SRFuO1bOF4I0NEO471jgWAkDQLj6DICgICB8iQZQJSag0b6aFoCl2UoWI3tiCIWnpTgDvR5gGQAeVibR4Wx3ggbYQR8YYWgseW3gsEk4AWo8o1uHpzAMSMcQ6dIfBiHhPA4VZlbMFUep%2BV2A7UR6N6k1iOl6Q8LA3pNPAWFJ0g4WIWIMkwUjOeAJMjF%2BvgDGABQADUd3x4KloO/hBBEMR2CkGRBEUFR1B53QFSUlBNssfRkyNSB1lQVI%2BlZ4TxWLUxLGsMwVtQLXiDtTdQ%2B6XpshcBh3E8Do9DCRYKiqPQiiyARpj8BVy76YYS5WTOBYEAYpnz/IFTqcb%2BnmevRmqCZBirvQZ1aPvlmqdYFB2nYJFm%2BbXp5taOHDfxkOE5DJCZJTThqK6NH304IFwQgSHOUEuFWXgjrp1Yzou677qf27HrmjgXtIJak4%2B2wQG%2Bm%2BtB/RgIgFAnFIYJHIJQCG3FEjHBuGJOGfA6BIyNBAVGPNcaYw1pggmRMSZs1IOTRgBAqY0zegzLwTNdKswOlgA23MVp8wFkLN6otxYUg1tLN%2BK05YKyVrsFaqt1YEK1jrJQ%2BsbhG1ALfU2TBzZWwNDbRgGsHbCFEOIV2qiPZqDeroZUfs45WEDnLEO2JVoR2yFHGO%2BxhSGITknFOadTFTx6M3PwEBXDDwVEXco/cy4ZArjkdu1c0gBLrsXPxndXHd1bm0YJI9ol9FiePUund5heMHmPCJE854bC2LPS%2BT0OALU/m9Zeq916b23jcXeyF96H2PvgBE599iX2vr9e%2Bl0brPyfkUj%2BX9eDL0%2Bn/H6t954cDMIvb%2BHB2ljM1qxbOkggA%3D).

So we need to override this feature test macro in *dev/functional/cxx_compiler_quirks.h*:
```c++

#if defined(SQLITE_ORM_CPP20_VIEWS_SUPPORTED) && (defined(__clang__) && (__clang_major__ <= 15))
#undef SQLITE_ORM_CPP20_VIEWS_SUPPORTED
#endif
```

I can do this if it is unclear from my description.

trueqbit (changes requested):
As per comments.

tests/static_tests/function_static_tests.cpp
fnc12: https://github.com/fnc12/sqlite_orm/actions/runs/25061868684/job/73418018018?pr=1482 it doesn't work

dev/statement_serializer.h
fnc12: please check

tests/user_defined_functions.cpp
```diff
@@ -569,7 +571,7 @@ TEST_CASE("generalized scalar udf") {
     storage.sync_schema();
 
     SECTION("freestanding function") {
-        constexpr auto err_fatal_error_f = "ERR_FATAL_ERROR"_scalar.quote(ERR_FATAL_ERROR);
```
trueqbit, line 572: Can we leave this, please? I understand that this is useful in _tests/static_tests/function_static_tests.cpp_ as the function pointer gets reused multiple times. But here we are testing the user-defined functions API and the whole idea about "quoting" is to quote a function directly, not a function pointer defined somewhere else...

tests/user_defined_functions.cpp
```diff
@@ -659,7 +661,7 @@ TEST_CASE("generalized scalar udf") {
     }
 #endif
     SECTION("specialized template function") {
-        constexpr auto clamp_int_f = "clamp_int"_scalar.quote(std::clamp<int>);
```
trueqbit, line 662: Same as for quoting `ERR_FATAL_ERROR` above: Can we leave this, please? ...

tests/user_defined_functions.cpp
```diff
@@ -670,7 +672,7 @@ TEST_CASE("generalized scalar udf") {
     }
     SECTION("overloaded template function") {
         constexpr auto clamp_int_f =
-            "clamp_int"_scalar.quote<const int&(const int&, const int&, const int&)>(std::clamp);
```
trueqbit, line 673: Same as for quoting `ERR_FATAL_ERROR` above: Can we leave this, please? ...

tests/user_defined_functions.cpp
```diff
@@ -703,7 +705,7 @@ TEST_CASE("generalized scalar udf") {
         storage.delete_scalar_function<offset0_f>();
     }
     SECTION("escaped function identifier") {
-        constexpr auto clamp_f = R"("clamp int")"_scalar.quote(std::clamp<int>);
```
trueqbit, line 706: Same as for quoting `ERR_FATAL_ERROR` above: Can we leave this, please? ...

trueqbit (approved):
I think you still need to:
- Update *sqlite_orm.h*?
- Restore `#if defined(__GNUC__) && !defined(__clang__) && (__GNUC__ < 12)` in *tests/user_defined_functions.cpp*.

tests/static_tests/function_static_tests.cpp
trueqbit: :( Would be nice to know the exact gcc-related problem. However skipping the tests is fine and works for now.

PR #1483, opened 2026-05-07
added example from the issue and fixed empty over

### LINKED ISSUE #475: Feature Request for OVER clause  [Window function ?]
Hello mate
Is there a possibilty of OVER cluse support in Lib may be it already exist and I overlooked.
I was trying to achive my pagination logic and needed to use OVER cluase with limit.
my sql logic may be off in this regard but I was trying to achive something like this

select id,firstName,lastName,count(id) OVER() from user_profile order by id limit 50;

it just saves me trouble of writing seperate query for just a count when i add multiple conditions
I am pretty sure this could be achived by left join or somthin similar or I could be shootin blind here 

cheers.

fnc12:


dev/statement_serializer.h
trueqbit: This serializer specialization is unused.

trueqbit: We can reuse `streaming_actions_tuple`:
```c++
ss << " OVER (" << streaming_actions_tuple(arguments, context) << ")";
```

It makes sense to rename `streaming_actions_tuple` to `streaming_definition_tuple`.

In general it would be nice to keep the `[first = true]`, `sep[std::exchange(first, false)]` logic that we have everywhere else.

trueqbit: We can reuse `streaming_actions_tuple`:
```c++
ss << "WINDOW " << statement.name << " AS (" << streaming_actions_tuple(statement.arguments, context) << ")";
```

It makes sense to rename `streaming_actions_tuple` to `streaming_definition_tuple`.

dev/statement_serializer.h
```diff
@@ -548,9 +548,10 @@ namespace sqlite_orm::internal {
     template<class Tuple, class Ctx>
     void serialize_over_arguments(std::stringstream& ss, const Tuple& arguments, const Ctx& context) {
         using args_tuple = std::decay_t<Tuple>;
```
trueqbit: Decay is unnecessary here.

dev/statement_serializer.h
fnc12: removed

dev/statement_serializer.h
fnc12: removed

dev/statement_serializer.h
fnc12: fixed

fnc12: fixed

dev/statement_serializer.h
```diff
@@ -534,33 +534,15 @@ namespace sqlite_orm::internal {
...
-                                                        const Ctx&) SQLITE_ORM_OR_CONST_CALLOP {
-            return statement.name;
-        }
-    };
-
     template<class Tuple, class Ctx>
     void serialize_over_arguments(std::stringstream& ss, const Tuple& arguments, const Ctx& context) {
-        using args_tuple = std::decay_t<Tuple>;
-        if constexpr (std::tuple_size_v<args_tuple> == 0) {
+        if constexpr (std::tuple_size_v<Tuple> == 0) {
```
trueqbit: Oh sorry, I forgot to include this in the last review: we still need to use `std::tuple_size<>::value`. We require C++17 core language features but we still can't rely on a fully implemented C++17 Standard Library.

dev/statement_serializer.h
```diff
@@ -534,33 +534,15 @@ namespace sqlite_orm::internal {
...
             ss << " OVER ()";
-        } else if constexpr (std::tuple_size_v<args_tuple> == 1 &&
-                             std::is_same_v<std::tuple_element_t<0, args_tuple>, window_ref_t>) {
+        } else if constexpr (std::tuple_size_v<Tuple> == 1 &&
```
trueqbit: Use `std::tuple_size<>::value`

trueqbit (approved):

dev/statement_serializer.h
fnc12: fixed

dev/statement_serializer.h
fnc12: fixed

trueqbit (approved):

PR #1486, opened 2026-06-02
Add static_assert to get*, remove to reject non-pk arguments at compile time  

### LINKED ISSUE #1485: Compile-time static assertions for the get* set of operations
## Note

This improvement is listed in the project's [TODO.md](https://github.com/fnc12/sqlite_orm/blob/dev/TODO.md):
> add `static_assert` in crud `get*` functions in case user passes `where_t` instead of id to make compilation error more clear

## Problem

`get<T>()`, `get_pointer<T>()`, `get_optional<T>()` and `get_no_throw<T>()` accept any `Ids...` as arguments. Passing a query expression like `where(...)` instead of a primary key value compiles successfully but crashes at runtime.

## Reproduction

```cpp
struct User { int id; std::string name; };

auto storage = make_storage(":memory:",
    make_table("users",
        make_column("id", &User::id, primary_key()),
        make_column("name", &User::name)));
storage.sync_schema();

// Compiles fine, crashes at runtime with std::system_error: Not found
auto user = storage.get<User>(where(is_equal(&User::name, "Alice")));
```

## Current behavior

Runtime crash:
```
terminate called after throwing an instance of 'std::system_error'
  what():  Not found
```

## Expected behavior

Compile-time error with a clear message, e.g.:
```
static_assert failed: "get() accepts primary key values only. 
Use get_all() with where() to filter by non-PK columns."
```

## Proposed fix

Add a `static_assert` in `get()`, `get_pointer()`, `get_optional()` and `get_no_throw()` checking that all `Ids...` satisfy `is_bindable<T>`, since only bindable types are valid primary key values

Alvov1:
## What

`get()`, `get_pointer()`, `get_optional()`, `get_no_throw()` and `remove()` silently accepted any `Ids...`, including query expressions like `where(...)`. This compiled fine but crashed at runtime with a cryptic `std::system_error`.

Now passing a non-bindable type produces a clear compile-time error:

    error: static assertion failed: Only primary key values are accepted as Ids

## How

Added `static_assert((is_bindable_v<Ids> && ...), ...)` to `get`, `get_pointer`, `get_optional` and `remove`. `get_no_throw` is not checked directly — it delegates to `get_pointer`, which already has the assert.

## Tests

Extended tests with `tests/static_tests/get_pk_ids.cpp` that validates all five functions with `STATIC_REQUIRE` (positive cases only)

Closes #1485

### DIFF
--- a/dev/storage.h
+++ b/dev/storage.h
@@ -392,6 +392,7 @@ namespace sqlite_orm::internal {
          */
         template<class O, class... Ids>
         void remove(Ids... ids) {
+            static_assert((internal::is_bindable_v<Ids> && ...), "Only primary key values are accepted as Ids");
             this->assert_mapped_type<O>();
             auto statement = this->prepare(sqlite_orm::remove<O>(std::forward<Ids>(ids)...));
             this->execute(statement);
@@ -535,6 +536,7 @@ namespace sqlite_orm::internal {
          */
         template<class O, class... Ids>
         O get(Ids... ids) {
+            static_assert((internal::is_bindable_v<Ids> && ...), "Only primary key values are accepted as Ids");
             this->assert_mapped_type<O>();
             this->assert_primary_key_type<O>();
             auto statement = this->prepare(sqlite_orm::get<O>(std::forward<Ids>(ids)...));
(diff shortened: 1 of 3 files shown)


fnc12:
thanks @Alvov1

PR #1491, opened 2026-08-05
Foreign Key Improvements

trueqbit:
* Story: I actually set out to `constexpr`-ify foreign key creation.
* The size of the `foreign_key_t` struct was reduced from 56 to 16 bytes!
* It's still not possible to make foreign key creation `constexpr` because of `on_update`/`on_delete` member variables that provide access to changing foreign key actions. We would need to introduce a breaking change to turn these into methods.

Stacked on PR #1490 [Split column and table constraints into dedicated header files].

### DIFF
--- a/dev/schema/constraints/foreign_key.h
+++ b/dev/schema/constraints/foreign_key.h
@@ -83,7 +83,35 @@ namespace sqlite_orm::internal {
         static_assert(polyfill::is_specialization_of_v<F, foreign_key_t>);
         using foreign_key_type = F;
 
-        foreign_key_type _make_fk(foreign_key_action newAction) const {
+        foreign_key_type no_action() const {
+            return this->copy_fk(foreign_key_action::no_action);
+        }
+
+        foreign_key_type restrict_() const {
+            return this->copy_fk(foreign_key_action::restrict_);
+        }
+
+        foreign_key_type set_null() const {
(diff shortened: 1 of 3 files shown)


dev/schema/constraints/foreign_key.h
```diff
@@ -58,141 +59,102 @@ namespace sqlite_orm::internal {
...
 
         foreign_key_type set_null() const {
-            auto res = this->fk;
-            if (update) {
-                res.on_update._action = foreign_key_action::set_null;
-            } else {
-                res.on_delete._action = foreign_key_action::set_null;
-            }
-            return res;
+            return _make_fk(foreign_key_action::set_null);
```
fnc12: no `this->`?

fnc12 (approved):

dev/schema/constraints/foreign_key.h
trueqbit: No as it starts with an underscore - just keeping the coding style we have at other places (like the `connection_holder`.

However I know you like and prefer the `this` qualifier. Maybe we should come up with a few extra coding preferences I am happy to adapt?

For this specific method I would actually make it private, remove the underscore and `this`-qualify the call.

fnc12: of course, just do it, thanks

fnc12 (approved):

PR #1492, opened 2026-08-06
Mapping tables via C++ reflection

trueqbit:
With this PR tables can also be defined by 'simply' leveraging C++ reflection to automatically map struct fields to columns including most constraints and to specify the table's name with a class-scope annotation.

In other words, tables created with `make_table()` don't require explicit column definitions - the columns are automatically determined from the table's object type through reflection.

```c++
struct[[= "employee"_orm_name]] Employee {
    [[= primary_key()]] int64 id;
    [[= collate_nocase()]] std::string name;
    [[= not_null()]] int64 department_id;
    [[= default_value(0.)]] double salary;
};

struct[[= "department"_orm_name]] Department {
    [[= primary_key()]] int64 id;
    [[= collate_nocase()]] std::string name;
    [[= collate_nocase()]] std::string location;
};

make_storage(
    "",
    make_table<Department>(),
    make_table<Employee>(foreign_key(&Employee::department_id).references(&Department::id)));
```

#### Out of scope (deferred until somebody asks)

- Per-field opt-out via `[[=transient]]`.
- Custom column renaming via annotation.
- Getter/setter columns via reflection.


examples/view.cpp
```diff
@@ -24,28 +23,29 @@
 using namespace sqlite_orm;
 using std::cout;
 using std::endl;
-using std::make_unique;
 using std::string;
 
+void handle_contract_violation(std::contracts::contract_violation const& violation) noexcept {}
```
fnc12: what is this? where is the implementation and where do we call it?

fnc12 (approved):

examples/view.cpp
trueqbit: Oh, that's a left-over of my attempt to compile it in a WSL box that I upgraded to Ubuntu 24.04. It was suddenly a missing symbol during linking. I'll remove it.

fnc12 (approved):

PR #1494, opened 2026-08-08
Harden sync_schema: ADD/DROP COLUMN edge cases, behavior-pinning tests, SQLite version CI matrix

fnc12:
## Fixes

- A new column that cannot be added with `ALTER TABLE ... ADD COLUMN` (column- or table-level `UNIQUE`, `PRIMARY KEY`, non-constant default like `CURRENT_TIMESTAMP`/`datetime(...)`) previously caused a runtime error mid-sync. Now it is classified as `dropped_and_recreated` and goes through the backup table, preserving data.
- Scanning of new columns no longer stops at a STORED generated column, so a NOT NULL column without a default in the same sync is correctly classified as `dropped_and_recreated_with_data_loss` instead of throwing.
- `ALTER TABLE ... DROP COLUMN` rejection (indexed column, primary key member) now falls back to the backup-table dance instead of throwing.
- On SQLite < 3.35 the backup path no longer tries to copy generated columns.
- Outdated comments fixed in `sync_schema_result.h` (stale `data_type` claim), `storage.h` (`sync_schema` doc now lists exactly what is and is not compared) and `table_info.h` (`hidden` semantics).

## Tests

New `tests/sync_schema_matrix_tests.cpp` pins the whole comparison matrix on a real database file: NOT NULL toggle, default presence toggle, regular<->generated, composite PK reorder/extension, plus explicit `already_in_sync` contracts for everything deliberately not compared (column type, default value, generated expression/storage kind, `UNIQUE`/`CHECK`/`COLLATE`, `WITHOUT ROWID`). Every scenario asserts the `sync_schema_simulate() == sync_schema()` invariant.

## CI

New `linux-sqlite-versions` job matrix builds and runs the full suite against pinned SQLite amalgamations **3.31.1 / 3.34.1 / 3.35.5 / 3.50.4**, so both the pre-3.35 and modern sync paths are exercised; the existing vcpkg jobs remain as the latest-version fallback. `tests/CMakeLists.txt` got a `SQLITE_ORM_SERIES_C_TAG` cache variable to download a matching `series.c`, and `issue1410` is gated by 3.34.0 (trigram tokenizer). Both extreme configurations were also verified locally (8496 and 8558 assertions passing).

https://claude.ai/code/session_01NopQNBFjcTmkQ9dzkakpeb


dev/schema/constraints/unique.h
```diff
@@ -24,6 +27,12 @@ namespace sqlite_orm::internal {
 
         constexpr unique_t(columns_tuple columns_) : columns(std::move(columns_)) {}
     };
+
+    template<class T>
+    constexpr bool is_unique_v = std::is_base_of<unique_base, T>::value;
```
trueqbit: Please test for `polyfill::is_specialization_of_v<unique_t>` as `unique_base` will go away in the future.

dev/implementations/storage_definitions.h
```diff
@@ -60,13 +61,27 @@ namespace sqlite_orm::internal {
 
                     if (schema_stat == sync_schema_result::old_columns_removed) {
 #if SQLITE_VERSION_NUMBER >= 3035000  //  DROP COLUMN feature exists (v3.35.0)
-                        for (auto& tableInfo: dbTableInfo) {
-                            this->drop_column(db, table.name, tableInfo.name);
+                        try {
+                            for (auto& tableInfo: dbTableInfo) {
+                                this->drop_column(db, table.name, tableInfo.name);
+                            }
+                        } catch (const std::system_error&) {
```
trueqbit: Let's test for specific SQLite category+error code(s)

dev/storage_base.h
```diff
@@ -1118,6 +1118,26 @@ namespace sqlite_orm::internal {
...
+         *  Checks whether a serialized default value allows the column to be added with
+         *  `ALTER TABLE ... ADD COLUMN`: SQLite only permits constant default values there,
+         *  i.e. no CURRENT_TIME/CURRENT_DATE/CURRENT_TIMESTAMP and no expressions.
+         */
+        static bool is_default_value_addable(const std::string& dfltValue) {
+            if (dfltValue.empty()) {
+                return true;
+            }
+            //  a string literal is always a constant
+            if (dfltValue.front() == '\'' || dfltValue.front() == '"') {
```
trueqbit: A string literal is only one quoted with single quotes [`constexpr char quoteChar = '\'';`]. I know there's a quirk fallback due to MySQL 3.x, but I think SQlite itself will store string literals in the schema definition with only single quotes.

dev/storage_base.h
```diff
@@ -1118,6 +1118,26 @@ namespace sqlite_orm::internal {
...
+                return true;
+            }
+            if (dfltValue == "CURRENT_TIME" || dfltValue == "CURRENT_DATE" || dfltValue == "CURRENT_TIMESTAMP") {
+                return false;
+            }
+            //  everything containing an opening parenthesis is a function call or a parenthesized expression
+            return dfltValue.find('(') == std::string::npos;
```
trueqbit: Nit: though this is perfectly valid, I would prefer an explicit test and return statement in the spirit of this function testing for several conditions.

dev/storage_base.h
```diff
@@ -1118,6 +1118,26 @@ namespace sqlite_orm::internal {
             return storage._busy_handler(triesCount);
         }
 
+        /**
+         *  Checks whether a serialized default value allows the column to be added with
+         *  `ALTER TABLE ... ADD COLUMN`: SQLite only permits constant default values there,
+         *  i.e. no CURRENT_TIME/CURRENT_DATE/CURRENT_TIMESTAMP and no expressions.
+         */
+        static bool is_default_value_addable(const std::string& dfltValue) {
```
trueqbit: `inline` instead of `static`?

tests/sync_schema_tests.cpp
```diff
@@ -624,3 +625,662 @@ TEST_CASE("sync_schema with generated columns") {
...
+    auto storagePath = "sync_schema_matrix_notnull.sqlite";
+    std::remove(storagePath);
+    SECTION("nullable -> NOT NULL") {
+        {
+            auto storage = make_storage(storagePath,
+                                        make_table("users",
+                                                   make_column("id", &UserNullable::id, primary_key()),
+                                                   make_column("age", &UserNullable::age)));
+            storage.sync_schema();
+            UserNullable user;
```
trueqbit: Please initialize directly.

tests/sync_schema_tests.cpp
```diff
@@ -624,3 +625,662 @@ TEST_CASE("sync_schema with generated columns") {
...
+    std::remove(storagePath);
+}
+
+TEST_CASE("sync_schema attribute matrix: composite primary key") {
+    struct Record {
+        int year = 0;
+        int month = 0;
+        int amount = 0;
+
+        bool operator==(const Record& other) const {
```
trueqbit, line 812: Please default if `SQLITE_ORM_DEFAULT_COMPARISONS_SUPPORTED` is defined.

trueqbit (changes requested):
Nice that we have checks for several SQLite versions now!

fnc12:
Thanks for the review! All 7 points addressed in 7c5796c5: `is_unique_v` now tests `polyfill::is_specialization_of_v<T, unique_t>`; the DROP COLUMN fallback rethrows unless the error is `SQLITE_ERROR`; `is_default_value_addable()` is a namespace-scope `inline` function, treats only single-quoted tokens as string literals and uses explicit if/return; tests use direct aggregate initialization and a defaulted `operator==` under `SQLITE_ORM_DEFAULT_COMPARISONS_SUPPORTED`.

dev/schema/constraints/unique.h
fnc12: done, thanks

dev/implementations/storage_definitions.h
fnc12: done, thanks

dev/storage_base.h
fnc12: fixed, thanks

dev/storage_base.h
fnc12: done, thanks

dev/storage_base.h
fnc12: refactored to free `inline` func

tests/sync_schema_tests.cpp
fnc12: done

tests/sync_schema_tests.cpp
fnc12: done, thanks

trueqbit (approved):

PR #1495, opened 2026-08-08
Add missing built-in function wrappers (printf, iif, concat, unixepoch, percentile family, ...)

fnc12:
Adds wrappers for built-in SQLite functions that were missing from `core_functions.h`:

- **Scalar**: `printf` (3.8.3), `iif` (3.32), `sign` (3.35), `format`, `unixepoch` (3.38), `unhex` (3.41), `octet_length`, `timediff` (3.43), `concat`, `concat_ws` (3.44), `if_` (3.48), `unistr`, `unistr_quote` (3.50)
- **Aggregate**: `string_agg` (3.44), `median`, `percentile`, `percentile_cont`, `percentile_disc` (3.51)

Each wrapper is gated by the `SQLITE_VERSION_NUMBER` it appeared in. The percentile family is gated by `SQLITE_ENABLE_PERCENTILE` (following the math-functions precedent — the amalgamation only compiles it in with that define). `iif`/`if_` compute the result type as the common type of the branch arguments, overridable via a template argument like `coalesce`.

Runtime tests + statement serializer tests for every function; verified locally against SQLite 3.31.1, 3.51.0 (with percentile) and the system SQLite. The recent-version CI job is bumped from 3.50.4 to 3.51.0 with `SQLITE_ENABLE_PERCENTILE`. Completed `iif()` and scalar math functions entries are removed from TODO.md.

### DIFF
--- a/dev/core_functions.h
+++ b/dev/core_functions.h
@@ -357,6 +357,134 @@ namespace sqlite_orm::internal {
             return "GROUP_CONCAT";
         }
     };
+
+#if SQLITE_VERSION_NUMBER >= 3008003
+    struct printf_string {
+        serialize_result_type serialize() const {
+            return "PRINTF";
+        }
+    };
+#endif
+
+#if SQLITE_VERSION_NUMBER >= 3032000
+    struct iif_string {
+        serialize_result_type serialize() const {
+            return "IIF";
+        }
+    };
+#endif
+
+#if SQLITE_VERSION_NUMBER >= 3035000
+    struct sign_string {
+        serialize_result_type serialize() const {
+            return "SIGN";
+        }
+    };
+#endif
+
+#if SQLITE_VERSION_NUMBER >= 3038000
+    struct format_string {
+        serialize_result_type serialize() const {
+            return "FORMAT";
+        }
+    };
+
+    struct unixepoch_string {
+        serialize_result_type serialize() const {
+            return "UNIXEPOCH";
+        }
+    };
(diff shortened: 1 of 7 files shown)


dev/core_functions.h
```diff
@@ -1933,6 +2061,229 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+#endif
+
+#if SQLITE_VERSION_NUMBER >= 3048000
+    /**
+     *  IF(X,Y,Z) function, an alias for IIF() https://www.sqlite.org/lang_corefunc.html#iif
+     *
+     *  The return type is the common type of Y and Z, unless it is explicitly specified as a template argument.
+     */
+    template<class R = void, class X, class Y, class Z>
+    constexpr auto if_(X x, Y y, Z z) -> internal::built_in_function_t<
```
trueqbit: I suggest copying `nullif`'s mechanism: enable the function only if a common field type can be determined. This means the conditional computation of the return type in the function body can be simplified to only the common type.

trueqbit (approved):

dev/core_functions.h
fnc12: Done in 7dee61b8 — iif/if_ now follow nullif's mechanism: SFINAE-enabled only when a common type of Y and Z is detectable (or R is explicit), with the return type selected via if constexpr in the body.

trueqbit (approved):

PR #1496, opened 2026-08-11
Dedicated (internal) vocabulary layer that separates classification of DSL nodes

trueqbit:
Introduces a dedicated `vocabulary` layer that separates classification of DSL nodes from composed logic built on top of it and decouples node roles/capabilities from concrete node structs.

**Benefits:**
- Decouples code operating on node roles/capabilities from concrete node structs; all of sqlite_orm depends on the vocabulary, not on struct identity.
- Favors one shared, named vocabulary, reducing dependencies and thinking about which headers to include.
- Establishes a repeatable placement rule (open/per-subject → traits; closed/composed → algorithms; concept-eager → separate `_concepts.h`) for future traits and predicates.

No behavioral changes; this is a structural pass hoisting already existing node traits into another meta layer as first-class citizens and establishing naming and placement conventions for future traits/predicates.

Details (especially for AI-assisted tooling):

- `vocabulary/traits/`: open, per-node/per-type customization points as `_fwd.h` files (grammar, semantic, structural, operand traits), specialized directly at each concrete node's own header in `schema/`/`ast/`. Closed accessor projections (`constraints_type_t`, `field_type_t`, ...) live alongside as `node_projections.h`.
- `vocabulary/algorithms/`: closed, composed predicates and index filters built from the vocabulary (`predicates.h`, `accessors.h`, `index_filters.h`, `field_predicates_fwd.h`/`.h`), consumed for validity checks, static assertions, overload constraining and uniform programming.
- `node_traits.h` / `node_algorithms.h`: declaration-only umbrellas aggregating the respective tier.
- `node_definitions.h` / `node_algorithm_definitions.h`: explicit manifests guaranteeing all specializations/definitions are pulled into the build, now that the `_fwd` split removed incidental transitive inclusion.
- The open/closed distinction is explicit in file naming (`_fwd.h` = extension point, plain `.h` = closed/derived), so intent is visible without reading implementation.
- Relocated `field_of.h` (is_field_of_v, compare_fields) next to `member_traits.h`, since both operate on pointer-to-member mechanics one level below the DSL vocabulary.

Field-level predicates (`is_rowid_alias_capable_v`, hidden-vtab-field checks) follow a fwd/impl split where the definition carries real header weight worth isolating; C++20 concept counterparts, which cannot be forward-declared, are kept eagerly visible in a separate `field_predicates_concepts.h`, together with their C++17 fallback.


dev/ast/cross_join.h
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
fnc12, line 1: lol how could we forget it...

dev/select_constraints.h
```diff
@@ -42,7 +42,7 @@ namespace sqlite_orm::internal {
     };
 
     /**
-     *  DISCTINCT generic container.
+     *  DISTINCT generic container.
```
fnc12, line 45: thank you!

dev/table_name_collector.h
```diff
@@ -30,64 +28,67 @@ namespace sqlite_orm::internal {
...
+            // ...
+            else if constexpr (is_object_node_v<ColRef> || polyfill::is_specialization_of_v<ColRef, table_rowid_t> ||
+                               polyfill::is_specialization_of_v<ColRef, table_oid_t> ||
+                               polyfill::is_specialization_of_v<ColRef, table__rowid_t>) {
+                using table_type = type_t<ColRef>;
+                this->table_names.emplace(lookup_table_name<table_type>(this->db_objects), "");
+            }
+            // ...
+            else {
+                // Do nothing for other types of expressions
```
fnc12, line 74: maybe we should add `static_assert(false)` here but I am not sure

examples/user_defined_functions.cpp
```diff
@@ -151,12 +151,12 @@ int main() {
      *  This function can be called at any time doesn't matter whether connection is open or not.
      *  To delete created scalar function use `storage.delete_scalar_function<T>()` function call.
      */
-    storage.create_scalar_function<sign>();
+    storage.create_scalar_function<signum>();
```
fnc12, line 154: this is a good catch!

fnc12 (approved):

examples/user_defined_functions.cpp
trueqbit: I noticed gcc was picking on it :)

dev/table_name_collector.h
trueqbit: This is not possible as it is called for each node during AST iteration.

fnc12 (approved):

PR #1497, opened 2026-08-13
Add SAVEPOINT support

fnc12:
Implements `SAVEPOINT` / `RELEASE SAVEPOINT` / `ROLLBACK TO SAVEPOINT` (https://www.sqlite.org/lang_savepoint.html), mirroring the existing transaction API on all three levels:

- **Manual**: `storage.savepoint("name")`, `storage.release_savepoint("name")`, `storage.rollback_to_savepoint("name")` — same connection handling and `no_active_transaction` programming-error checks as `begin_transaction()`/`commit()`/`rollback()`.
- **RAII**: `storage.savepoint_guard("name")` returns a `savepoint_t` (new `dev/savepoint.h`, modeled after `transaction_guard_t`): the destructor executes `ROLLBACK TO SAVEPOINT` + `RELEASE SAVEPOINT` unless `release()` was called or `release_on_destroy` is set. `rollback_to()` is deliberately non-terminal — the savepoint stays on the transaction stack per SQLite semantics, so it can be used in retry loops.
- **Functional**: `storage.savepoint("name", lambda)` mirroring `storage.transaction(lambda)` — `true` releases, `false`/exception rolls back.

Savepoint names are quoted with `streaming_identifier`. Tests cover the manual/guard/functional APIs including nesting inside `transaction_guard`, same-name savepoint stacks (`RELEASE` pops to the most recent match), the inner-release-undone-by-outer-rollback semantics and unknown-name errors; logger tests assert the exact emitted SQL. Comes with `examples/savepoint.cpp` (partial rollback + retry loop) and removes the SAVEPOINT entry from TODO.md.

https://claude.ai/code/session_01NopQNBFjcTmkQ9dzkakpeb

### DIFF
--- a/dev/savepoint.h
+++ b/dev/savepoint.h
@@ -70,7 +70,7 @@ namespace sqlite_orm::internal {
             this->rollback_to_func();
         }
 
-      protected:
+      private:
         connection_ref connection;
         std::function<void()> release_func;
         std::function<void()> rollback_to_func;
--- a/dev/savepoint.h
+++ b/dev/savepoint.h
@@ -1,0 +1,79 @@
+#pragma once
+
+#ifndef SQLITE_ORM_IMPORT_STD_MODULE
+#include <functional>  //  std::function
+#include <utility>  //  std::move
+#endif
+
+#include "connection_holder.h"
+
+namespace sqlite_orm::internal {
+    /**
+     *  Class used as a guard for a savepoint. Calls `ROLLBACK TO` and `RELEASE` in destructor.
+     *  Has explicit `release()` and `rollback_to()` functions. After the `release()` function is fired
+     *  the guard won't do anything in its d-tor. Note that unlike a rollback of a transaction
+     *  `rollback_to()` doesn't finish the savepoint: the savepoint remains on the transaction stack,
+     *  so `rollback_to()` may be called multiple times, and the guard is still armed afterwards.
+     *  Also you can set `release_on_destroy` to true to make the guard call `RELEASE` only on destroy,
+     *  keeping the changes.
+     *
+     *  Note: The guard's destructor is explicitly marked as potentially throwing,
+     *  so exceptions that occur during release or rollback are propagated to the caller.
+     */
+    struct savepoint_t {
+        /**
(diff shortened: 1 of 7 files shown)


dev/savepoint.h
```diff
@@ -0,0 +1,79 @@
...
+        /**
+         *  Call `ROLLBACK TO` explicitly. The savepoint remains on the
+         *  transaction stack, so the guard is still armed after this call:
+         *  its destructor will fire unless `release()` is called.
+         */
+        void rollback_to() {
+            this->rollback_to_func();
+        }
+
+      protected:
```
trueqbit: We should change accessibility to `private` unless needed. In general.

dev/storage_base.h
```diff
@@ -257,6 +278,12 @@ namespace sqlite_orm::internal {
             this->executor.perform_void_exec(db, sql.c_str());
         }
 
+        static std::string savepoint_sql(serialize_arg_type statementPrefix, const std::string& savepointName) {
+            std::stringstream ss;
+            ss << statementPrefix << streaming_identifier(savepointName) << std::flush;
```
trueqbit: No `std::flush` needed as we return from the function.

examples/savepoint.cpp
```diff
@@ -0,0 +1,79 @@
...
+ *  ROLLBACK TO SAVEPOINT name;          -> storage.rollback_to_savepoint("name") / guard.rollback_to()
+ */
+#include <sqlite_orm/sqlite_orm.h>
+#include <iostream>
+
+using std::cout;
+using std::endl;
+
+struct Order {
+    int id = 0;
```
trueqbit: Let's use `int64` for primary columns!

trueqbit (approved):

dev/savepoint.h
fnc12: Done in eebb8f8c.

dev/storage_base.h
fnc12: Done in eebb8f8c.

examples/savepoint.cpp
fnc12: Done in eebb8f8c.

PR #1499, opened 2026-08-14
Sync database objects in dependency order

fnc12:
`sync_schema()` used to iterate database objects in **reverse declaration order** — a legacy artifact which forced indexes/triggers to be declared before their tables (breaking code generated by sqlite2orm in straight order) and, worse, **silently lost dependent objects**: with a straight declaration order an index or trigger was synced before its table, so when the table was recreated through the backup procedure after a schema change, its indexes and triggers were destroyed with it and nothing recreated them until the next sync.

## What this does

The sync order is now computed **entirely at compile time** in the new `dev/vocabulary/algorithms/sync_order.h`: a stable topological sort over the db objects tuple which keeps the declaration order and only defers an index or a trigger until after its target table (or view, for `INSTEAD OF` triggers). `sync_schema()`/`sync_schema_simulate()` iterate via the resulting `index_sequence` — zero runtime cost, no dynamic containers.

## Why the graph is this small (verified empirically)

| reference | at creation time |
|---|---|
| index → its table | **hard** (`no such table`) |
| trigger → its `ON` table/view | **hard** (`no such table`) |
| table → FK parent (incl. cross- and self-references) | lazy — validated at DML time |
| view → selected tables/views | lazy |
| trigger body → other tables | lazy |
| fts5 `content=` → external content table | lazy (and a runtime string anyway) |

So the hard-dependency graph is trivially acyclic (indexes/triggers point at tables/views, which depend on nothing) — no cycle handling is needed, and cross-referencing foreign keys need **no special treatment**, which a dedicated test proves with foreign key enforcement enabled.

## Tests

- creation from a straight declaration order (the sqlite2orm case) — previously a runtime error;
- dependents recreated after every table-recreation flavor (`dropped_and_recreated`, backup paths) — previously silently lost;
- in-place alterations (`new_columns_added`) keep dependents untouched;
- interleaved declarations, fans of dependents, self-/cross-referencing FKs;
- compile-time `STATIC_REQUIRE`s on the computed sequence, including the conservative handling of multiple tables mapping the same C++ type.

Both declaration orders keep working: legacy reverse-order schemas are unaffected, straight-order schemas become valid.


dev/schema/algorithms/sync_order.h
trueqbit, line 1: `sync_order.h` should sit under a new folder `schema/algorithms`.
Reason: Everything under vocabulary/algorithms/ so far — predicates, index_filters, accessors — consumes the classification vocabulary (grammar/semantic/structural/operand traits) to answer questions framed in those terms ("is this a select expression," "give me the main DML node," "which indices are columns").
A dependency-order computation over database objects is operating on schema relationships.

Just a note: this will anyway become a good place for things like 'storage lookup'.

dev/vocabulary/algorithms/sync_order.h
```diff
@@ -0,0 +1,162 @@
...
+#endif
+
+    template<class DBO>
+    using sync_dependency_source_t = typename sync_dependency_source<DBO>::type;
+
+    /**
+     *  Whether database object `D` must be synced after database object `S`.
+     */
+    template<class D, class S>
+    inline constexpr bool sync_depends_on_v =
```
trueqbit: No `inline` necessary as it is a (variable) template. I began to remove it in all other places already.

dev/vocabulary/algorithms/sync_order.h
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     *  Computes the order in which the database objects must be synced as an array of tuple indices:
+     *  a stable topological sort which keeps the declaration order and only defers indexes and triggers
+     *  until after their target table or view.
+     *  The dependency graph is trivially acyclic since indexes and triggers depend on tables and views only,
+     *  which in turn depend on nothing.
+     */
+    template<class... DBO>
+    constexpr std::array<size_t, sizeof...(DBO)> compute_sync_order() {
+        constexpr size_t size = sizeof...(DBO);
```
trueqbit: Maybe we can rename the variable from `size` to something more meaningful like `nDBOs` similar to other places like in `storage_t<>::assert_updatable_type()`.

dev/vocabulary/algorithms/sync_order.h
```diff
@@ -0,0 +1,162 @@
...
+
+    template<class DBO>
+    using sync_dependency_target_t = typename sync_dependency_target<DBO>::type;
+
+    /**
+     *  Maps a database object to the mapped object type it provides as a dependency target,
+     *  or `void` if it cannot be a target of a hard dependency.
+     */
+    template<class DBO>
+    struct sync_dependency_source {
```
trueqbit: D

dev/vocabulary/algorithms/sync_order.h
```diff
@@ -0,0 +1,162 @@
...
+
+#ifdef SQLITE_ORM_WITH_VIEW
+    template<class O, class Select, class... Cs>
+    struct sync_dependency_source<query_view<O, Select, Cs...>> {
+        using type = O;
+    };
+#endif
+
+    template<class DBO>
+    using sync_dependency_source_t = typename sync_dependency_source<DBO>::type;
```
trueqbit: Simply alias `sync_dependency_source_t` to the `object_type_t` alias template projection, please. Include `dev/vocabulary/node_traits.h` to have it available.
The `sync_dependency_source` struct can then be removed.

dev/vocabulary/algorithms/sync_order.h
```diff
@@ -0,0 +1,162 @@
...
+
+    template<class O, class Select, class... Cs>
+    struct query_view;
+
+    /**
+     *  Maps a database object to the mapped object type of the schema object
+     *  it hard-depends on at creation time, or `void` if it depends on nothing.
+     */
+    template<class DBO>
+    struct sync_dependency_target {
```
trueqbit: As noted in my comment on `sync_source_target_t` below on R78 I would actually like to see this replaced by a simple new node projection. `index_t` has already `table_mapped_type`, so this should be in `trigger_t` as well, then the new alias template projector is called `table_mapped_type_t`. Maybe think about the best type name, however it's fine if we keep `table_mapped_type`.

trueqbit (changes requested):
Wow, awesome @fnc12!

dev/schema/algorithms/sync_order.h
fnc12: Done in 53f176d4.

dev/vocabulary/algorithms/sync_order.h
fnc12: Done in 53f176d4.

dev/vocabulary/algorithms/sync_order.h
fnc12: Done in 53f176d4.

dev/vocabulary/algorithms/sync_order.h
fnc12: Done in 53f176d4.

dev/vocabulary/algorithms/sync_order.h
fnc12: Done in 53f176d4.

trueqbit (approved):

PR #1500, opened 2026-08-15
Compile-time strictness for statement clauses (DSL strictness, phase 1)

fnc12:
First phase of tightening the DSL surface so that only meaningful types are accepted where they belong ([full audit](https://github.com/fnc12/sqlite_orm/blob/dev/TODO.md) — the query static check TODO item). Empirical probes showed that misplaced or misnested clauses compiled and generated **invalid SQL**:

| expression | generated SQL before |
|---|---|
| `select(x, group_by(...), where(...))` | `GROUP BY ... WHERE ...` |
| `select(x, limit(5), order_by(...))` | `LIMIT 5 ORDER BY ...` |
| `select(x, where(...), cross_join<V>())` | `WHERE ... CROSS JOIN ...` |
| `where(group_by(...))` / `where(where(...))` | `WHERE (GROUP BY ...)` / `WHERE (WHERE ...)` |
| `order_by(limit(5))` | `ORDER BY LIMIT 5` |
| `columns(&U::id, where(...))` | `SELECT "id", WHERE (...)` |
| `select(&U::id, 42)` | `SELECT "id" FROM "users" 42` |

All of these are now compile errors with readable messages. `validate_conditions()` gains three asserts driven by a `clause_rank_v` classification over the grammar traits:

1. **whitelist** — every conditions-pack argument must be a recognized statement clause;
2. **canonical order** — FROM → JOINs → WHERE → GROUP BY → ORDER BY → LIMIT (the serializer streams clauses positionally, so any other order is broken SQL);
3. **no clause inside a clause** — the payloads of `where`/`group_by`(+`.having`)/`order_by`/`multi_order_by`/`limit` must be expressions.

`columns()`/`struct_()` reject statement clauses among projections; `union_`/`union_all`/`except`/`intersect` now require select statements (previously only the arity was checked).

**Deliberately not restricted** (raw SQL is equally permissive, verified by probes): bindable literals as conditions (`WHERE 42` is valid SQL), subselects/expressions in `LIMIT` (`limit(select(count<User>()))` keeps working — probed), borderline `ORDER BY` literals. The predicates are constexpr and covered positively and negatively by `STATIC_REQUIRE`s in `tests/static_tests/conditions_validity.cpp`; the whole existing test suite compiles unchanged, defining the compatibility envelope.

Next phases (planned): operator-parity constraints for the named comparison/arithmetic factories, and schema-level strictness (make_index columns, using_fts5, trigger chain, make_storage spec rejection).

[...]

PR #1502, opened 2026-08-25
Gate the named expression factories on valid operands (DSL strictness, phase 2)

fnc12:
Second phase of the DSL strictness work (#1500 was phase 1). The named expression factories (`eq`/`lt`/`and_`/`like`/`between`/`in`/`is_null`/`add`/`assign`/...) accepted arbitrary types: `eq(User{}, 5)`, `add(std::vector<int>{}, 1)`, `assign(42, "x")`, `exists(42)`, `like(User{}, "a%")` all compiled and failed with 200-line template errors deep in the serializer, or generated invalid SQL.

## The gate

Every argument must now be **a bindable value or a recognized operand**, asserted with a readable message. The predicates live in the new `vocabulary/algorithms/operand_predicates.h`:

- `is_operand_or_bindable<T>` — operator-argument vocabulary ∪ member pointers ∪ bindable values ∪ scalar subqueries ∪ row-value tuples ∪ `std::reference_wrapper` (prepared-statement rebinding);
- `are_valid_operands<L, R>` — both sides checked;
- `is_referencable_operand<T>` — the stricter gate for `assign()`'s target;
- `exists()` requires a select statement.

**Deliberately as permissive as SQL itself**: raw literals stay valid operands, because the named factories are the only spelling for literal-only expressions — `bitwise_or(60, 13)` ⇒ `SELECT 60 | 13` (a tested use case), `and_(42, 42)` ⇒ valid SQL. What's rejected is what can never serialize: arbitrary structs, containers, statement clauses.

## Vocabulary additions surfaced by the compatibility envelope

Compiling the full test suite + examples against the gate revealed operand kinds missing from the operand vocabulary, now registered at their node headers per the vocabulary rules: `new_t`/`old_t` (trigger `NEW.`/`OLD.`), `excluded_t` (upsert), `cast_t`, `simple_case_t`. These registrations also benefit the overloaded operators.

Positive and negative `STATIC_REQUIRE`s in `tests/static_tests/operand_validity.cpp`; the whole suite and all examples compile unchanged.

### DIFF
--- a/dev/ast/between.h
+++ b/dev/ast/between.h
@@ -4,6 +4,7 @@
 #include <utility>  //  std::move
 #endif
 #include "tags.h"
+#include "../vocabulary/algorithms/operand_predicates.h"
 
 namespace sqlite_orm::internal {
     /**
@@ -31,6 +32,9 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
      */
     template<class A, class T>
     internal::between_t<A, T> between(A expression, T lower, T upper) {
+        static_assert(internal::is_operand_or_bindable<A>::value,
+                      "the tested expression must be a bindable value or one of sqlite_orm-recognized operands: member "
+                      "pointers, column pointers, c()-wrapped values, aliases or expressions");
         return {std::move(expression), std::move(lower), std::move(upper)};
     }
 }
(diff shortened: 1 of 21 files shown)


trueqbit (approved):
Immediately followed by phase 3 in PR #1503!

PR #1507, opened 2026-08-28
Draw the C++17 Standard Library base line

trueqbit:
Draws the C++17 Standard Library base line, enforces it in `cxx_check_prerequisites.h`, and prunes the polyfills and feature guards it retires. Rationale and base line: `docs/plans/2026-08-28-cxx17-stdlib-baseline.md`.

We drew the *compiler* base line deliberately (#1433, #1465, #1469) and left the
*library* side open just as deliberately, because a C++17 compiler still turns up
on a pre-C++17 stdlib. fdc33d69 names the running cost of that split — "as long
as we don't require a C++17 Standard Library base line we cannot assume that
variable templates are available" — and it's also ~72 `SQLITE_ORM_*_SUPPORTED`
sites plus the polyfills.

The question is *when*, not *how*: draw it now, or wait for the C++20 line where
`<version>` is guaranteed and the probe collapses to `#include <version>`.

A `<version>`-else-`<ciso646>` probe with a hard `__cpp_lib_*` check enforces:

| stdlib | first version that passes | ruled out |
|---|---|---|
| libstdc++ | **GCC 9** | GCC 7-8 — no `<version>`, and `<ciso646>` yields 0 `__cpp_lib_*` |
| libc++ | **libc++ 8** (~Xcode 11) | libc++ 7 — ships `<version>` as an empty stub |
| MSVC STL | **VS 2017 15.9** | VS 2015, early VS 2017 |

Measured against libstdc++ 13/15, MSVC 14.16/14.29/14.44 and the libc++ release
tags; all of it sits below our CI matrix (g++-10, clang-12, windows-2022,
macos-14).

The catch worth arguing about: the cut is really "C++17 stdlib **with a populated
`<version>`**", so GCC 7-8 and libc++ 7 fall even though they have the features —
exactly the toolchains the current policy protects. Separately, the Apple
deployment target (macOS 10.14 / iOS 12) is invisible to any macro test, so that
part becomes documentation rather than detection.

If we draw it now, `README.md`'s "C++17 compatible compiler" becomes named
versions. That's the call I'd like a second opinion on before any guards come out.

The second commit adds the enforcement block itself so CI exercises the base line
across the matrix. It is deliberately separable — drop it and the PR is plan-only.


fnc12 (approved):

tests/tests.cpp
```diff
@@ -103,11 +103,7 @@ TEST_CASE("Limits") {
 }
 
 TEST_CASE("Custom collate") {
-#if defined(SQLITE_ORM_STRING_VIEW_SUPPORTED) || SQLITE_VERSION_NUMBER >= 3008008
```
fnc12, line 106: why did we remove sqlite version number check?

fnc12 (approved):

tests/tests.cpp
trueqbit: Thanks for catching this. The version number check was already made dead in f6ac5d13 by flipping `&&` to `||`, I assume accidentally by me back then! Claude simply followed correct logic here and removed the check entirely...

fnc12 (approved):

PR #1508, opened 2026-08-29
Hoist the pack-indexed moniker digit into a variable for clang before 20.1

fnc12:
[...]

PR #1509, opened 2026-08-29
Apply index redefinitions in sync_schema

fnc12:
Redefining an index under an existing name — changed columns, a changed ordering or collation, or switching to `make_unique_index()` — used to be a silent no-op: `sync_schema()` executed `CREATE INDEX IF NOT EXISTS`, which leaves the stale index in place, and `schema_status()` reported `already_in_sync` unconditionally. The database kept serving the old index while the storage declared the new one.

The fix follows the pattern triggers and views already use: `schema_status()` compares the serialization against the statement text stored in the schema table, and `sync_dbo()` drops and recreates the index when they differ. Two consequences:

- The index serializer no longer emits `IF NOT EXISTS` — SQLite strips it from the text it stores in `sqlite_master`, so the comparison must be verbatim; creation is guarded by `schema_status()` instead, as for the other schema objects. Indexes created by previous sqlite_orm versions compare equal, since the stored text never contained the clause.
- `sync_schema()` now reports honest statuses for indexes: `new_table_created` on first creation and `dropped_and_recreated` on redefinition, instead of `already_in_sync` always.

Note: the first commit is #1508 (the clang pack-indexing quirk), which this branch needs to build on current Xcode; it will drop out once #1508 lands and this is rebased.

### DIFF
--- a/tests/statement_serializer_tests/schema/index.cpp
+++ b/tests/statement_serializer_tests/schema/index.cpp
@@ -18,37 +18,37 @@ TEST_CASE("statement_serializer index") {
     SECTION("simple") {
         auto index = make_index("id_index", &User::id);
         value = internal::serialize(index, context);
-        expected = R"(CREATE INDEX IF NOT EXISTS "id_index" ON "users" ("id"))";
+        expected = R"(CREATE INDEX "id_index" ON "users" ("id"))";
     }
     SECTION("desc") {
         auto index = make_index("idx_users_id", indexed_column(&User::id).desc());
         value = internal::serialize(index, context);
-        expected = R"(CREATE INDEX IF NOT EXISTS "idx_users_id" ON "users" ("id" DESC))";
+        expected = R"(CREATE INDEX "idx_users_id" ON "users" ("id" DESC))";
     }
     SECTION("asc") {
         auto index = make_index("idx_users_id", indexed_column(&User::id).asc());
         value = internal::serialize(index, context);
-        expected = R"(CREATE INDEX IF NOT EXISTS "idx_users_id" ON "users" ("id" ASC))";
+        expected = R"(CREATE INDEX "idx_users_id" ON "users" ("id" ASC))";
     }
     SECTION("collate") {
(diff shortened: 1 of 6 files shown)


dev/functional/cxx_compiler_quirks.h
```diff
@@ -82,3 +82,12 @@
...
 #define SQLITE_ORM_BROKEN_CPP20_VIEWS
 #endif
+
+// clang implements C++26 pack indexing (and advertises it via `__cpp_pack_indexing` as an extension
+// in earlier language modes), but as of Apple clang 17 indexing a non-type template parameter pack
+// produces an expression of '<dependent type>' when its value is used directly in a constant
+// expression context such as `static_assert` or a variable initializer, in every language mode.
+// Returning the indexed value from a function is unaffected.
+#if defined(SQLITE_ORM_PACK_INDEXING_SUPPORTED) && defined(__clang__)
+#define SQLITE_ORM_BROKEN_NTTP_PACK_INDEXING_VALUE_USE
```
fnc12: fix for Xcode 16

tests/statement_serializer_tests/schema/index.cpp
```diff
@@ -60,7 +60,7 @@ TEST_CASE("statement_serializer index") {
             auto index = make_index<User>("idx", indexed_column(json_extract<bool>(&User::name, "$.field")));
             value = internal::serialize(index, context);
         }
-        expected = "CREATE INDEX IF NOT EXISTS \"idx\" ON \"users\" (JSON_EXTRACT(\"name\", '$.field'))";
+        expected = "CREATE INDEX \"idx\" ON \"users\" (JSON_EXTRACT(\"name\", '$.field'))";
```
trueqbit: Let's convert this into a raw-string literal!

trueqbit (approved):
Seems to be quite important and was overlooked in PR #1480...

tests/statement_serializer_tests/schema/index.cpp
fnc12: done

Issue #1445, opened 2025-08-20
Encryption support


itsRythem:
With this library, is it possible to keep the database on disk encrypted and decrypt it in memory at runtime when it is used?

fnc12:
yep. Most encryption plugins for SQLite I saw require to run encode func right after opening. That's why `storage` class has `on_open`  callback which can be used like this:

```c++
storage.on_open = [] (sqlite3 *db) {
    encode(db);
};
```

itsRythem:
im confused by your response, what im looking for is a way to keep the file on disk (the database file), always encrypted. Only decrypting it when its used, in memory, never decrypting the file on disk

fnc12:
Please tell me how would you use encryption with raw SQLite without sqlite_orm

fnc12:
@itsRythem are you there?

PR #1510, opened 2026-08-29
Rebuild TODO.md from a full sweep of the SQLite release notes

fnc12:
Walked every 3.x release in [changes.html](https://sqlite.org/changes.html) — 248 releases, ~2500 change items — filtered to what an ORM interface library can implement, verified each candidate against `dev/` (greps, not memory), and recorded the remainder in TODO.md, grouped by area: SQL language, functions, schema synchronisation, C API/runtime.

Notable additions found by the sweep that the previous list lacked: `RIGHT`/`FULL OUTER JOIN`, `IS (NOT) DISTINCT FROM`, `NULLS FIRST/LAST`, ORDER BY inside aggregates, HAVING without GROUP BY, the 3.35 UPSERT generalizations, row values, `INDEXED BY`, TEMP schema objects, `DEFERRABLE` FKs, `VACUUM INTO`/`ANALYZE`/`REINDEX`/`EXPLAIN`, UDF function flags (`SQLITE_DETERMINISTIC` & co.), `sqlite3_serialize`/`deserialize`, the small-C-API and PRAGMA wrapper batches, FTS5 `bm25()`/`snippet()`, and the session extension.

Removed as already implemented: the clause-order static check (done by the DSL strictness phases) and the vague "rest of core functions" entry — the verified remainder (`likely()` family, introspection functions, `substring()` alias) is now listed explicitly.

Verified as already implemented and therefore *not* listed (spot-checks during the sweep): `match()`, `json_quote`/`json_patch`, `unicode()`, `zeroblob()`, multi-row VALUES, LIMIT/ORDER BY on DELETE/UPDATE, indexes on expressions, table-valued functions in FROM, window functions incl. EXCLUDE/chaining, `VACUUM`, `table_xinfo` and the already-wrapped pragmas.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -2,19 +2,50 @@
 
 `sqlite_orm` is a wonderful library but there are still features that are not implemented. Here you can find a list of them:
 
+## SQL language
+
 * `RETURNING` clause for INSERT/UPDATE/DELETE https://sqlite.org/lang_returning.html
-* `FOREIGN KEY` - sync_schema fk comparison and ability of two tables to have fk to each other (`PRAGMA foreign_key_list(%table_name%);` may be useful)
-* rest of core functions(https://sqlite.org/lang_corefunc.html)
-* `ATTACH`
-* blob incremental I/O https://sqlite.org/c3ref/blob_open.html
-* query static check for correct order (e.g. `GROUP BY` after `WHERE`)
-* named constraints: constraint can have name `CREATE TABLE heroes(id INTEGER CONSTRAINT pk PRIMARY KEY)`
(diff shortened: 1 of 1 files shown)


trueqbit (approved):
Perhaps add encryption support via the [SQLite Encryption Extension (SEE)](https://sqlite.org/com/see.html)? I think this corresponds to https://github.com/fnc12/sqlite_orm/issues/1445.

fnc12:
Added — SEE with `sqlite3_key`/`sqlite3_rekey`, plus a note on compatible implementations like SQLCipher with their `PRAGMA key` spelling, referencing #1445.

PR #1513, opened 2026-08-30
Add the planner-hint functions likely(), unlikely() and likelihood()

fnc12:
Second S-tier item of the feature-design pass: `likely()` (3.8.6), `unlikely()` and `likelihood()` (3.8.1), gated accordingly. All three pass their argument through unchanged, so the wrappers return `field_type_or_type_t` of the argument (the `iif()` mechanism).

The subtle part is binding, verified against the C API first: SQLite requires the second argument of `likelihood()` to be a floating point *constant* between 0.0 and 1.0 — preparing `SELECT likelihood(a, ?)` fails outright. The factory therefore takes a `double` and stores it in a `literal_holder`, which serializes inline and never binds; a prepared-statement test with a bound `WHERE` value next to the literal probability pins the behavior.

Serializer tests, runtime tests (single closing assertion over sections) and examples included.

### DIFF
--- a/dev/core_functions.h
+++ b/dev/core_functions.h
@@ -370,11 +370,17 @@ namespace sqlite_orm::internal {
     };
 
 #if SQLITE_VERSION_NUMBER >= 3008001
-    //  tag only; the serialization lives in the statement serializer
-    struct likelihood_t {};
+    struct likelihood_string {
+        std::string_view serialize() const {
+            return "LIKELIHOOD";
+        }
+    };
 
-    //  tag only; the serialization lives in the statement serializer
-    struct unlikely_t {};
+    struct unlikely_string {
+        std::string_view serialize() const {
+            return "UNLIKELY";
+        }
+    };
 #endif
 
 #if SQLITE_VERSION_NUMBER >= 3008003
@@ -386,8 +392,11 @@ namespace sqlite_orm::internal {
 #endif
 
 #if SQLITE_VERSION_NUMBER >= 3008006
-    //  tag only; the serialization lives in the statement serializer
-    struct likely_t {};
+    struct likely_string {
+        std::string_view serialize() const {
+            return "LIKELY";
+        }
+    };
 #endif
 
 #if SQLITE_VERSION_NUMBER >= 3032000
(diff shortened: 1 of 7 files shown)


dev/core_functions.h
```diff
@@ -2087,6 +2110,33 @@ SQLITE_ORM_EXPORT namespace sqlite_orm {
...
+     *  The probability is stored as a literal, never as a bound parameter:
+     *  SQLite requires the second argument to be a floating point constant between 0.0 and 1.0,
+     *  and rejects a statement that binds it.
+     */
+    template<class X>
+    constexpr internal::built_in_function_t<internal::field_type_or_type_t<X>,
+                                            internal::likelihood_string,
+                                            X,
+                                            internal::literal_holder<double>>
+    likelihood(X x, double probability) {
```
trueqbit, line 2127: Let's check `probability`'s value if constant-evaluated...

trueqbit (approved):

fnc12:
Done — when the call is constant-evaluated, an out-of-range probability fails compilation (`std::is_constant_evaluated()` + throw, guarded by `__cpp_lib_is_constant_evaluated`); verified both directions with compile probes (`constexpr auto bad = likelihood(20, 1.5);` errors, `0.5` stays constant-evaluable). At runtime the branch is dead and SQLite's own prepare-time error remains the check.

dev/core_functions.h
fnc12: please check - I am not sure whether I did it right

dev/core_functions.h
```diff
@@ -369,17 +369,11 @@ namespace sqlite_orm::internal {
     };
 
 #if SQLITE_VERSION_NUMBER >= 3008001
-    struct likelihood_string {
-        std::string_view serialize() const {
-            return "LIKELIHOOD";
-        }
-    };
+    //  tag only; the serialization lives in the statement serializer
```
trueqbit: Hmm, I am not voting for this.

It breaks with all the current built-in functions.
It is better to handle them in one sweep - AND I have an idea how to generalize this using the now established "C++20 aliases", where we could simply define all built-in functions with their name and signature.

fnc12:
Reverted to the `*_string` convention — agreed that three functions spelled differently is an inconsistency, not a migration. The constant-evaluated probability check and the feature macro stay. Looking forward to the one-sweep generalization over the C++20 aliases; happy to help with it when you have the design.

dev/core_functions.h
fnc12: reverted. Please check

trueqbit (approved):
Hmm, there's warning sign next to `sqlite_orm.h` for this commit...

fnc12:
@trueqbit yeah I also see it in my and your PR today

PR #1517, opened 2026-08-31
Add vacuum_into()

fnc12:
Fourth S-tier item of the feature-design pass: `storage.vacuum_into(filename)` next to `vacuum()`, gated on SQLite 3.27.

`VACUUM INTO` writes a vacuumed, transactionally consistent copy of the database into a new file, leaving the database itself untouched — a one-statement compacted backup that complements the page-wise backup API (clone-as-is, incremental) with a rebuild-from-scratch copy (defragmented, smaller). The filename is quoted via `quote_string_literal`.

Tests: a roundtrip (populate → vacuum_into → open the copy → compare rows) and the error path (the target must not exist — vacuuming into an existing file throws).

### DIFF
--- a/dev/storage_base.h
+++ b/dev/storage_base.h
@@ -248,9 +248,9 @@ namespace sqlite_orm::internal {
          *  into a new file, leaving the database itself untouched. The file must not exist yet.
          *  More info: https://www.sqlite.org/lang_vacuum.html#vacuuminto
          */
-        void vacuum_into(const std::string& filename) {
+        void vacuum_into(std::string_view filename) {
             std::stringstream ss;
-            ss << "VACUUM INTO " << quote_string_literal(filename) << std::flush;
+            ss << "VACUUM INTO " << quote_string_literal(std::string{filename}) << std::flush;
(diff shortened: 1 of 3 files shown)


dev/storage_base.h
```diff
@@ -242,6 +242,20 @@ namespace sqlite_orm::internal {
             this->executor.perform_void_exec(connection.get(), "VACUUM");
         }
 
+#if SQLITE_VERSION_NUMBER >= 3027000
+        /**
+         *  `VACUUM INTO` query: writes a vacuumed, transactionally consistent copy of the database
+         *  into a new file, leaving the database itself untouched. The file must not exist yet.
+         *  More info: https://www.sqlite.org/lang_vacuum.html#vacuuminto
+         */
+        void vacuum_into(const std::string& filename) {
```
trueqbit: Should we use `std::string_view` now that it's always available?

trueqbit (approved):

fnc12:
Done — `std::string_view` by value, matching the storage API (`drop_view` et al.).

PR #1519, opened 2026-09-01
Add the small C API wrappers batch

fnc12:
Sixth S-tier item of the feature-design pass — the mechanical C API wrappers, all following the `changes()`/`busy_handler()` precedents.

Connection side (`storage_base`):
- `changes64()` / `total_changes64()` (3.37)
- `interrupt()` + `is_interrupted()` (3.41) — probed first: an interrupt issued while idle stays reported until the next statement
- `txn_state()` across all schemas and `txn_state(schema)` for one, with the new `transaction_state` enum (`dev/transaction_state.h`, 3.34); an unknown schema is an empty optional
- `error_offset()` (3.38)
- `db_name(index)` (3.39) — empty optional when out of range
- `db_readonly(schema)` beside the existing parameterless `db_readonly()` (which I only discovered mid-build — hence the overload, not a default argument)
- `setlk_timeout(ms, blockOnConnect)` (3.50) — probed: returns SQLITE_OK on builds without blocking-lock support too

Statement side (`prepared_statement_base`): `readonly()`, `busy()`, `is_explain()` (3.28).

Every behavioral claim in the tests was probed against the C API before wrapping.

### DIFF
--- a/dev/storage_base.h
+++ b/dev/storage_base.h
@@ -410,9 +410,9 @@ namespace sqlite_orm::internal {
          *  sqlite3_is_interrupted function: whether an interrupt is currently in effect
          *  on this connection.
          */
-        bool is_interrupted() {
+        int is_interrupted() {
             auto connection = this->get_connection();
-            return sqlite3_is_interrupted(connection.get()) != 0;
+            return sqlite3_is_interrupted(connection.get());
         }
 #endif
 
--- a/dev/storage_base.h
+++ b/dev/storage_base.h
@@ -7,6 +7,7 @@
 #include <functional>  //  std::function, std::bind, std::bind_front
 #include <string>  //  std::string
 #include <string_view>  //  std::string_view
+#include <optional>  //  std::optional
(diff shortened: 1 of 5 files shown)


dev/storage_base.h
```diff
@@ -376,6 +378,117 @@ namespace sqlite_orm::internal {
...
+
+#if SQLITE_VERSION_NUMBER >= 3039000
+        /**
+         *  sqlite3_db_name function: the name of the schema at the given index.
+         *  Returns an empty optional if the index is out of range.
+         */
+        std::optional<std::string> db_name(int index) {
+            auto connection = this->get_connection();
+            if (orm_gsl::czstring name = sqlite3_db_name(connection.get(), index)) {
+                return std::string{name};
```
trueqbit: Simply return `name` otherwise a temporary `string` is created unnecessarily.

trueqbit (approved):

fnc12:
Done — `return name;` converts in place.

PR #1522, opened 2026-09-02
Add NULLS FIRST / NULLS LAST to ORDER BY

fnc12:
Eighth S-tier item of the feature-design pass: the null-placement clause of SQLite 3.30.

`nulls_first()` / `nulls_last()` are copy-builders on `order_by_t` beside `asc()`/`desc()`, so they compose in any order with the direction and collation builders — `.collate_nocase().desc().nulls_last()` and `.nulls_last().desc()` serialize identically (pinned by a test). The tri-state lives in `order_by_base`, which makes the single ordering term, `multi_order_by()` and `dynamic_order_by()` (propagated through `push_back`) all carry it; the suffix is emitted at the one shared serialization choke point, after ASC/DESC as the ordering-term grammar prescribes. Gated on `SQLITE_VERSION_NUMBER >= 3030000`.

Tests cover the full combination matrix:
- serializer: bare, asc/desc × first/last, builder-order independence, collation + direction + placement, `multi_order_by` with mixed placements;
- runtime: the SQLite defaults pinned first (ASC puts NULLs first, DESC puts them last), then both non-default placements flipping them, plus a `dynamic_order_by` entry.

### DIFF
--- a/tests/select_constraints_tests.cpp
+++ b/tests/select_constraints_tests.cpp
@@ -2,6 +2,7 @@
 #include <catch2/catch_all.hpp>
 
 #include <ctime>  //  std::time_t
+#include <optional>  //  std::optional
 
 using namespace sqlite_orm;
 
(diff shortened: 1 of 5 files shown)


tests/select_constraints_tests.cpp
trueqbit: Compilers warn on signed/unsigned mismatch...

trueqbit (approved):

fnc12:
Fixed — `static_cast<size_t>` on the `count()` side.

PR #1523, opened 2026-09-03
Treat signed/unsigned comparison mismatches as errors

fnc12:
Follow-up to the review find on #1522: make the class of bug fail CI instead of relying on reviewer eyes.

Follows the `-Werror=reorder` precedent on the `sqlite_orm` interface target (so tests and examples inherit it in every configuration): `-Werror=sign-compare` for GCC/Clang — the warning is not part of any enabled set there today, so it was invisible — and `/we4018 /we4388 /we4389` for MSVC, where `/W4` showed the warnings without failing.

A clean full build (tests + all 48 examples) under the flag surfaced three violations, all in test code, fixed here: two `int(...) == name.length() + 1000` comparisons in `tests/operators/arithmetic_operators.cpp` and the `rows.size() == storage.count<Track2>()` from the #1522 review (fixed identically there; the merges reconcile). The library headers compile clean under the flag.

Note: like `-Werror=reorder`, the interface-target flag propagates to downstream CMake consumers of `sqlite_orm::sqlite_orm`.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -77,6 +77,16 @@ if(MSVC)
         # C4458: 'symbol': deprecated
         # disabled due to unit tests testing deprecated features
         /wd4996)
+    if(MSVC_VERSION GREATER_EQUAL 1930)
+        target_compile_options(unit_tests PUBLIC
+            # third-party headers (Catch2) are exempt from warnings; VS 2022+
+            /external:anglebrackets
+            /external:W0
+            # signed/unsigned comparison mismatches are errors: C4018 (relational), C4388/C4389 (equality)
+            /we4018
+            /we4388
+            /we4389)
+    endif()
     if (CMAKE_CXX_FLAGS MATCHES "/D_UNICODE")
         # explicitly set the entry point of the executable file,
         # otherwise for some reason the linker will not pick up `wmain`, which is provided by the static Catch2 library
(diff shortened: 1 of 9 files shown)


fnc12:
Reworked after the experimental VS 2017 failure: C4388 fired inside Catch2's own headers on v141, which also exposed that an interface-target flag imposes the errors on downstream consumers' code — unlike `-Werror=reorder`, sign-compare fires anywhere, not just at our class definitions. The flags are now scoped to the `unit_tests` and example targets; on MSVC they require VS 2022+ and come with `/external:anglebrackets /external:W0` so third-party headers stay exempt. The v141 job builds as before.

tests/CMakeLists.txt
```diff
@@ -77,6 +77,16 @@ if(MSVC)
         # C4458: 'symbol': deprecated
         # disabled due to unit tests testing deprecated features
         /wd4996)
+    if(MSVC_VERSION GREATER_EQUAL 1930)
+        target_compile_options(unit_tests PUBLIC
+            # third-party headers (Catch2) are exempt from warnings; VS 2022+
+            /external:anglebrackets
```
trueqbit, line 83: The intent is great but I think this disables all warnings for sqlite_orm code as well (not just the unit tests). This is because sqlite_orm is included by angle brackets...

fnc12:
Closing in favor of the full -Wall/-Wextra warnings cleanup (option B) discussed on Discord — the /external:anglebrackets concern raised here will be addressed properly there.

PR #1525, opened 2026-09-07
Add extension loading support

fnc12:
## Summary

* `storage.enable_load_extension(bool)` — toggles extension loading through `sqlite3_db_config`'s `SQLITE_DBCONFIG_ENABLE_LOAD_EXTENSION` option, which enables the C API while deliberately leaving the `load_extension()` SQL function disabled, as [recommended for security](https://www.sqlite.org/c3ref/c_dbconfig_defensive.html#sqlitedbconfigenableloadextension). Returns the C API result code.
* `storage.load_extension(file[, entryPoint])` — wraps `sqlite3_load_extension`, propagating its error message (e.g. the dlopen failure text) in the thrown `std::system_error`.

No wrapper for the `load_extension()` SQL function — deliberate, per the security recommendation above.

## Availability gate

Both wrappers sit behind a new `SQLITE_ORM_LOAD_EXTENSION_SUPPORTED` macro in `functional/sqlite3_config.h` (next to `SQLITE_ORM_JSON_SUPPORTED`): on unless SQLite was built with `SQLITE_OMIT_LOAD_EXTENSION`, and off on Apple platforms by default — Apple's system SQLite strips `sqlite3_load_extension` from both the library and its header *without* defining the omit macro (while still shipping `sqlite3ext.h`, so no `__has_include` telltale exists). Building against an unrestricted SQLite on Apple (Homebrew, vcpkg) can be declared with `SQLITE_ORM_ENABLE_LOAD_EXTENSION`.

## Testing

* New `load extension` test: the load of an inexistent file must fail with a translated error both with loading disabled (default) and enabled. The assertion is deliberately just `REQUIRE_THROWS_AS`, since Debian-family SQLite builds enable the C API by default, changing which error fires.
* Verified the wrappers empirically against Homebrew SQLite 3.53.3 on macOS (`not authorized` → enable → dlopen error propagated → disable) and Ubuntu's system SQLite on Linux.
* Full unit test suite passes; all examples build.


dev/storage_base.h
```diff
@@ -77,6 +77,9 @@ namespace sqlite_orm::internal {
         std::map<std::string, collating_function> collatingFunctions;
         const int cachedForeignKeysCount;
         std::function<int(int)> _busy_handler;
+#ifdef SQLITE_ORM_LOAD_EXTENSION_SUPPORTED
+        std::optional<bool> _loadExtensionEnabled;
```
trueqbit, line 81: Either remove the leading underscore or do not `this`-qualify accessing this member...

dev/storage_base.h
```diff
@@ -495,6 +498,53 @@ namespace sqlite_orm::internal {
...
+        }
+
+        /**
+         *  sqlite3_load_extension function: loads a run-time loadable extension from a shared library file.
+         *  `entryPoint` names the extension's initialization function;
+         *  if empty, SQLite derives the name from the file name.
+         *  Extension loading must have been turned on with `enable_load_extension()` beforehand.
+         *  More info: https://www.sqlite.org/c3ref/load_extension.html
+         */
+        void load_extension(std::string_view file, std::string_view entryPoint = {}) {
```
trueqbit: `load_extension()` should accept parameters of type `const std::string&` as `sqlite3_load_extension()` requires null-terminated C strings.

trueqbit: `load_extension()` should accept parameters of type `const std::string&` as `sqlite3_load_extension()` requires null-terminated C strings.

Also the method should declare `errorMessage` with type `orm_gsl::zstring` and automatically free `errorMessage`.

All in all `load_extension()` can be simplified to:

```c++
        void load_extension(const std::string& file, const std::string& entryPoint = {}) {
            orm_gsl::zstring errorMessage = nullptr;
            const int rc = sqlite3_load_extension(connection.get(),
                                                  file.c_str(),
                                                  entryPoint.empty() ? nullptr : entryPoint.c_str(),
                                                  &errorMessage);
            if (rc != SQLITE_OK) {
                const scope_guard freeErrorMessageGuard{[errorMessage] {
                    sqlite3_free(errorMessage);
                }};
                throw std::system_error{sqlite_errc(rc), errorMessage ? errorMessage : ""};
            }
        }
```

trueqbit (changes requested):

trueqbit (approved):

PR #1526, opened 2026-09-09
Simplification

trueqbit:
A round of internal simplification, plus a small user-visible widening of what `select()` accepts as a literal.

### Header move

`type_traits.h` moves to `functional/type_traits.h`. Everything in it is language-level
machinery with no DSL knowledge — the same subject matter as the rest of `functional/`,
which it already depended on and which never includes anything from above itself. At
each of its 37 consumers the include joins the other `functional/` ones, restoring the
bottom-up order the header blocks otherwise follow.

### Generic ref forwarding

`get_ref_t`/`get_object_t` were class templates specialized per DML node to answer a
question that isn't about DSL nodes at all: hand me the object behind a member that may
or may not be a `std::reference_wrapper`. They are replaced by `forward_ref()`,
`forward_lvalue_ref()` and `unwrap_ref_or_forward_t`; the per-node `get_object_t`
specializations collapse into a single `access_dml_object()`, since insert, replace and
update all merely carried an `object` member.

Adds `polyfill::forward_like`, used by the tuple transformers and by
`quoted_expression_t` — whose deducing-this overloads now hinge on
`SQLITE_ORM_DEDUCING_THIS_SUPPORTED` alone instead of additionally requiring the C++23
library feature.

### CTE column name collection

`cte_column_names_collector` no longer iterates the table's columns by hand into a
reserved vector; it builds the name list with `create_from_tuple` over
`col_index_sequence_of` and `&column_identifier::name`, reusing the generic tuple
machinery instead of restating the traversal.

### String literals

`string_data` helpers are gone in favour of `std::string_view`/`std::wstring_view`, and
raw `const wchar_t*` is spelled `orm_gsl::cwzstring` in binders and serializers.
`column_result_t` now accepts `std::string_view`, `std::wstring_view`, `std::wstring`
and `orm_gsl::cwzstring` for simple `SELECT 'literal'` queries, all yielding
`std::string`.

### Result-type proxies

`unique_ptr_result_of` is replaced by `nullable_result_proxy`, which carries the
`expression_type` and so deduces `std::unique_ptr` results of built-in functions —
aggregates such as `MIN`/`MAX` in particular — more robustly, consolidating several
`column_result_t` specializations.

### `statement_binder`

`bind()` and `result()` are qualified consistently via `SQLITE_ORM_STATIC_CALLOP` /
`SQLITE_ORM_OR_CONST_CALLOP`, and `result()` is explicitly deleted for the
`std::unique_ptr` and `std::optional` specializations: direct extraction into those
nullable types is not supported.

### Tests

New coverage in `tests/simple_query.cpp` and `tests/static_tests/column_result_t.cpp`
for the widened literal support.


dev/cte_column_names_collector.h
```diff
@@ -106,13 +107,11 @@ namespace sqlite_orm::internal {
...
-            std::vector<std::string> columnNames;
-            columnNames.reserve(size_t(table.template count_of<is_column>()));
-
-            table.for_each_column([&columnNames](const column_identifier& column) {
-                columnNames.push_back(column.name);
-            });
-            return columnNames;
+            using table_type = polyfill::remove_cvref_t<decltype(table)>;
+            using column_index_sequence = col_index_sequence_of<elements_type_t<table_type>>;
+            return create_from_tuple<std::vector<std::string>>(table.elements,
```
fnc12, line 112: this notation looks amazing (but less intuitive, but it is not about our code, it is about modern C++ 🤓)

fnc12 (approved):

PR #1527, opened 2026-09-10
Feature/dml vocabulary

trueqbit:
The DML nodes were the last big family still matched and decomposed by concrete type. They are
now classified through the vocabulary layer like every other node, and live where the layering
rules say they should.

- Lifted the DML node traits into `vocabulary/traits/grammar_traits_fwd.h` and added the
  missing `is_insert_explicit`, `is_update`, `is_remove`. The statement serializer, AST
  iterator, node tuple and `storage_t` match and decompose through those instead of through
  `insert_t<T>` & co.
- Lifted the clause traits the DML factories consume along with them: `is_into`,
  `is_values`/`is_dynamic_values`, `is_set`/`is_dynamic_set`/`is_any_set`, `is_upsert_clause`.
  `is_set` now denotes `set_t` alone; `is_any_set` covers either spelling.
- `is_raw_dml_expression` / `is_object_dml_expression` are specialized per node, next to that
  node's own grammar trait.
- Moved the DML expressions out of `prepared_statement.h` into `ast/dml/` — one header per
  statement kind, plus the clause nodes only they take (`into`, `set`, `default_values`,
  `upsert_clause`) — and `expression_object_type.h` into
  `vocabulary/algorithms/accessors.h`, where seven per-node specializations collapse to two.
- Fixes a gap the trait work exposed: `insert(into<T>(), values(vector))` was rejected, and
  would have been serialized with one set of parentheses too many. Both now go through the
  `is_any_values` grouping trait.

No public API change beyond that last point, which only widens what compiles.


dev/ast_iterator.h
```diff
@@ -372,40 +358,40 @@ namespace sqlite_orm::internal {
...
-        using node_type = update_all_t<S, Wargs...>;
+    template<class T>
+    struct ast_iterator<T, std::enable_if_t<is_update_all_v<T>>> {
+        using node_type = T;
 
         template<class L>
-        SQLITE_ORM_STATIC_CALLOP void operator()(const node_type& u, L& lambda) SQLITE_ORM_OR_CONST_CALLOP {
-            iterate_ast(u.set, lambda);
-            iterate_ast(u.conditions, lambda);
+        SQLITE_ORM_STATIC_CALLOP void operator()(const node_type& node, L& lambda) SQLITE_ORM_OR_CONST_CALLOP {
```
fnc12, line 366: yay!

fnc12 (approved):

PR #1528, opened 2026-09-10
Add the FTS5 bm25() and snippet() auxiliary functions

fnc12:
## Summary

* `bm25(anyField[, weights...])` — the [FTS5 bm25 function](https://www.sqlite.org/fts5.html#the_bm25_function), optionally taking one weight per table column; the result type is `double`.
* `snippet(anyField, columnIndex, matchOpen, matchClose, ellipses, tokenCount)` — the [FTS5 snippet function](https://www.sqlite.org/fts5.html#the_snippet_function); the result type is `std::string`.

Both factories mirror the modern `highlight()` overloads: they identify the FTS5 table via the hidden 'any' field (`Post::hidden::any_field` or the member pointer form), in both the C++20-concepts and the SFINAE flavors.

## Design

Instead of cloning `highlight_t` twice, a single generic node `internal::fts5_auxiliary_function_t<R, S, T, Args...>` derives from `built_in_function_t<R, S, Args...>`, so AST iteration, `node_tuple` and `column_result_t` all come from the existing generic machinery. Only two FTS5-gated additions were needed elsewhere:

* a `statement_serializer` partial specialization that serializes the looked-up table name as the first SQL argument — `BM25("posts", ...)` — followed by the streamed argument tuple;
* a `table_name_collector` branch (like `highlight_t`'s), so the FROM table is inferred when only the function is selected.

## Testing

* New serializer test pinning the generated SQL for `bm25`, `bm25` with weights, and `snippet`, using the same harness as the `match` serializer test.
* Runtime tests in the FTS5 section of `virtual_table.cpp`: `snippet` against a match whose title is short enough to come back whole (keeping the expectation independent of the snippet-picking algorithm across SQLite versions), and `bm25` with and without weights asserting the documented negative-score contract.
* Full unit test suite passes locally; amalgamation regenerated; formatter-clean.


dev/core_functions.h
trueqbit: Let's morph this into a `fts5_auxiliary_function_t` as it is one of them!

dev/core_functions.h
```diff
@@ -839,6 +839,40 @@ namespace sqlite_orm::internal {
...
+    };
+
+    /*
+     *  An FTS5 auxiliary function: a function whose first argument in SQL is the FTS5 table itself,
+     *  which is identified by the mapped object type `T` and serialized as the looked-up table name.
+     *  Deliberately not a `built_in_function_t`, so that the general built-in function machinery,
+     *  which knows nothing about the table argument, never matches it.
+     */
+    template<class R, class S, class T, class... Args>
+    struct fts5_auxiliary_function_t : S {
```
trueqbit: We also need a grammar trait `is_fts_auxiliary_function_v`/`is_fts_auxiliary_function` in `dev/vocabulary/traits/grammar_traits_fwd.h`.

dev/node_tuple.h
```diff
@@ -74,6 +74,11 @@ namespace sqlite_orm::internal {
     template<class T, class X, class Y, class Z>
     struct node_tuple<highlight_t<T, X, Y, Z>, void> : node_tuple_for<X, Y, Z> {};
 
+#if SQLITE_VERSION_NUMBER >= 3009000 || defined(SQLITE_ORM_ENABLE_FTS5)
+    template<class R, class S, class T, class... Args>
+    struct node_tuple<fts5_auxiliary_function_t<R, S, T, Args...>, void> : node_tuple_for<Args...> {};
```
trueqbit: Match by vocabulary trait `is_fts_auxiliary_function`.

dev/statement_serializer.h
```diff
@@ -283,6 +283,26 @@ namespace sqlite_orm::internal {
         }
     };
 
+#if SQLITE_VERSION_NUMBER >= 3009000 || defined(SQLITE_ORM_ENABLE_FTS5)
+    template<class R, class S, class T, class... Args>
+    struct statement_serializer<fts5_auxiliary_function_t<R, S, T, Args...>, void> {
```
trueqbit: Match by vocabulary trait `is_fts_auxiliary_function`.

dev/table_name_collector.h
trueqbit: Can be removed if this is a `fts5_auxiliary_function_t`.

trueqbit (changes requested):

trueqbit (commented):
I think it makes sense to move "fts5" functions under `dev/ast/fts5_functions.h` now that we are moving in this direction in general.

dev/ast_iterator.h
```diff
@@ -14,6 +14,7 @@
 #include "alias.h"
 #include "operators.h"
 #include "core_functions.h"
+#include "ast/fts5_functions.h"
```
trueqbit: Unnecessary as the vocabulary is available via `vocabulary/node_traits.h` included above and the definitions are later aggregated via umbrella header `node_definitions.h`

dev/column_result.h
```diff
@@ -19,6 +19,7 @@
 #include "vocabulary/node_traits.h"
 #include "mapped_type_proxy.h"
 #include "core_functions.h"
+#include "ast/fts5_functions.h"
```
trueqbit: Unnecessary...

dev/node_tuple.h
```diff
@@ -14,6 +14,7 @@
 #include "prepared_statement.h"
 #include "optional_container.h"
 #include "core_functions.h"
+#include "ast/fts5_functions.h"
```
trueqbit: Unnecessary...

dev/statement_serializer.h
```diff
@@ -35,6 +35,7 @@
 #include "ast/is_null.h"
 #include "ast/is_not_null.h"
 #include "core_functions.h"
+#include "ast/fts5_functions.h"
```
trueqbit: Unnecessary...

dev/table_name_collector.h
```diff
@@ -11,6 +11,7 @@
 #include "rowid.h"
 #include "alias.h"
 #include "core_functions.h"
+#include "ast/fts5_functions.h"
```
trueqbit: Unnecessary...

trueqbit (approved):

PR #1530, opened 2026-09-12
Default the return type of built-in function factories

### LINKED ISSUE #1529: Built-in function factories: default the explicit return type overloads
Every math function, plus `sign`, `json_array_length`, `json_type` and `json_remove`, comes in two overloads: one with a fixed return type and one taking it as a template argument. The pairs can be folded into the templated overload with the fixed type as default (`template<class R = double, class X> acos(X)`), which halves the API surface and merges the duplicated doc comments.

While at it, close the runtime coverage gaps: `typeof_`, `unicode`, `random`, `last_insert_rowid`, `changes`, `total_changes`, `soundex`, `unistr_quote` and the aggregates `count`, `avg`, `sum`, `total`, `max`, `min`, `group_concat` were only serialized, static-checked or used incidentally by other tests.

trueqbit:
Fixes #1529.

- Fold each fixed-return/templated overload pair of the math functions, `sign`, `json_array_length`, `json_type` and `json_remove` into one factory with a defaulted `R`, and merge the doc comments.
- Exercise the remaining built-in functions at runtime (`typeof_`, `unicode`, `random`, `last_insert_rowid`, `changes`, `total_changes`, `soundex`, `unistr_quote`, and the aggregates incl. empty-group results); drop a duplicated `tanh` section.
- Document that the v141 toolset must be built without `/permissive-` (VS 2017 fails to emit two `std::forward_as_tuple` specializations used by the statement serializer); the VS 2017 CI job passes only because CMake never adds the switch.


tests/built_in_functions_tests/core_functions_tests.cpp
```diff
@@ -1471,6 +1478,153 @@ TEST_CASE("substring") {
...
+    }
+    SECTION("text") {
+        auto rows = storage.select(typeof_("abc"));
+        decltype(rows) expected{"text"};
+        REQUIRE(rows == expected);
+    }
+    SECTION("null") {
+        auto rows = storage.select(typeof_(nullptr));
+        decltype(rows) expected{"null"};
+        REQUIRE(rows == expected);
```
fnc12: nit: we can have only one `REQUIRE(rows == expected);` in the bottom so this test will perform the same but in less amount of lines

tests/built_in_functions_tests/core_functions_tests.cpp
```diff
@@ -1471,6 +1478,153 @@ TEST_CASE("substring") {
...
+    SECTION("max") {
+        auto rows = storage.select(sqlite_orm::max(&Score::value));
+        REQUIRE(rows.size() == 1);
+        REQUIRE(rows.front());
+        REQUIRE(*rows.front() == 3);
+    }
+    SECTION("max of an empty group is null") {
+        auto rows = storage.select(sqlite_orm::max(&Score::value), where(c(&Score::id) > 3));
+        REQUIRE(rows.size() == 1);
+        REQUIRE_FALSE(rows.front());
```
fnc12: I'd avoid parting comparing and it is better to replace it with a single `REQUIRE(rows == expected);` cause it is more readable

fnc12 (approved):

tests/built_in_functions_tests/core_functions_tests.cpp
trueqbit: Done in a7725d5c: `rows`/`expected` hoisted out of the sections, single `REQUIRE(rows == expected)` at the end.

tests/built_in_functions_tests/core_functions_tests.cpp
trueqbit: `sum`, `max` and `min` return `std::vector<std::unique_ptr<T>>`, and `unique_ptr` equality compares addresses, so `REQUIRE(rows == expected)` would fail for any non-null value (it only works for the empty-group cases, where both sides hold null). The three-step check is the same one `tests/row_id.cpp` uses for `max(rowid)`. I left these as they are; happy to change if you have a preferred idiom for `unique_ptr` rows.

fnc12: I am sure we can specify custom comparison function in catch2 or using C++ api. But it is not required though

fnc12 (approved):

tests/built_in_functions_tests/core_functions_tests.cpp
trueqbit: Done in bd8dc128: `tests/catch_matchers.h` now has `PointeesEqual`, a thin wrapper over Catch2's `RangeEquals(range, predicate)` that compares what the nullable elements hold, so these become e.g. `REQUIRE_THAT(rows, PointeesEqual<double>({6.0}));` / `PointeesEqual<int>({std::nullopt})`. Also applied it to the same pattern in `json.cpp` and `row_id.cpp`.

fnc12 (approved):

PR #1533, opened 2026-09-14
Hoist generic trait and documentation changes

trueqbit:
Generic changes that surfaced while working on the C++20-style built-in functions, separated out so that they can land on their own.

- `common_type_of` moves from `tuple_helper/same_or_void.h` to `functional/type_traits.h`: it is a generic trait over a pack, and `column_result.h` included `same_or_void.h` for it alone.
- New field predicate `is_text_value` (narrow/wide C string, string view, string); the six `column_result_t` specializations yielding `std::string` for a text value collapse into one over the predicate.
- The `orm_function_sig` concept moves from `function.h` to `functional/type_traits.h`, next to the internal `stateless` concept.
- `docs/internals/vocabulary-layer.md` gains an "Including the layer" section: consumers include the umbrellas, never a `traits/`, `projections/` or `algorithms/` file directly; the `*_fwd.h` files and the definitions manifest are the only exceptions.


dev/vocabulary/algorithms/field_predicates.h
```diff
@@ -36,4 +39,12 @@ namespace sqlite_orm::internal {
                                                   (sizeof(F) != sizeof(sqlite_int64) ||
                                                    std::is_signed<F>::value != std::is_signed<sqlite_int64>::value)>> =
             true;
+
+    template<class T>
+    constexpr bool is_text_value_v = std::disjunction<std::is_same<T, orm_gsl::czstring>,
```
fnc12, line 44: this is really neat!

fnc12 (approved):

dev/vocabulary/algorithms/field_predicates.h
trueqbit: Yeah, finally in our toolbox :)

PR #1535, opened 2026-09-14
Define the built-in functions by name and overload set in C++20 builds

trueqbit:
> Builds on #1537 (`feature/fts5-functions-as-builtins`).

In C++20 builds, every built-in SQL function is now **one definition of its name and its overload set** instead of a per-function `*_string` tag plus `builtin_function_t` factory:

```cpp
inline constexpr auto lower = "LOWER"_builtin.scalar<std::string(std::string_view)>();
inline constexpr auto substr = "SUBSTR"_builtin.scalar<std::string(std::string_view, int),
                                                      std::string(std::string_view, int, int)>();
inline constexpr auto max = "MAX"_builtin.function<aggregate_sig<std::unique_ptr<argument<0>>(anything)>,
                                                   scalar_sig<std::unique_ptr<argument<0>>(anything, anything, variadic<anything>)>>();
```

Calling the definition picks the overload by argument count and yields a call node of the overload's kind — `builtin_function_call`, or `builtin_aggregate_function_call` with `.filter(where(...))` and `.over(...)`. Arities are enforced from the signatures per the SQLite documentation instead of accepted blindly. The call sites and the existing tests are unchanged; the legacy nodes and factories stay in the C++17 branch and go when that baseline does.

### Design points

- **Signatures are nominal.** Parameter types state the SQL contract and fix the arity; no callable receives them and arguments are not checked against them. The return type is not nominal — it is what a select yields. `variadic<T>` marks an open-ended tail, `anything` a parameter that is polymorphic in SQL.
- **Argument-dependent return types** are placeholders — `argument<I>`, `common_argument_type<I...>` — that `column_result_t` substitutes with the schema at hand (that part is #1534, merged, and holds in both standards).
- **Caller-chosen return types** stay: `builtin_function::operator()<R>(...)` overrides the declared return type, and the functions that take `R` as a template argument (`acos<R>(x)`, `median<R>`, `json_extract<R>`, …) are function template facades over the internal definition object. So are `count(X)` and `replace(X,Y,Z)` (they share their name with other function templates), `sqlite_offset`, `likelihood` and the `json_insert` family (they check their arguments), the FTS5 `highlight`/`bm25`/`snippet` (they accept only the table's hidden column), and **`abs`, `round`, `time`, `random`, `strftime`, `printf`**: a built-in whose name is also a global C library function must be a function, since under `using namespace sqlite_orm;` an object and a function of one name are an ambiguous lookup, whereas two functions are an overload set.
- `""_builtin` is deliberately internal — sqlite_orm defines all built-ins itself; the public surface is the `orm_builtin_function` concept.
- `ast/builtin_function.h` is one header for both standards: shared nodes on top, legacy nodes under `#ifndef SQLITE_ORM_WITH_CPP20_ALIASES`, the mechanism under `#else`.
- Definitions need not live in `core_functions.h`: the FTS5 auxiliary functions (#1537) are defined in `vtabs/fts5_functions.h` by the same `"…"_builtin` literal.

Design, decisions (1–14), phases and the merge notes are in `docs/plans/2026-09-12-built-in-function-vocabulary-design.md`; static tests in `tests/static_tests/builtin_function_static_tests.cpp`.

### Known trade-offs

- Only the built-ins published as objects are `orm_builtin_function`s. Every facade — the `R`-taking functions, `count`, `replace`, the argument-checking ones, the C-library-named six — is a function template to the user, and only its internal definition object is a definition. Whether the definitions should be reachable publicly in a uniform way is left open.
- A node's nested `return_type` now holds the placeholder rather than the resolved type where the result is argument-dependent; the result type is `column_result_of_t`'s business (also #1534).


dev/ast/builtin_function.h
```diff
@@ -138,3 +110,331 @@ namespace sqlite_orm::internal {
...
+
+        constexpr std::string_view name() const {
+            return {_nme, N - 1};
+        }
+
+        consteval builtin_function(const char (&name)[N]) {
+            std::copy_n(name, N, _nme);
+        }
+
+        char _nme[N];
```
fnc12, line 380: why not `_name`?

dev/core_functions.h
```diff
@@ -578,165 +584,1408 @@ namespace sqlite_orm::internal {
...
+#if SQLITE_VERSION_NUMBER >= 3007016
+    /**
+     *  CHAR(X1,X2,...,XN) function https://sqlite.org/lang_corefunc.html#char
+     */
+    inline constexpr orm_builtin_function auto char_ = internal::char_;
+
+    /**
+     *  RANDOM() function https://www.sqlite.org/lang_corefunc.html#random
+     */
+    constexpr auto random() {
```
fnc12, line 1560: why not `inline random = internal::random`?

tests/builtin_functions_tests/core_functions_tests.cpp
```diff
@@ -135,6 +135,16 @@ TEST_CASE("lower") {
...
     REQUIRE(rows.front() == "ototo");
+    // built-in function call node as an operator operand
+    auto flags = storage.select(lower("OTOTO") == "ototo");
+    REQUIRE(flags.size() == 1);
+    REQUIRE(flags.front());
+#ifdef SQLITE_ORM_WITH_CPP20_ALIASES
+    // the new call node is a general operator argument, hence chainable
+    auto concatenated = storage.select(lower("OTO") || "TO");
+    REQUIRE(concatenated.size() == 1);
+    REQUIRE(concatenated.front() == "otoTO");
```
fnc12: comparing expected vector will be more beautiful imho

fnc12 (approved):

dev/core_functions.h
trueqbit: [...] and **`abs`, `round`, `time`, `random`, `strftime`, `printf`**: a built-in whose name is also a global C library function must be a function, since under `using namespace sqlite_orm;` an object and a function of one name are an ambiguous lookup, whereas two functions are an overload set.

dev/ast/builtin_function.h
trueqbit: Well, I copied it over from `quoted_scalar_function` and the idea was to reduce confusion with the `name()` method as public accessor.

tests/builtin_functions_tests/core_functions_tests.cpp
trueqbit: Not my construction site, but I'll address them 😃

dev/ast/builtin_function.h
trueqbit: Happy to rename it, though...

fnc12 (approved):

PR #1540, opened 2026-09-17
Fix the connection leak in the transaction begin methods

### LINKED ISSUE #1539: File descriptor leaked due to function not being exception safe
The `storage_base::begin_transaction_internal` method is not exception safe. It invokes `this->connection->retain()`, after which it does a bunch of other stuff that may throw an exception. If such an exception occurs, `this->connection->release()` is never invoked, causing the `connection_holder` to never invoke `sqlite3_close`, so the file handle is leaked.

```
#include <cassert>

int main(int argc, char* argv[]) {
  using namespace sqlite_orm;
  auto storage = make_storage("leaky.sqlite");

  // Use a scope so that the (outer) transaction_guard is automatically destroyed, rolling back the outer transaction
  {
    auto outer = storage.transaction_guard();
    try {
      auto inner = storage.transaction_guard();
    } catch (...) {
      // storage_base::begin_transaction_internal has invoked "this->connection->retain()" and then thrown an exception.
      // Since a transaction_guard_t was never instantiated for the inner transaction, it can't invoke storage_base::rollback,
      // which would have invoked "this->connection->release()".
    }
  }

  // Assertion fails because its connection->retain_count() is still 1
  assert(!storage.is_opened());

  // The sqlite3_close function hasn't been (and won't be) invoked, so the file handle remains open
}
```

The behavior is problematic because the `transaction_guard` mechanism promises RAII compliance and doesn't deliver. From [the documentation](https://fnc12.github.io/sqlite_orm/):

> use `transaction_guard` function which returns a guard object which works just like `lock_guard` for `std::mutex`.

fnc12:
Fixes #1539.

## The bug

`begin_transaction()` (and the deferred/immediate/exclusive variants and `savepoint()`) retained the connection **before** executing the `BEGIN`/`SAVEPOINT` statement. A failing statement — most easily a nested `BEGIN` — threw after the retain, the matching release never happened, and the connection (and its file descriptor) stayed open for the storage's lifetime.

On the current dev the reporter's exact `transaction_guard` scenario no longer leaks (the guard holds an RAII `connection_ref` since the thread-safe connection rework), but the manual methods still had the hole — and on top of that, `commit()`/`rollback()`/`release_savepoint()` lost their `release()` in that same rework, so even a **successful** manual transaction kept the connection open.

## The fix

* The five begin-style methods share a new exception-safe `begin_transaction_internal()`: an RAII `connection_ref` covers the statement execution, and the reference the open transaction owns is retained only once the statement has succeeded.
* `commit()`, `rollback()` and `release_savepoint()` drop that reference again — after a successful statement only, so a failed `COMMIT` leaves the transaction (and its reference) intact. `rollback_to_savepoint()` deliberately keeps it: the savepoint stays on the stack and the transaction continues.

## Testing

New `transaction does not leak the connection` test on a file-based storage (an in-memory one is opened forever, hiding the leak): nested guard, nested manual begin, manual begin+commit / begin+rollback, savepoint release and rollback-to — each section must end with the connection closed. Verified the reporter's repro leaks on v1.9.1 (open fd shown by lsof) and is clean with this fix; full unit test suite passes.


dev/storage_base.h
```diff
@@ -968,6 +976,8 @@ namespace sqlite_orm::internal {
             if (connection_ptr maybeConnection = *this->connection) {
                 this->executor.perform_void_exec(maybeConnection.get(),
                                                  this->savepoint_sql("RELEASE SAVEPOINT ", savepointName).c_str());
+                //  drop the reference owned by the savepoint since `savepoint()`
```
trueqbit: Hmm, dereferencing the shared connection should be moved before executing the `RELEASE SAVEPOINT`. Otherwise the connection is leaked at this point...

dev/storage_base.h
```diff
@@ -994,6 +1004,8 @@ namespace sqlite_orm::internal {
         void commit() {
             if (connection_ptr maybeConnection = *this->connection) {
                 this->executor.perform_void_exec(maybeConnection.get(), "COMMIT");
+                //  drop the reference owned by the transaction since `begin_transaction()`
```
trueqbit: Hmm, dereferencing the shared connection should be moved before executing the `COMMIT`. Otherwise the connection is leaked at this point...

dev/storage_base.h
```diff
@@ -1004,6 +1016,8 @@ namespace sqlite_orm::internal {
         void rollback() {
             if (connection_ptr maybeConnection = *this->connection) {
                 this->executor.perform_void_exec(maybeConnection.get(), "ROLLBACK");
+                //  drop the reference owned by the transaction since `begin_transaction()`
```
trueqbit: Hmm, dereferencing the shared connection should be moved before executing the `ROLLBACK`. Otherwise the connection is leaked at this point...

dev/storage_base.h
fnc12: Fixed in 8dc94e16

dev/storage_base.h
fnc12: Fixed in 8dc94e16

dev/storage_base.h
fnc12: Fixed in 8dc94e16

trueqbit (approved):

